PR #18, opened 2025-04-28
Feat/stop streaming env

CougarTasker:
Features: 
 - added new enviroment varible `NEURACORE_LIVE_DATA_ENABLED` to allow streaming to be disabled at all times without a call to `nc.stop_live_data` for use in testing
- the the streaming disabled state to be more robust

Item:
 - [Errors from Streaming when backend not needed](https://trello.com/c/quBL7mHT)


stepjam (approved):
LGTM. 
Can you tidy up the commit history though; you may need to rebranch from staging and then rebase

PR #19, opened 2025-05-11
AsyncRuntime to manage an asyncio event loop in a background thread

stepjam:
- Introduced AsyncRuntime to manage an asyncio event loop in a background thread
- Rewrote ClientStreamingManager to use AsyncRuntime for all async operations

Do not merge before validating in that it solves issue in ROS

### DIFF
--- a/neuracore/core/streaming/client_stream/client_stream_manager.py
+++ b/neuracore/core/streaming/client_stream/client_stream_manager.py
@@ -1,5 +1,4 @@
 import asyncio
-import threading
 import traceback
 from datetime import timedelta
 from typing import Dict, List, Tuple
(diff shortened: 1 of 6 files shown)


CougarTasker:
I like these changes, looks good :+1:

neuracore/core/streaming/client_stream/client_stream_manager.py
```diff
@@ -24,93 +22,94 @@
...
-
-
 # must be less than zero -> a reconnection delay of more
 # than one second is considered dead
 # TODO: resubmit tracks if connection is re-established
 # after more than one second
 MINIMUM_BACKOFF_LEVEL = -2
 
 
+class AsyncRuntime:
```
CougarTasker: I think this should be in its own file, Perhaps we make it a singleton?

stepjam: Done

PR #20, opened 2025-05-12
Remote Recording + Robot Instances

CougarTasker:
Features:
 - Assocated Datastreams with their relvant robot instances
 - Consolidated Recording state and logic
 - Added Enviroment varible to disable starting recording remotely

 Images:
image
image
 
 Item:
 - [Start Remote Recording](https://trello.com/c/k4lcfuPW)
 
 Related Pull Requests:
  - [Frontend](https://github.com/Neuraco/neuracore_frontend/pull/47)
  - [Backend](https://github.com/Neuraco/neuracore_backend/pull/43)

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -133,7 +133,7 @@ def stop_recording(
         logger.warning("No active recordings to stop.")
         return
     recording_id = robot.get_current_recording_id()
-    robot.stop_recording(recording_id, blocking=True)
+    robot.stop_recording(recording_id)
     if wait:
         while backend_utils.get_num_active_streams(recording_id) > 0:
             time.sleep(2.0)
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -1,6 +1,5 @@
 import logging
 import time
-from threading import Thread
 from typing import Optional
 
 from neuracore.core.streaming.client_stream.client_stream_manager import (
@@ -18,7 +17,7 @@ from .globals import GlobalSingleton
 logger = logging.getLogger(__name__)
 
 
-def _get_robot(robot_name: str, instance: Optional[int] = 0) -> Robot:
+def _get_robot(robot_name: str, instance: int) -> Robot:
     """Get a robot by name and instance."""
     robot: Robot = GlobalSingleton()._active_robot
     if robot_name is None:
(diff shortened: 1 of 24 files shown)


neuracore/api/core.py
```diff
@@ -137,17 +129,11 @@ def stop_recording(
...
-    threads: Thread = []
-    for sname, stream in GlobalSingleton()._data_streams.items():
-        if sname.startswith(robot.instanced_id):
-            with stream.lock:
-                if stream.is_recording():
-                    threads.append(stream.stop_recording())
-    recording_id = GlobalSingleton()._active_recording_ids.pop(robot.instanced_id)
-    robot.stop_recording(recording_id)
+    recording_id = robot.get_current_recording_id()
+    robot.stop_recording(recording_id, blocking=True)
```
stepjam: I believe you want `blocking=wait`.
Also, just for consistency, better to either rename blocking-> wait, or wait->blocking. If latter, ensure to update any usages

neuracore/core/const.py
```diff
@@ -1,5 +1,8 @@
 import os
 
+REMOTE_RECORDING_TRIGGER_ENABLED = (
+    os.getenv("NEURACORE_REMOTE_RECORDING_TRIGGER_ENABLED", "True").lower() == "true"
```
stepjam, line 6: Could we add a comment here saying why this might be a useful env variable to expose?

neuracore/core/robot.py
```diff
@@ -31,10 +37,16 @@ def __init__(
         self.mjcf_path = mjcf_path
         self.overwrite = overwrite
         self.shared = shared
-        self.id: str = None
-        self.instanced_id: str = None
+        self.id: str | None = None
         self._auth: Auth = get_auth()
         self._temp_dir = None
+        self.__data_streams: dict[str, DataStream] = dict()
```
stepjam: nit: why `__`? name mangling not ideal unless we absolutely have to

stepjam (commented):
Some comments; want to have another pass through after looking at backend and frontend before LGTMing

neuracore/core/const.py
CougarTasker: Yeah I was adding it for the same reason as the `LIVE_DATA_ENABLED` I thought it would be helpful in some intergration tests etc that you might not want to run the listener if it is not `required/wanted`

stepjam (approved):
LGTM

PR #34, opened 2025-06-09
Fix device mismatch issue on test

Ke-Wang1017:
When running `pytest tests/` it will give errors that not all tensors on the same device.
Screenshot from 2025-06-09 10-17-13

This PR fixes this by manually moving both the data and the model on the same device and the errors don't appear anymore


stepjam (approved):
LGTM

PR #39, opened 2025-06-12
Fix/mypy typing temp

StevenJacobs61:
Fixed mypy typing errors and updated run_validation test

### DIFF
--- a/neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
+++ b/neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
@@ -2,6 +2,7 @@
 
 import logging
 import time
+from typing import Any, Dict, Tuple
 
 import numpy as np
 import torch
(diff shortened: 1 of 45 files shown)


neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
```diff
@@ -35,11 +36,11 @@ def __init__(
         self,
         model_init_description: ModelInitDescription,
         hidden_dim: int = 256,
-        unet_down_dims: list[int] = [
+        unet_down_dims: Tuple[int, ...] = (
```
Ke-Wang1017, line 39: why it has to be tuple, is it because tuple is immutable? :)

neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
```diff
@@ -277,7 +281,7 @@ def _prepare_global_conditioning(
 
     @staticmethod
     def _make_noise_scheduler(
-        name: str, **kwargs: dict
+        name: str, **kwargs: Dict[str, Any]
```
Ke-Wang1017, line 284: As we discussed yesterday, you can change `name` to `noise_scheduler_type` which can be less confusing

PR #40, opened 2025-06-12
Fixed mypy errors

StevenJacobs61:
Fixed typing for MyPY pre-commit errors

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,38 +2,37 @@ default_install_hook_types: [pre-commit]
 default_stages: [commit]
 
 repos:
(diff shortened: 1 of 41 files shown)


tests/unit/ml/algorithms/test_cnnmlp.py
```diff
@@ -202,5 +202,6 @@ def test_run_validation(tmp_path: Path):
         output_dir=tmp_path,
         algorithm_dir=algorithm_dir,
         port=random.randint(10000, 20000),
+        skip_endpoint_check=True,
```
stepjam: skip_endpoint_check should be False (same for other tests)

PR #52, opened 2025-06-27
fix: nc-select-org

CougarTasker:
Changes:
 - Improved error handling with calling select org from command line 

Note:
 - Was previously related to the syncronised lossy video feature now seperate


stepjam (approved):
LGTM

PR #57, opened 2025-07-07
Added support for resurrecting streams from the dead :ghost: 

CougarTasker:
Features:
 - Checkes if a stream was killed when reconnected, re-submits tracks if they were dead
 - Reworked the client stream manger to not block when processing singalling messages 
   - Ensured ice events waited untill the remote description is set

Item:
 - [Recover tracks Item](https://trello.com/c/wtUfFT63/263-recover-stream-tracks-after-internet-drops)

Related Prs:
 - [Backend](https://github.com/Neuraco/neuracore_backend/pull/86)

### DIFF
--- a/neuracore/core/streaming/client_stream/connection.py
+++ b/neuracore/core/streaming/client_stream/connection.py
@@ -7,6 +7,7 @@ Handles SDP negotiation, ICE candidate exchange, and connection lifecycle.
 
 import asyncio
 import json
+import logging
 from dataclasses import dataclass, field
 from typing import Callable
 
@@ -32,6 +33,8 @@ ICE_SERVERS = [
     RTCIceServer(urls="stun:stun1.l.google.com:19302"),
 ]
 
+logger = logging.getLogger(__name__)
+
 
 @dataclass
 class PierToPierConnection:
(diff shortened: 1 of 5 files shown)


neuracore/core/streaming/recording_state_manager.py
```diff
@@ -240,14 +245,8 @@ async def connect_recording_notification_stream(self) -> None:
...
 
-            except ConnectionError as e:
-                print(f"Connection error: {e}")
-                await asyncio.sleep(2 ^ backoff)
-                backoff += 1
             except Exception as e:
-                print(f"Unexpected error: {e}")
-                await asyncio.sleep(2 ^ backoff)
-                backoff += 1
+                print(f"Recording signalling error: {e}")
```
stepjam, line 249: Should we turn these prints to logger.infos instead?

neuracore/core/streaming/client_stream/connection.py
```diff
@@ -294,24 +281,31 @@ async def send_offer(self) -> None:
...
+                await self.send_handshake_message(
+                    MessageType.SDP_OFFER, self.connection.localDescription.sdp
+                )
+
+                await asyncio.wait_for(self.received_answer_event.wait(), timeout=30)
+            except asyncio.TimeoutError:
+                # waiting 30 seconds for answer -> assuming peer is gone
+                await self.close()
+            except Exception:
+                print("Signalling Error: Failed to send offer")
```
stepjam: Same; print -> logging.info

PR #62, opened 2025-07-18
Added remote nodes to endpoint predictions

CougarTasker:
Features: 
 - Created a remote stream data gatherer
   - Added a consumer for remote nodes infomation
   - Setup signalling to support both directions 
   - Created a consumer manager for receiving streamed data and constructing sync points
 - Added support for streaming more kinds of tracks, `language`, `gripper`, `custom`, etc 
 - Added support for a custom lossless png video streams
   
Refactor:
 - moved common event stream functionality to a base class 
 - consolidated background async threads and added new background task tracker for lightweight fire and forget tasks 
 - consolidated encoding and decoding for png<->string convertions
 - moved shared types to nc_types
 
Items:
  - [python-libary-p2p-endpoints](https://trello.com/c/EFpjs67I/265-python-libary-p2p-endpoints)
 
Related PRs:
  - [javascript-frontend](https://github.com/Neuraco/neuracore_frontend/pull/97)
  - [backend](https://github.com/Neuraco/neuracore_backend/pull/91)


neuracore/core/streaming/p2p/stream_manager_factory.py
```diff
@@ -0,0 +1,216 @@
...
+                ),
+            )
+
+        return self.provider_managers[key]
+
+
+_p2p_manager_factory: Optional[StreamManagerFactory] = None
+
+
+def get_p2p_manager_factory() -> StreamManagerFactory:
```
stepjam: Considering this is a singleton, why not use a standard pattern here, e.g. within StreamManagerFactory, having:

```
class StreamManagerFactory:
    _instance = None

    def __new__(cls):
        if cls._instance is None:
            cls._instance = super(StreamManagerFactory, cls).__new__(cls)
        return cls._instance
```

neuracore/core/streaming/p2p/stream_manager_factory.py
```diff
@@ -0,0 +1,216 @@
...
+from neuracore.core.streaming.p2p.provider.global_live_data_enabled import (
+    global_consume_live_data_manager,
+    global_provide_live_data_manager,
+)
+from neuracore.core.streaming.p2p.signalling_events_consumer import (
+    SignallingEventsConsumer,
+)
+
+
+class StreamManagerFactory(BaseStreamManagerFactory):
```
stepjam: Maybe factor is not a great name here; factories don't tend to cache, and they never do anything other than create, whereas you have some remove ability here too.
The best term is manager, but I can see why you didn't use that because you would have ended up with `StreamManagerManger`.  
How about `StreamManagerOrchestrator`, or similar?

neuracore/api/endpoints.py
```diff
@@ -210,3 +211,28 @@ def delete_endpoint(endpoint_id: str) -> None:
         response.raise_for_status()
     except Exception as e:
         raise ValueError(f"Error deleting endpoint: {e}")
+
+
+def get_data_gatherer(
```
stepjam: Perhaps better here to have this function use the data gatherer to get the latest sync point, i.e. by calling `data_gatherer.get_latest_data()` and then retuning a SyncPoint. Also, I think we need to make it as clear/verbose as possible from the name what this actually does. Maybe something like `get_latest_sync_point() -> SyncPoint` or even `get_latest_sync_point_from_live_data() -> SyncPoint`

neuracore/core/streaming/p2p/consumer/sync_point_parser.py
```diff
@@ -0,0 +1,155 @@
...
+    Returns:
+        SyncPoint: A SyncPoint object containing the parsed data.
+
+    Raises:
+        ValueError: If the track kind is unsupported or data validation fails.
+    """
+    try:
+        if track_details.kind == TrackKind.JOINTS:
+            joint_data = JointData.model_validate_json(message_data)
+            return SyncPoint.model_validate(
```
stepjam, line 38: Why is this sync point constructed differently than the others?

tests/integration/test_consume_stream.py
stepjam, line 1: Note that integration tests are not run in any actions. If this is a lightweight test, then perhaps add it as an addition pytest call in the github actions config

neuracore/core/streaming/p2p/consumer/sync_point_parser.py
CougarTasker: This case covers all joint data channels like joint position, angle, etc. This parse operation pretty much mirrors the logging.py methods 1-1 this would be `_log_joint_data`

tests/integration/test_consume_stream.py
CougarTasker: Yeah good idea, although its a bit of a catch 22 it won't pass until the backend goes into production, I will do it in another PR

PR #63, opened 2025-07-21
SyncPoints have consistent ordering & predictions return SyncPoint

stepjam:
Still needs to be tested fully before merging, but ready for review

### DIFF
--- a/neuracore/api/endpoints.py
+++ b/neuracore/api/endpoints.py
@@ -24,14 +24,13 @@ from neuracore.core.get_latest_sync_point import (
 from neuracore.core.get_latest_sync_point import (
     get_latest_sync_point as _get_latest_sync_point,
 )
-from neuracore.core.nc_types import SyncPoint
+from neuracore.core.nc_types import DataType, SyncPoint
 
 
 def policy(
     train_run_name: Optional[str] = None,
     model_file: Optional[str] = None,
-    robot_name: Optional[str] = None,
-    instance: int = 0,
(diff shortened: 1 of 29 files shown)


neuracore/api/endpoints.py
```diff
@@ -24,14 +24,15 @@
...
-from neuracore.core.nc_types import SyncPoint
+from neuracore.core.nc_types import DataType, SyncPoint
 
 
 def policy(
     train_run_name: Optional[str] = None,
     model_file: Optional[str] = None,
     robot_name: Optional[str] = None,
     instance: int = 0,
+    output_mapping: Optional[dict[DataType, list[str]]] = None,
```
CougarTasker, line 33: Please add a pydoc description for this arg

tests/unit/core/test_nc_types.py
CougarTasker, line 1: `test_nc_types` sounds a little cursed to me. I feel as if mixing logic with these types may lead to headaches down the line. I can see the apeal of putting the order method on the nc data but perhaps we move it into its own method?

stepjam: I looked at this again, and I'm not sure this would work. Theres not much to pull out into its own method (most of the code is just constructing the object that is returned). Even if we were to extract it out, we would introduce a cyclic dependency on SyncData. Unless you had something else in mind?

stepjam: Discussed offline. Plan is to separate out nc_types later down the line and rename to models (as they are not actually types)

PR #66, opened 2025-07-22
Throws InsufficientSyncPointError on prediction with insufficient data

stepjam:


neuracore/core/nc_types.py
```diff
@@ -878,3 +878,9 @@ class RecordingNotification(BaseModel):
         BaseRecodingUpdatePayload,
     ]
     id: str = Field(default_factory=lambda: uuid4().hex)
+
+
+class InsufficientSyncPointError(Exception):
+    """Exception when SyncPoint doesn't contain sufficient data for inference."""
+
+    pass
```
CougarTasker: For consitency I think this should be in the `neuracore/core/exceptions.py` file instead

examples/example_local_endpoint.py
```diff
@@ -1,9 +1,12 @@
+import time
+
 import matplotlib.pyplot as plt
 from common.constants import BIMANUAL_VIPERX_URDF_PATH, EPISODE_LENGTH
 from common.ee_sim_env import sample_box_pose
 from common.sim_env import BOX_POSE, make_sim_env
 
 import neuracore as nc
+from neuracore.core.nc_types import InsufficientSyncPointError
```
CougarTasker: I don't think we should be encoraging our users to import from core directly as we will end up needing to make a major bump almost every time. I think something like these would be better
```python
from neuracore import InsufficientSyncPointError # like pydantic
from neuracore.exceptions import InsufficientSyncPointError # like numpy
```

the advantage of the former is that it allows users to do `nc.InsufficientSyncPointError` rather than adding an extra import if they prefer but perhaps we offer both?

examples/example_local_endpoint.py
```diff
@@ -59,7 +62,13 @@ def main():
...
             idx_in_horizon = i % horizon
             if idx_in_horizon == 0:
-                predicted_sync_points = policy.predict()
+                while True:
+                    try:
+                        predicted_sync_points = policy.predict()
+                        break
+                    except InsufficientSyncPointError:
+                        time.sleep(0.5)
+
```
CougarTasker: I know we discussed this yesterday but the more I look at this the more I dislike it, it looks really hacky. 

Exceptions should not be thrown during normal execution or used for control flow. I think either we add a blocking option to predict

```suggestion
                predicted_sync_points = policy.predict(block_untill_connected=True)
```

or we use a special return value something like this,

```suggestion
                while not (predicted_sync_points := policy.predict()):
                    time.sleep(0.5)
```

examples/example_server_endpoint.py
CougarTasker, line 1: please see my comments to the local endpoint example

neuracore/core/nc_types.py
stepjam: Done

neuracore/ml/utils/policy_inference.py
```diff
@@ -606,6 +607,61 @@ def _model_prediction_to_sync_points(
...
+        Args:
+            sync_point: SyncPoint containing data from a single time step.
+
+        Raises:
+            ValueError: If the sync point does not contain required data types.
+        """
+        input_data_types = self.model.model_init_description.input_data_types
+        for data_type in input_data_types:
+            if data_type == DataType.JOINT_POSITIONS and not sync_point.joint_positions:
+                raise InsufficientSyncPointError(
```
CougarTasker: Instead of rasing immediately it would be better to aggregate all of the missing data types and inform the user in one go. As it is frustrating to have to add them one at a time waiting to see what the next error will be.

CougarTasker (approved):
LGTM

PR #73, opened 2025-08-05
Feat/log cloud data

favour-neuraco:
Enhance Point Cloud Logging with Base64 Support/ Compression & Model Updates

Features:

Enhanced log_point_cloud to support Base64-encoded and optionally compressed point cloud data (points_b64, rgb_points_b64), improving efficiency for large datasets.

Refined PointCloudData model: clarified optional fields, added explicit support for encoded data, and aligned defaults for better backend compatibility.

Improved validation and encoding logic for point cloud and RGB data to ensure robustness and reduce risk of malformed payloads.

Fixes:

Adjusted payload structure in log_point_cloud to maintain consistency between binary and non-binary modes.

Tag:
@stepjam

### DIFF
--- a/neuracore/api/logging.py
+++ b/neuracore/api/logging.py
@@ -764,24 +764,22 @@ def log_point_cloud(
     timestamp = timestamp or time.time()
     if not isinstance(points, np.ndarray):
         raise ValueError("Point cloud must be a numpy array")
-    if points.dtype != np.float32:
-        raise ValueError("Point cloud must be float32")
+    if points.dtype not in (np.float16, np.float32):
+        raise ValueError("Point cloud must be float16 or float32")
     if points.shape[1] != 3:
         raise ValueError("Point cloud must have 3 columns")
     if points.shape[0] > 307200:
         raise ValueError("Point cloud must have at most 307200 points")
+
     if rgb_points is not None:
         if not isinstance(rgb_points, np.ndarray):
             raise ValueError("RGB point cloud must be a numpy array")
         if rgb_points.dtype != np.uint8:
             raise ValueError("RGB point cloud must be uint8")
         if rgb_points.shape[0] != points.shape[0]:
-            raise ValueError(
-                "RGB point cloud must have the same number of points as the point cloud"
-            )
+            raise ValueError("RGB point cloud must match number of points")
         if rgb_points.shape[1] != 3:
             raise ValueError("RGB point cloud must have 3 columns")
-        rgb_points = rgb_points.tolist()
 
     extrinsics, intrinsics = _validate_extrinsics_intrinsics(extrinsics, intrinsics)
     robot = _get_robot(robot_name, instance)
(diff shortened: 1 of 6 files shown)


neuracore/api/logging.py
```diff
@@ -744,6 +755,9 @@ def log_point_cloud(
     robot_name: Optional[str] = None,
     instance: int = 0,
     timestamp: Optional[float] = None,
+    binary: bool = True,
+    compress: bool = True,
+    compression_level: int = 1,
```
CougarTasker: So i get where you are comming from but I don't think the user should know or have to consider how we represent the point cloud data internally

neuracore/core/nc_types.py
```diff
@@ -148,10 +148,13 @@ class PointCloudData(NCData):
...
     """
 
-    points: list[list[float]]
+    points: Optional[list[list[float]]] = []
     rgb_points: Optional[list[list[int]]] = None
+    points_b64: Optional[str] = None
+    rgb_points_b64: Optional[str] = None
     extrinsics: Optional[list[list[float]]] = None
     intrinsics: Optional[list[list[float]]] = None
+    compressed: bool = False
```
CougarTasker: I think lets just pick whether the points will be compressed and stick to it by supporting both I think we will be making the code un-necessarily complicated

neuracore/core/nc_types.py
```diff
@@ -148,10 +148,13 @@ class PointCloudData(NCData):
     optional color information and camera calibration for registration.
     """
 
-    points: list[list[float]]
+    points: Optional[list[list[float]]] = []
```
CougarTasker: Making this optinal is going to require code to handle each of the decoding methods. I think using a custom [@field_serializer](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer) and  [@field_validator](https://docs.pydantic.dev/latest/concepts/validators/#using-the-decorator-pattern) will make this seamless

CougarTasker (changes requested):
I think the idea of using GZip is great but I think this code opens up to many paths where the points and rgb points can be encoded as any combination of lists, binary and compressed binary.

I think that sticking to one representation will make this change easier to deal with downstream

neuracore/core/nc_types.py
```diff
@@ -148,11 +150,55 @@ class PointCloudData(NCData):
...
+        decompressed = gzip.decompress(base64.b64decode(data))
+        arr = np.frombuffer(decompressed, dtype=dtype).reshape(-1, cols)
+        return arr.tolist()
+
+    @staticmethod
+    def _encode(arr: list, dtype: type) -> str:
+        """Encode a list of points into compressed Base64 format."""
+        np_arr = np.array(arr, dtype=dtype)
+        compressed = gzip.compress(np_arr.tobytes(), compresslevel=3)
+        return base64.b64encode(compressed).decode("utf-8")
```
CougarTasker, line 170: This encode method is not used?

CougarTasker (changes requested):
lets fix up this encode method :)

PR #75, opened 2025-08-07
feat: Improved connection error messages.

CougarTasker:
Features:
 - added specialized error message for connection issues helps users identify that this is not an application error


stepjam (approved):
LGTM

PR #79, opened 2025-08-26
doc: added documentation for tmpdir location

CougarTasker:
What do you think about making this explicit? this env controls the tmpdir used for datasets


stepjam:
Agreed - explicit may be nicer; default being the tmp dir.

stepjam (approved):

PR #82, opened 2025-09-01
Accept and return string-encoded point clouds

favour-neuraco:
Implemented validators and serializers on PointCloudData so that points and rgb_points can accept base64-gzip encoded strings and re-encode lists on output. This ensures the backend can handle incoming encoded data while the frontend receives the string format it expects, fixing validation errors and empty point clouds.

### DIFF
--- a/neuracore/core/nc_types.py
+++ b/neuracore/core/nc_types.py
@@ -6,6 +6,8 @@ and model predictions. All data types include automatic timestamping and
 support for serialization via Pydantic.
 """
 
+import base64
+import gzip
 import time
 from datetime import datetime, timezone
 from enum import Enum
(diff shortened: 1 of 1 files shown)


neuracore/core/nc_types.py
```diff
@@ -142,17 +144,57 @@ def order(self) -> "EndEffectorData":
...
+    """3D point cloud data with optional RGB colouring and camera parameters.
 
     Represents 3D spatial data from depth sensors or LiDAR systems with
-    optional color information and camera calibration for registration.
+    optional colour information and camera calibration for registration.
     """
 
-    points: list[list[float]]
-    rgb_points: Optional[list[list[int]]] = None
+    points: Optional[Union[str, list[list[float]]]] = None
```
CougarTasker: why are we changing the type here we are using before validators so this should not be neccisarry. lets keep the type the same for compatability

CougarTasker (changes requested):
Looks good just need to revert that one type

CougarTasker (approved):
LGTM

PR #95, opened 2025-10-20
fix device issue on test/ml/algorithms

kwangneuraco:
### Bugfixes

 - Fix bugs appears in the test with `device mismatch`


aditya-neuraco (approved):
LGTM!

PR #96, opened 2025-10-21
fix: background tasks completed after threadpool shutdown

muneebneura:
### Features

 - None

### Bugfixes
This bug occurs when the threadpool executor has shutdown and we still have background tasks running which try to create subtasks, such as making request to update num active streams and removing background task. When a ThreadPoolExecutor is shutdown it cannot accept new tasks.

To avoid the problem entirely we can either ensure tasks are only submitted to the ThreadPoolExecutor from one context, e.g. the point in the program where the ThreadPoolExecutor is created and shutdown. We can use a queue to connect done callback functions to the main thread to signal what and when subtasks should be issued.

This exception however,  will be resolved when we switch to daemon.

 - Suppressed the runtime error (Cannot schedule new futures after shutdown)
 - Tested with random camera frames.
 - Occurs sometimes with the usual data collection as well with a slow internet connection

### Items

 - [NC-265](https://neuracore.atlassian.net/browse/NC-265)

### Related PRs

 - None

### DIFF
--- a/neuracore/core/utils/background_coroutine_tracker.py
+++ b/neuracore/core/utils/background_coroutine_tracker.py
@@ -39,6 +39,12 @@ class BackgroundCoroutineTracker:
             task.result()
         except asyncio.CancelledError:
             logger.info("Background task cancelled")
+        except RuntimeError as e:
+            if "cannot schedule new futures after shutdown" in str(e).lower():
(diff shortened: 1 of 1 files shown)


CougarTasker (approved):
As we have discussed this is okay for now :)

PR #101, opened 2025-10-24
Delete stored checkpoints on disk if not logged to cloud

StevenJacobs61:
Delete stored checkpoints on disk if not logged to cloud


aditya-neuraco (approved):
LGTM!

PR #109, opened 2025-10-28
fix: handle expired recordings

muneebneura:
### **Description**

Recordings have a 5-minute cap enforced by a backend cron job. The Python client was unaware of this and continued sending upload and stream-update requests after expiry, resulting in multiple 500 errors and shutdown.

This PR adds handling for expired recordings, once a recording is marked expired, all further uploads and active-stream updates are stopped locally to prevent backend errors.

### Bugfixes

 - Added timer to track active recordings duration
 - Gated requests for the expired recordings
 - Timer cancellation for recording id, if a recording was saved

### Items

 - [NC-303](https://neuracore.atlassian.net/browse/NC-303)

### DIFF
--- a/neuracore/core/streaming/bucket_uploaders/streaming_video_uploader.py
+++ b/neuracore/core/streaming/bucket_uploaders/streaming_video_uploader.py
@@ -23,6 +23,7 @@ from neuracore.core.config.get_current_org import get_current_org
 from neuracore.core.const import API_URL
 from neuracore.core.exceptions import EncodingError
 from neuracore.core.nc_types import CameraData
+from neuracore.core.streaming.recording_state_manager import get_recording_state_manager
 
 from .bucket_uploader import BucketUploader
 from .resumable_upload import ResumableUpload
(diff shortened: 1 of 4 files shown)


neuracore/core/streaming/bucket_uploaders/streaming_video_uploader.py
```diff
@@ -179,6 +181,13 @@ def _check_codec_support(self) -> None:
                 " 2. Rebuild PyAV against it (`pip install --no-binary av av`)"
             )
 
+    def is_recording_expired(self) -> bool:
+        """Check if the current recording id is expired."""
+        expired_recording_ids = self._recording_manager.get_recordings_expired()
+        if self.recording_id in expired_recording_ids:
+            return True
+        return False
```
CougarTasker: this is repeated i think it might make more sense for the `is_recording_expired(recording_id)` to belong to the recording state manager

neuracore/core/streaming/recording_state_manager.py
```diff
@@ -113,6 +120,28 @@ def is_recording(self, robot_id: str, instance: int) -> bool:
         )
         return instance_key in self.recording_robot_instances
 
+    def get_recordings_expired(self) -> list[str]:
+        """Returns a list of expired recording IDs."""
+        return list(self._expired_recording_ids)
```
CougarTasker: as i mentioned before lets provide a `is_recording_expired` instead this reduces the coupling so give us more freedom in the future. should be simple `return recording_id in self._expired_recording_ids`

CougarTasker (approved):
LGTM \w comment

PR #97, opened 2025-10-21
feat: added end effector pose and parallel gripper open amount logging

aditya-neuraco:
### Features
- Added two new data types in NC Core types. 
   - EndEffectorPoseData
   - ParallelGripperOpenAmount
- They store dictionaries of end-effector/gripper names as keys and poses/open-amounts as values.
- Added logging functions for these data types. 

### Items
 - [NC-302](https://neuracore.atlassian.net/browse/NC-302)

### Related PRs
 - https://github.com/NeuracoreAI/neuracore_backend/pull/151

PR #110, opened 2025-10-29
Small Fix on Examples

kwangneuraco:
### Features

1. resets box pose every reset and multiple validation episode
2. add if condition to speed up data collection

### DIFF
--- a/examples/example_data_collection_vx300s.py
+++ b/examples/example_data_collection_vx300s.py
@@ -57,11 +57,6 @@ def main(args):
         nc.log_language("Pick up the cube and pass it to the other robot", timestamp=t)
         nc.log_rgb(CAM_NAME, obs.cameras[CAM_NAME].rgb, timestamp=t)
 
-        if obs.end_effector_poses is not None:
-            nc.log_end_effector_poses(obs.end_effector_poses, timestamp=t)
-        if obs.gripper_open_amounts is not None:
-            nc.log_parallel_gripper_open_amounts(obs.gripper_open_amounts, timestamp=t)
-
         # Execute action trajectory while logging
         for action in action_traj:
             obs, reward, done = env.step(np.array(list(action.values())))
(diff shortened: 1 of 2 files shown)


examples/example_data_collection_vx300s.py
```diff
@@ -75,8 +75,12 @@ def main(args):
...
             nc.log_joint_target_positions(action, timestamp=t)
             nc.log_rgb(CAM_NAME, obs.cameras[CAM_NAME].rgb, timestamp=t)
-            nc.log_end_effector_poses(obs.end_effector_poses, timestamp=t)
-            nc.log_parallel_gripper_open_amounts(obs.gripper_open_amounts, timestamp=t)
+            if obs.end_effector_poses is not None:
+                nc.log_end_effector_poses(obs.end_effector_poses, timestamp=t)
+            if obs.gripper_open_amounts is not None:
+                nc.log_parallel_gripper_open_amounts(
+                    obs.gripper_open_amounts, timestamp=t
+                )
```
aditya-neuraco: I believe we are checking if `obs.end_effector_poses` is not None in the `base_env.py` file so no need to add the checks in examples. #97 might give you more context.

```python
  def get_end_effector_poses(self) -> Dict[str, List[float]]:
      """Get end effector poses from MuJoCo body positions and orientations.

      Returns:
          Dictionary mapping end effector names to
          7-element pose lists [x, y, z, qx, qy, qz, qw].
      """
      poses = {}

      left_ee_pose = self._get_end_effector_pose("vx300s_left/gripper_link")
      if left_ee_pose is not None:
          poses["left_ee"] = left_ee_pose

      right_ee_pose = self._get_end_effector_pose("vx300s_right/gripper_link")
      if right_ee_pose is not None:
          poses["right_ee"] = right_ee_pose

      return poses
```

kwangneuraco: Thanks but adding this will slow down the streaming and let's discuss together with Stephen

neuracore/ml/train.py
```diff
@@ -346,8 +346,8 @@ def run_training(
             logger.info("Starting training...")
             trainer.train(start_epoch=start_epoch)
             logger.info("Training completed successfully!")
-        except Exception:
-            logger.error("Training failed.", exc_info=True)
+        except Exception as e:
+            logger.error(f"Training failed with error: {e}", exc_info=True)
```
stepjam: Putting e in the message is not necessary when `exc_info=True`

PR #111, opened 2025-10-30
Feat/quick fix ram leak

kwangneuraco:
### Features
This is what we discussed as a quick fix, deep copy the syncedPoint to the disk and avoid ram leak

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -1,5 +1,6 @@
 """Synchronized recording iterator."""
 
+import copy
 import logging
 import os
 from concurrent.futures import ThreadPoolExecutor, as_completed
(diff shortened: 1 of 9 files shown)


neuracore/ml/utils/memory_monitor.py
```diff
@@ -53,6 +53,10 @@ def check_memory(self, log: bool = False) -> None:
         """
         ram = psutil.virtual_memory()
         ram_used_ratio = ram.used / ram.total
+        logger.info(f"Memory usage of ram: {ram.used / (1024**3)} GB")
```
stepjam: Please remove logs as we already have logs below

neuracore/core/data/synced_recording.py
```diff
@@ -362,8 +363,8 @@ def __getitem__(self, idx: Union[int, slice]) -> Union[SyncPoint, list[SyncPoint
         if idx < 0 or idx >= len(self):
             raise IndexError("Index out of range")
 
-        # Get sync point data
-        sync_point: SyncPoint = self._recording_synced.frames[idx]
+        # Get sync point data, use deepcopy to avoid the memory leak
```
stepjam: Please be a bit more thorough with comment. Imagine someone comes across this who was not in our discussion

stepjam (approved):
LGTM w/ comments

stepjam (approved):
LGTM. Please squash before merge

PR #115, opened 2025-11-03
Use loaded syncpoint instead of creating new ones when using joint_target_positions as output

ypang-neuraco:
When using joint_target_position as output, a new list of syncpoints is created as we want to use the joint_target_position from the current timestep. This causes the video frames to be loaded again slowing down training.

This fix creates the syncpoint list by combining the syncpoint at the current timestep with the future_sync_points list that are already loaded earlier instead of creating a new variable.

### DIFF
--- a/neuracore/ml/datasets/pytorch_synchronized_dataset.py
+++ b/neuracore/ml/datasets/pytorch_synchronized_dataset.py
@@ -299,7 +299,7 @@ class PytorchSynchronizedDataset(PytorchNeuracoreDataset):
                     # We dont need to shift the sync_point by 1, since we are
                     # using the target joint positions as the action
                     jtp_points = [sync_point] + future_sync_points
-                    jtp_points = jtp_points[:self.output_prediction_horizon]
+                    jtp_points = jtp_points[: self.output_prediction_horizon]
 
                     sample.outputs.joint_target_positions = (
                         self._create_joint_maskable_output_data(


neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -298,14 +298,9 @@ def load_sample(
...
-                        list[SyncPoint],
-                        synced_recording[
-                            timestep : timestep + self.output_prediction_horizon
-                        ],
-                    )
-                    for _ in range(self.output_prediction_horizon - len(jtp_points)):
-                        jtp_points.append(jtp_points[-1])
+                    jtp_points = [sync_point] + future_sync_points
+                    jtp_points = jtp_points[: self.output_prediction_horizon]
+
```
kwangneuraco, line 303: The old code pads values if the episode nears end and doesn't have enough length. Is it the feature we want to have? Does the new code handles it?

kwangneuraco: what if jtp_points is too short

ypang-neuraco: This is handled earlier in the code where future_sync_points is paded
https://github.com/NeuracoreAI/neuracore/blob/465be514000c61f61e7ff0e521ac6ae291044db3/neuracore/ml/datasets/pytorch_synchronized_dataset.py#L151

kwangneuraco (approved):
LGTM!!

PR #116, opened 2025-11-03
speed up synced_recording 

kwangneuraco:
…rd decode if iterating frames in the video

### Features
Speed up the synced_recording by caching the video and do monotonic forwarding if iterating through the video.
Synced_recording is called when we do training at 1st epoch or want to visualize the data. It is the main speed bottleneck for training at 1st epoch

### JIRA Ticket
[NCRL-10](https://neuracore.atlassian.net/browse/NCRL-10?atlOrigin=eyJpIjoiOTE0ZDQ2YWIwZDU3NGIyNjk5ZDZhMmU4ODE0NjdlZDEiLCJwIjoiaiJ9)

### Test
Tested with `examples/example_view_dataset.py`, and measuring the time used to load data. The speed is around 7x faster

Screenshot from 2025-11-03 16-16-40
Screenshot from 2025-11-03 16-17-00

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -101,8 +101,6 @@ class SynchronizedRecording:
         if self._per_step_workers and self._per_step_workers > 0:
             self._executor = ThreadPoolExecutor(max_workers=self._per_step_workers)
 
-    # ---------------- API calls ----------------
-
     def _get_synced_data(self) -> SyncedData:
         """Retrieve synchronized metadata for the recording.
 
@@ -213,8 +211,6 @@ class SynchronizedRecording:
                     for f in as_completed(futures):
                         f.result()
 
-    # ---------------- optimized video frame access ----------------
-
     def _ensure_open(self, camera_type: str, cam_id: str, video_file: Path) -> None:
         """Open a video file and initialize decode state.
 
@@ -233,8 +229,8 @@ class SynchronizedRecording:
             self._video_streams[camera_type][cam_id] = stream
             # initialize decode state
             self._decode_state[camera_type][cam_id] = {
-                "last_pts": -1,  # last decoded pts
-                "last_rel_ts": -1.0,  # last requested relative seconds
+                "last_pts": -1,
+                "last_rel_ts": -1.0,
             }
 
     def _decode_to_target(
@@ -242,13 +238,22 @@ class SynchronizedRecording:
         camera_type: str,
         cam_id: str,
         target_rel_ts: float,
-        *,
-        allow_reseek_back_gap_s: float = 0.05,
     ) -> np.ndarray:
         """Decode forward to the frame at/after target_rel_ts (seconds from t0).
 
         We only reseek if the target is sufficiently behind the last decoded
         position.
+
+        Args:
+            camera_type: Type of camera (e.g., "rgbs", "depths").
+            cam_id: Unique identifier for the camera.
+            target_rel_ts: Relative timestamp to decode to (seconds from t0).
+
+        Returns:
+            Numpy array of the decoded frame.
+
+        Raises:
+            ValueError: If no frame is found for the camera at the specified timestamp.
         """
         container = self._video_containers[camera_type][cam_id]
         stream = self._video_streams[camera_type][cam_id]
@@ -257,6 +262,7 @@ class SynchronizedRecording:
         fps = float(stream.average_rate) if stream.average_rate else 0.0
         tbase = float(stream.time_base) if stream.time_base else (1.0 / max(fps, 1.0))
         target_pts = int(target_rel_ts / tbase)
+        allow_reseek_back_gap_s = 1.0 / self.frequency
 
         last_pts = int(state["last_pts"])
         last_rel_ts = float(state["last_rel_ts"])
@@ -354,8 +360,6 @@ class SynchronizedRecording:
 
         return out
 
-    # ---------------- public helpers ----------------
-
     def _populate_video_frames(
         self,
         camera_data: dict[str, CameraData],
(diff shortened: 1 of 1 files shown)


neuracore/core/data/synced_recording.py
```diff
@@ -1,4 +1,4 @@
-"""Synchronized recording iterator."""
+"""Synchronized recording iterator (optimized)."""
```
stepjam, line 1: remove `(optimized)`

neuracore/core/data/synced_recording.py
```diff
@@ -65,13 +65,42 @@ def __init__(
...
+        self.cache_manager = CacheManager(self.cache_dir)
 
         self._iter_idx = 0
         self._suppress_wget_progress = True
 
+        # Caches for open video containers/streams
+        self._video_containers: dict[str, dict[str, av.container.Container]] = {}
+        self._video_streams: dict[str, dict[str, av.video.stream.VideoStream]] = {}
+
+        # NEW: per-camera decode state (to avoid seek each step)
```
stepjam, line 77: remove `new`

neuracore/core/data/synced_recording.py
```diff
@@ -65,13 +65,42 @@ def __init__(
...
+        # camera_type -> cam_id -> dict(last_pts:int, last_rel_ts:float)
+        self._decode_state: dict[str, dict[str, dict[str, float | int]]] = {
+            "rgbs": {},
+            "depths": {},
+        }
+
+        # NEW: precompute t0 metadata once (absolute timestamps per camera)
```
stepjam, line 84: remove `new`

neuracore/core/data/synced_recording.py
```diff
@@ -274,29 +373,61 @@ def _populate_video_frames(
...
+
+    def __del__(self) -> None:
+        """Cleanup when object is garbage collected (best effort)."""
+        try:
+            if hasattr(self, "_video_containers"):
+                self.close()
+        except Exception:
+            pass
+
+    # ---------------- iteration / indexing ----------------
```
stepjam, line 430: remove comment

neuracore/core/data/synced_recording.py
stepjam, line 1: Make sure your code is Python 3.9+ compatible. Unfortunately. it looks like there are no unit tests for this class, otherwise it would have been picked up. (would be good to add a ticket for unit testing this class!)

You need to import Dict/List, etc from typing

neuracore/core/data/synced_recording.py
```diff
@@ -65,13 +65,42 @@ def __init__(
...
 
         self._iter_idx = 0
         self._suppress_wget_progress = True
 
+        # Caches for open video containers/streams
+        self._video_containers: dict[str, dict[str, av.container.Container]] = {}
+        self._video_streams: dict[str, dict[str, av.video.stream.VideoStream]] = {}
+
+        # NEW: per-camera decode state (to avoid seek each step)
+        # camera_type -> cam_id -> dict(last_pts:int, last_rel_ts:float)
```
stepjam, line 78: This comment is not super helpful. Please add more detail about this mapping, or instead make it a dataclass to make it clearer what this mapping is

neuracore/core/data/synced_recording.py
stepjam: And same as above the comment doesn't do much to explain the code.

neuracore/core/data/synced_recording.py
```diff
@@ -65,13 +65,42 @@ def __init__(
...
+        }
+        if _rgb:
+            for cid, cdata in _rgb.items():
+                self._t0_by_type["rgbs"][cid] = cdata.timestamp
+        if _depth:
+            for cid, cdata in _depth.items():
+                self._t0_by_type["depths"][cid] = cdata.timestamp
+
+        # Optional: small thread pool for per-step multi-camera decode
+        self._per_step_workers = int(
```
stepjam, line 97: why is this a member variable?

neuracore/core/data/synced_recording.py
stepjam: Also, it might be worth adding a summary of this class as there are a few moving parts

neuracore/core/data/synced_recording.py
```diff
@@ -354,16 +487,14 @@ def __getitem__(self, idx: Union[int, slice]) -> Union[SyncPoint, list[SyncPoint
...
         """
         if isinstance(idx, slice):
             # Handle slice objects
             start, stop, step = idx.indices(len(self))
             return [cast(SyncPoint, self[i]) for i in range(start, stop, step)]
 
-        if idx < 0:
-            idx += len(self)
-        if idx < 0 or idx >= len(self):
+        idx = idx if idx >= 0 else idx + len(self)
```
stepjam, line 496: nit: I think the old line was semantically clearer

neuracore/core/data/synced_recording.py
```diff
@@ -162,104 +190,175 @@ def _ensure_videos_downloaded(self, camera_type: str) -> None:
...
             video_cache_path = self.cache_dir / f"{self.id}_videos" / camera_type
             video_cache_path.mkdir(parents=True, exist_ok=True)
-
-            # Ensure cache space and download videos if needed
             for cam_id in camera_ids:
                 video_file = video_cache_path / f"{cam_id}.mp4"
                 if not video_file.exists():
-                    # Ensure space before downloading
                     self.cache_manager.ensure_space_available()
                     self._download_video(camera_type, cam_id, video_cache_path)
```
stepjam, line 330: I have just realised that this case should be impossible, right? We have already called `_ensure_videos_downloaded` and so this is likely dead code?

PR #117, opened 2025-11-04
fix: clear stale current_org_id on account switch

favour-neuraco:
### Features
 - None

### Bugfixes
 - Fix stale org selection after account switch: clear current_org_id when API key changes to prevent org-scoped requests using stale org or returning 403s.

### Items
 - [NC-397](https://neuracore.atlassian.net/browse/NCP-397)

### Related PRs
 - None

### DIFF
--- a/neuracore/core/auth.py
+++ b/neuracore/core/auth.py
@@ -12,6 +12,7 @@ from typing import Optional
 import requests
 from neuracore_types import __version__ as nc_types_version
 
+from neuracore.api.orgs_fetch import fetch_org_ids
 from neuracore.core.config.config_manager import get_config_manager
 from neuracore.core.config.get_api_key import get_api_key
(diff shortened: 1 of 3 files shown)


neuracore/core/auth.py
```diff
@@ -66,6 +66,16 @@ def login(self, api_key: Optional[str] = None) -> None:
                 )
             token_data = response.json()
             self._access_token = token_data["access_token"]
+
+            config_manager = get_config_manager()
+            prev_key = config_manager.config.api_key
+
+            if prev_key != api_key:
+                # account switched - drop stale org
+                if config_manager.config.current_org_id is not None:
```
CougarTasker: The old org may not be stale here:

  - the api key may just have been refreshed (this is not the user id) so this might be the same account
  - the org may have been shared between multiple accounts

please can we check if this org is still valid by checking the user's orgs list before removing it.

PR #121, opened 2025-11-06
Cache images rather than videos

stepjam:
### Features
 - Cache images rather than videos.
 - Option for prefetching video data
 - Tests for data classes

### Items
 - [NCRL-10](https://neuracore.atlassian.net/browse/NCRL-10)

### Note
- There is some double caching happening (outside of this PR) when training. There will be a image cache and a tensor cache via the pytorch data loader. It may be possible to actually remove the tensor cache (the image cache is more general), but this will need to be profiled. Ticket created for this here: https://neuracore.atlassian.net/browse/NCRL-23


neuracore/core/data/synced_dataset.py
```diff
@@ -31,14 +36,41 @@ def __init__(
...
+            for rec in self.dataset.recordings:
+                cache = self.dataset.cache_dir / f"{rec['id']}" / f"{self.frequency}Hz"
+                if not cache.exists():
+                    prefetch_needed = True
+                    break
+            if prefetch_needed:
+                self._perform_videos_prefetch(max_workers=max_workers)
+
+    def _perform_videos_prefetch(self, max_workers: int) -> None:
+        """Prefetch video data for all recordings using multiple threads."""
```
kwangneuraco: should we add docs on the input parameter?

neuracore/core/data/dataset.py
```diff
@@ -56,6 +57,7 @@ def __init__(
         self.tags = tags
         self.is_shared = is_shared
         self.recordings = recordings or self._get_recordings()
+        self.data_types = data_types or []
```
aditya-neuraco, line 60: Why are we making the `data_types` optional? Won't a dataset always have data_types?

neuracore/core/data/synced_recording.py
```diff
@@ -47,31 +47,29 @@ def __init__(
...
             frequency: Frequency at which to synchronize the recording.
             data_types: List of DataType to include in synchronization.
+            prefetch_videos: Whether to prefetch video data to cache on initialization.
         """
         self.dataset = dataset
         self.id = recording_id
         self.frequency = frequency
         self.data_types = data_types or []
-        self.cache_dir: Optional[Path] = dataset.cache_dir
+        self.cache_dir: Path = dataset.cache_dir
```
kwangneuraco, line 56: so cache_dir is a necessary now?

neuracore/core/data/dataset.py
stepjam: So data_types can be empty (if its an empty dataset), but I agree that it should not be optional, and we can just pass in an empty list in that case

neuracore/core/data/synced_recording.py
stepjam: Yes. In fact it was always set, the type was misleading

neuracore/core/data/synced_dataset.py
```diff
@@ -31,14 +36,41 @@ def __init__(
...
         self.data_types = data_types or []
         self.dataset_description = dataset_description
+        self._prefetch_videos = prefetch_videos
         self._recording_idx = 0
         self._synced_recording_cache: dict[int, SynchronizedRecording] = {}
 
+        if prefetch_videos:
+            prefetch_needed = False
+            for rec in self.dataset.recordings:
+                cache = self.dataset.cache_dir / f"{rec['id']}" / f"{self.frequency}Hz"
```
aditya-neuraco: Nit: `cache_dir` would be a better word for this variable.

neuracore/core/data/dataset.py
```diff
@@ -261,7 +266,10 @@ def _synchronize(
         return SyncedDataset.model_validate(dataset_json)
 
     def synchronize(
-        self, frequency: int = 0, data_types: Optional[list[DataType]] = None
+        self,
+        frequency: int = 0,
+        data_types: Optional[list[DataType]] = None,
+        prefetch_videos: bool = False,
```
kwangneuraco, line 272: should we set the default value of prefetch_videos=True?

neuracore/core/data/synced_dataset.py
```diff
@@ -23,6 +26,8 @@ def __init__(
         frequency: int,
         data_types: Optional[list[DataType]],
         dataset_description: DatasetDescription,
+        prefetch_videos: bool = False,
```
kwangneuraco, line 29: same question here

neuracore/core/data/synced_dataset.py
```diff
@@ -79,14 +117,13 @@ def __getitem__(
         """
         if isinstance(idx, slice):
             # Handle slice
-            self.dataset.recordings = self.dataset.recordings[
-                idx.start : idx.stop : idx.step
-            ]
+            dataset = self.dataset[idx.start : idx.stop : idx.step]
```
aditya-neuraco, line 120: This is not equivalent?

Earlier we were slicing the recordings and now we are slicing the dataset.

kwangneuraco (approved):
LGTM

neuracore/core/data/synced_dataset.py
```diff
@@ -79,14 +117,13 @@ def __getitem__(
...
             ds = SynchronizedDataset(
-                dataset=self.dataset,
+                dataset=cast("Dataset", dataset),
                 frequency=self.frequency,
                 data_types=self.data_types,
                 dataset_description=self.dataset_description,
+                prefetch_videos=False,  # Avoid prefetching again
             )
             return ds
```
aditya-neuraco: What does `cast()` do here? Is it for the type-checker to figure out the right types?

NIT: Also, instead of creating a `ds` variable and just returning it on the next line, we should just return the SynchronizedDataset.

neuracore/core/data/synced_dataset.py
stepjam: Discussed offline

neuracore/core/data/synced_dataset.py
stepjam: discussed offline

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+        assert isinstance(synced, SynchronizedRecording)
+        assert synced.frequency == 30
+        assert synced.id == "rec1"
+        assert synced.robot_id == "robot1"
+        assert synced.instance == 1
+
+    def test_synchronize_with_zero_frequency(self, recording):
+        """Test that synchronizing with frequency=0 raises an error."""
+        with pytest.raises(DatasetError, match="Frequency must be greater than 0"):
+            recording.synchronize(frequency=0)
```
aditya-neuraco: Don't we just return the aperiodic data at zero frequency? The `recording` will correctly raise an error to validate this test so the test is correct.

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+
+        assert isinstance(synced, SynchronizedRecording)
+        assert synced.frequency == 30
+        assert synced.id == "rec1"
+        assert synced.robot_id == "robot1"
+        assert synced.instance == 1
+
+    def test_synchronize_with_zero_frequency(self, recording):
+        """Test that synchronizing with frequency=0 raises an error."""
+        with pytest.raises(DatasetError, match="Frequency must be greater than 0"):
```
aditya-neuraco: Instead of a `DatasetError`, doesn't it make more sense to use a `SynchronizationError`?

tests/unit/core/data/test_recording.py
stepjam: Yes but was already there. I agree we should maybe allow this in a future PR.

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+        data_types = [DataType.RGB_IMAGE, DataType.JOINT_POSITIONS]
+        synced = recording.synchronize(frequency=30, data_types=data_types)
+
+        assert isinstance(synced, SynchronizedRecording)
+        assert synced.data_types == data_types
+
+    def test_synchronize_with_invalid_data_types(self, recording):
+        """Test synchronizing with invalid data types raises an error."""
+
+        data_types = [DataType.RGB_IMAGE, DataType.DEPTH_IMAGE]
```
aditya-neuraco, line 76: Invalid DataTypes could also be something like this: `["rgb_images", "depth_image"]. We can definitely check if the data_types are DataType enum values. 

Something like this:

```python
# Check validity of data types (must be DataType enum values)
if not all(isinstance(dt, DataType) for dt in data_types):
    raise ValueError(
        "Invalid data types provided. All items must be DataType enum values."
    )
```

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+
+        assert isinstance(synced, SynchronizedRecording)
+        assert synced.data_types == []
+
+    def test_iter_raises_runtime_error(self, recording):
+        """Test that iterating over unsynchronized recording raises RuntimeError."""
+        with pytest.raises(
+            RuntimeError, match="Only synchronized recordings can be iterated over"
+        ):
+            iter(recording)
```
aditya-neuraco, line 102: `__iter__` always returns a RuntimeError. How does it know the recording is synchronized?

```python
def __iter__(self) -> None:
    """Initialize iterator over synchronized recording data.

    Raises:
        RuntimeError: Always raised to indicate that this method is not
        supported for unsynchronized recordings.
    """
    raise RuntimeError(
        "Only synchronized recordings can be iterated over. "
        "Use the synchronize method to create a synchronized recording."
    )
```

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_iter_in_for_loop_raises_error(self, recording):
+        """Test that using unsynchronized recording in for loop raises error."""
+        with pytest.raises(
+            RuntimeError, match="Only synchronized recordings can be iterated over"
+        ):
+            for _ in recording:
+                pass
```
aditya-neuraco, line 110: Same concern as above.

tests/unit/core/data/test_recording.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_synchronize_different_frequencies(self, recording):
+        """Test synchronizing the same recording at different frequencies."""
+        synced_30hz = recording.synchronize(frequency=30)
+        synced_60hz = recording.synchronize(frequency=60)
+        synced_100hz = recording.synchronize(frequency=100)
+
+        assert synced_30hz.frequency == 30
+        assert synced_60hz.frequency == 60
+        assert synced_100hz.frequency == 100
```
aditya-neuraco: NIT: We can make this into a parametrized test.

neuracore/core/data/dataset.py
```diff
@@ -18,7 +17,7 @@
 from ..exceptions import DatasetError
 from ..nc_types import DataType, SyncedDataset
 
-DEFAULT_CACHE_DIR = Path(tempfile.gettempdir() + "/neuracore_cache")
+DEFAULT_CACHE_DIR = Path.home() / ".neuracore_cache"
```
aditya-neuraco, line 20: I don't think this is a good idea. On servers/clusters, the $HOME folder is usually size limited. It can also have a limit on the number of files. 

For example, when I was using the NYU Cluster for my experiments, all the data needed to be stored elsewhere in the /scratch folder. The home folder quota got depleted really quickly on using conda virtual environments and then it wasn't possible to login to that node.

tests/unit/core/data/test_recording.py
stepjam: Done

tests/unit/core/data/test_recording.py
stepjam: Done anyway

tests/unit/core/data/test_recording.py
stepjam: Allowed zero freq

tests/unit/core/data/test_recording.py
stepjam: Iterating over an unsyched recording is not allowed.

stepjam: All `Recording`s are unsynched

tests/unit/core/data/test_recording.py
stepjam: as above

tests/unit/core/data/test_recording.py
stepjam: Will remove anyway as its not necessary

neuracore/core/data/dataset.py
stepjam: Discussed offline

PR #122, opened 2025-11-07
fix resume checkpoint

kwangneuraco:


neuracore/ml/config/config.yaml
```diff
@@ -38,7 +38,7 @@ local_output_dir: ${oc.env:HOME}/neuracore_logs
 algorithm_params: null
 
 # Resume training
-resume: null  # Path to checkpoint to resume from
+resume_checkpoint: null  # Path to checkpoint to resume from
```
aditya-neuraco: I think a better name would be `resume_checkpoint_path`

aditya-neuraco (approved):
LGTM w/ comments.

PR #123, opened 2025-11-07
Save algorithm config when training locally

ypang-neuraco:
### Bugfixes
When training locally, algorithm_config.json is always empty. This means when the model parameters are modified for training, they are not saved for endpoint deployment. This PR reads the config from the hydra yaml file and save them in the artifact in algorithm_config.json.

### Items
 - [NCRL-25](https://neuracore.atlassian.net/browse/NCRL-25)


stepjam (approved):
LGTM

PR #126, opened 2025-11-07
fix batch size auto tuning

mark-neuracore:
The PR introduces a new dataset type names "DecoyDataset".

This is mainly to speed up the data loading for batch size auto tuning. 

It takes one sample from the training dataset, saves it (to avoid the costly get_sample() call from the real dataset), and returns it on its own get_sample(), which is called by the PyTorch DataLoader.

The Pr introduces some parameters to the Hydra config file (min and max batch sizes and num of workers for batches data loaders)

### DIFF
--- a/neuracore/ml/config/config.yaml
+++ b/neuracore/ml/config/config.yaml
@@ -42,6 +42,8 @@ algorithm_params: null
 # Resume training
 resume_checkpoint_path: null  # Path to checkpoint to resume from
 
+batch_size_autotuning_num_workers: 4 # number of workers to use for batch size autotuning
+
 # Hydra configuration
 hydra:
   run:
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -9,13 +9,13 @@ from typing import Any, Dict, Optional, Tuple, Union
 import hydra
 import torch
 import torch.multiprocessing as mp
-from neuracore_types import DataType, ModelInitDescription
 from omegaconf import DictConfig, OmegaConf
 from torch.utils.data import DataLoader, DistributedSampler, random_split
 
 import neuracore as nc
-from neuracore.core.data.synced_dataset import SynchronizedDataset
+from neuracore.core.nc_types import DataType, ModelInitDescription
 from neuracore.ml import NeuracoreModel
+from neuracore.ml.datasets.pytorch_single_sample_dataset import SingleSampleDataset
 from neuracore.ml.datasets.pytorch_synchronized_dataset import (
     PytorchSynchronizedDataset,
 )
@@ -110,26 +110,18 @@ def convert_data_types(data_types_list: list[str]) -> list[DataType]:
 
 def determine_optimal_batch_size(
     cfg: DictConfig,
-    synchronized_dataset: SynchronizedDataset,
+    dataset: SingleSampleDataset,
     device: Optional[torch.device] = None,
 ) -> int:
     """Run batch size autotuning on a single GPU and return the result."""
+    # if no gpu, raise an error and prevent the auto tuning from starting
+    if not torch.cuda.is_available():
+        raise ValueError("No GPU available. Autotuning is only supported on GPUs.")
+
     logger.info("Starting batch size autotuning on GPU 0...")
 
     input_data_types = convert_data_types(cfg.input_data_types)
     output_data_types = convert_data_types(cfg.output_data_types)
-
-    # Setup dataset for autotuning
-    dataset = PytorchSynchronizedDataset(
-        synchronized_dataset=synchronized_dataset,
-        input_data_types=input_data_types,
-        output_data_types=output_data_types,
-        output_prediction_horizon=cfg.output_prediction_horizon,
-    )
-
-    # Create a smaller subset for autotuning
-    train_size = len(dataset)
-    train_dataset = torch.utils.data.Subset(dataset, list(range(train_size)))
 
     model_init_description = ModelInitDescription(
         dataset_description=dataset.dataset_description,
@@ -142,16 +134,25 @@ def determine_optimal_batch_size(
         cfg, model_init_description, device=device
     )
 
+    max_batch_size = cfg.max_batch_size if "max_batch_size" in cfg else len(dataset)
+    min_batch_size = cfg.min_batch_size if "min_batch_size" in cfg else 2
+    num_workers = cfg.batch_size_autotuning_num_workers
+
+    logger.info(
+        f"using max_batch_size: {max_batch_size}, "
+        f"min_batch_size: {min_batch_size}, "
+        f"num_workers: {num_workers}"
+    )
+
     # Determine per-GPU batch size
     optimal_batch_size = find_optimal_batch_size(
-        dataset=train_dataset,
+        dataset=dataset,
         model=model,
         model_kwargs=algorithm_config,
-        min_batch_size=2,
-        max_batch_size=4096,
-        gpu_id=0,
+        min_batch_size=min_batch_size,
+        max_batch_size=max_batch_size,
         dataloader_kwargs={
-            "num_workers": 4,
+            "num_workers": num_workers,
             "pin_memory": True,
             "persistent_workers": True,
             "collate_fn": dataset.collate_fn,
@@ -167,6 +168,7 @@ def determine_optimal_batch_size(
     logger.info(
         f"Autotuning complete. Optimal batch size per GPU: {optimal_batch_size}"
     )
+
     return optimal_batch_size
 
 
@@ -175,7 +177,7 @@ def run_training(
     world_size: int,
     cfg: DictConfig,
     batch_size: int,
-    synchronized_dataset: SynchronizedDataset,
+    dataset: PytorchSynchronizedDataset,
     device: Optional[torch.device] = None,
 ) -> None:
     """Run the training process for a single GPU."""
@@ -196,14 +198,6 @@ def run_training(
 
         input_data_types = convert_data_types(cfg.input_data_types)
         output_data_types = convert_data_types(cfg.output_data_types)
-
-        # Setup dataset
-        dataset = PytorchSynchronizedDataset(
-            synchronized_dataset=synchronized_dataset,
-            input_data_types=input_data_types,
-            output_data_types=output_data_types,
-            output_prediction_horizon=cfg.output_prediction_horizon,
-        )
 
         # Split dataset
         dataset_size = len(dataset)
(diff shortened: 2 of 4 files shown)


neuracore/ml/train.py
```diff
@@ -392,13 +393,22 @@ def main(cfg: DictConfig) -> None:
     )
 
     # Login and get dataset
+    logger.info("logging in to neuracore...")
```
ypang-neuraco: Please capitalise the first word to be consistent with the rest of the logger messages. Same for the rest

neuracore/ml/train.py
```diff
@@ -166,7 +175,7 @@ def run_training(
     world_size: int,
     cfg: DictConfig,
     batch_size: int,
-    synchronized_dataset: SynchronizedDataset,
+    dataset: PytorchNeuracoreDataset,
```
ypang-neuraco: Why use PytorchNeuracoreDataset instead of PytorchSynchronizedDataset?

neuracore/ml/train.py
```diff
@@ -88,26 +90,18 @@ def convert_data_types(data_types_list: list[str]) -> list[DataType]:
 
 def determine_optimal_batch_size(
     cfg: DictConfig,
-    synchronized_dataset: SynchronizedDataset,
+    dataset: PytorchNeuracoreDataset,
```
ypang-neuraco: Why use PytorchNeuracoreDataset instead of DecoyDataset?

neuracore/ml/config/config.yaml
```diff
@@ -40,6 +40,10 @@ algorithm_params: null
 # Resume training
 resume_checkpoint_path: null  # Path to checkpoint to resume from
 
+max_batch_size: 512 # max batch size to try
+min_batch_size: 2 # min batch size to try
```
ypang-neuraco: find_optimal_batch_size_linear in batch_autotuner.py has a hard coded minimum batch size of 8. Should we lower that to 2 as well since we have now sped up the autotuner?

neuracore/ml/train.py
```diff
@@ -88,26 +90,18 @@ def convert_data_types(data_types_list: list[str]) -> list[DataType]:
...
 ) -> int:
     """Run batch size autotuning on a single GPU and return the result."""
+    # if no gpu, raise an error and prevent the finetuning from starting
+    if not torch.cuda.is_available():
+        raise ValueError("No GPU available. Finetuning is only supported on GPUs.")
+
```
ypang-neuraco: Finetuning is a bit misleading, I think it should be renamed to batch size autotuning.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -101,7 +103,7 @@ def find_optimal_batch_size(self) -> int:
             gc.collect()
 
         # Reduce by 15% to be safe
```
ypang-neuraco, line 104: Update comment to reflect the new limit

neuracore/ml/train.py
```diff
@@ -133,16 +127,30 @@ def determine_optimal_batch_size(
             "must be provided in the configuration"
         )
 
+    max_batch_size = cfg.max_batch_size if "max_batch_size" in cfg else 512
+    min_batch_size = cfg.min_batch_size if "min_batch_size" in cfg else 2
```
ypang-neuraco, line 138: Should we align the default of find_optimal_batch_size (8) to the same as this?

neuracore/ml/train.py
```diff
@@ -417,24 +427,47 @@ def main(cfg: DictConfig) -> None:
...
+        input_data_types=input_data_types,
+        output_data_types=output_data_types,
+        output_prediction_horizon=cfg.output_prediction_horizon,
+    )
+    logger.info("finished creating pytorch synchronized dataset")
+
     # Handle batch size configuration
     if isinstance(batch_size, str) and batch_size.lower() == "auto":
-        optimal_batch_size = determine_optimal_batch_size(cfg, synchronized_dataset)
+        logger.info("starting batch size autotuning...")
```
ypang-neuraco: Do we need these messages? We have plenty of logging in batch_autotuner.py already

neuracore/ml/train.py
```diff
@@ -417,24 +427,47 @@ def main(cfg: DictConfig) -> None:
...
+    logger.info("creating a pytorch synchronized dataset...")
+    input_data_types = convert_data_types(cfg.input_data_types)
+    output_data_types = convert_data_types(cfg.output_data_types)
+    pytorch_dataset = PytorchSynchronizedDataset(
+        synchronized_dataset=synchronized_dataset,
+        input_data_types=input_data_types,
+        output_data_types=output_data_types,
+        output_prediction_horizon=cfg.output_prediction_horizon,
+    )
+    logger.info("finished creating pytorch synchronized dataset")
```
ypang-neuraco: I think we don't need these messages as we have the synchronization pbar already

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/config/config.yaml
stepjam: To add to this, I'm not sure if we want to be making this parameterised. Batch size auto tuner, by definition, should find us the best batch size. I'm fine with us having min max to bound it,  ut I think they should be hard-coded constants, and be much larger than 512

neuracore/ml/datasets/pytorch_decoy_dataset.py
```diff
@@ -0,0 +1,43 @@
+"""Dataset that returns the same sample from a real dataset for quick testing."""
+
+from neuracore.ml import BatchedTrainingSamples
+from neuracore.ml.datasets.pytorch_neuracore_dataset import PytorchNeuracoreDataset
+
+
+class DecoyDataset(PytorchNeuracoreDataset):
```
stepjam: Please find better name than `DecoyDataset`. How about `OnlyFirstSampleDataset` or even  `BatchAutoTuneDataset`.

neuracore/ml/datasets/pytorch_decoy_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+        super().__init__(
+            input_data_types=dataset.input_data_types,
+            output_data_types=dataset.output_data_types,
+            output_prediction_horizon=dataset.output_prediction_horizon,
+            num_recordings=self.num_samples,
+        )
+        self.dataset_description = dataset.dataset_description
+
+        # Create a template sample from the first sample of the dataset
+        self.sample = dataset[0]
```
stepjam: please make these private member variables `self._`

neuracore/ml/train.py
stepjam: +1

neuracore/ml/train.py
```diff
@@ -133,16 +127,30 @@ def determine_optimal_batch_size(
...
     # Determine per-GPU batch size
     optimal_batch_size = find_optimal_batch_size(
-        dataset=train_dataset,
+        dataset=dataset,
         model=model,
         model_kwargs=algorithm_config,
-        min_batch_size=2,
-        max_batch_size=4096,
+        min_batch_size=min_batch_size,
+        max_batch_size=max_batch_size,
```
stepjam, line 153: As notes in cfg, let's have these as constants at the top of the file.

neuracore/ml/datasets/pytorch_decoy_dataset.py
```diff
@@ -0,0 +1,43 @@
...
+class DecoyDataset(PytorchNeuracoreDataset):
+    """Fast dataset wrapper that loads and saves the first sample from a real dataset.
+
+    It saves this sample to avoid costly loading of the samples
+    every time __getitem__ or load_sample is called.
+    """
+
+    def __init__(
+        self,
+        dataset: PytorchNeuracoreDataset,
```
stepjam: Why not just have this extend `PytorchSynchronizedDataset`

neuracore/ml/train.py
stepjam: I'd suggest just removing this log.

neuracore/ml/train.py
```diff
@@ -392,13 +393,22 @@ def main(cfg: DictConfig) -> None:
...
+    logger.info(
+        f"getting dataset with id: {cfg.dataset_id} and/or name: {cfg.dataset_name}..."
+    )
     if cfg.dataset_id is not None:
         dataset = nc.get_dataset(id=cfg.dataset_id)
     elif cfg.dataset_name is not None:
         dataset = nc.get_dataset(name=cfg.dataset_name)
+
+    logger.info(f"synchronizing dataset with frequency: {cfg.frequency}")
     synchronized_dataset = dataset.synchronize(
```
stepjam: Please remove all logs

neuracore/ml/train.py
mark-neuracore: Both DecoyDataet and PytrochSynchronizedDataset inherit from PytorchNeuracoreDataset. The decision to use the parent class as the type here is to say that this method will work with either the newly introduced DecoyDataset or the PytrochSynchronizedDataset, should we ever need to use it again. 

The only difference is that Decoydataset __getitem__ method (or in our case get_sample() ) returns the same item other than actually sampling the dataset for a real sample, which doesn't affect how this method works at all.

neuracore/ml/train.py
mark-neuracore: I think this was more helpful to me when i was printing the timing each section was taking before using cProfile. can remove

neuracore/ml/train.py
mark-neuracore: I think i just mixed up naming there. will change.

neuracore/ml/train.py
```diff
@@ -133,16 +126,30 @@ def determine_optimal_batch_size(
             "must be provided in the configuration"
         )
 
+    max_batch_size = cfg.max_batch_size if "max_batch_size" in cfg else len(dataset)
+    min_batch_size = cfg.min_batch_size if "min_batch_size" in cfg else 2
+    num_workers = (
+        cfg.batch_size_autotuning_num_workers
+        if "batch_size_autotuning_num_workers" in cfg
+        else 4
+    )
```
stepjam: why not just:
```
num_workers =  cfg.batch_size_autotuning_num_workers
```

The cfg already has the default value set.

neuracore/ml/train.py
```diff
@@ -309,7 +309,7 @@ def run_training(
             f"{sum(p.numel() for p in model.parameters()):,} parameters"
         )
 
-        training_logger: Union[TensorboardTrainingLogger, CloudTrainingLogger]
+        training_logger: TensorboardTrainingLogger | CloudTrainingLogger
```
stepjam: Please make sure to use >= Python 3.9. `|` was introduced in Py 3.10

neuracore/ml/train.py
```diff
@@ -417,24 +419,52 @@ def main(cfg: DictConfig) -> None:
...
+        )
+
+        optimal_batch_size = determine_optimal_batch_size(cfg, single_sample_dataset)
         batch_size = optimal_batch_size
+
+        # Clean up
+        del single_sample_dataset
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
```
stepjam: Shouldn't this cleanup happen inside the batch auto tuner? 
Also:
```
del single_sample_dataset
gc.collect()
```
Shouldn't be necessary

stepjam (approved):
LGTM w/ comments.
Please also update the PR message before merging in

PR #128, opened 2025-11-09
Misc training improvements

stepjam:
### Features
 - Improve device management
 - Only keep last N checkpoints
 - Remove tensor cache

### DIFF
--- a/neuracore/ml/algorithms/cnnmlp/cnnmlp.py
+++ b/neuracore/ml/algorithms/cnnmlp/cnnmlp.py
@@ -49,7 +49,7 @@ class CNNMLP(NeuracoreModel):
         model_init_description: ModelInitDescription,
         device: torch.device,
         hidden_dim: int = 512,
-        cnn_output_dim: int = 64,
+        cnn_output_dim: int = 512,
         num_layers: int = 3,
         lr: float = 1e-4,
         lr_backbone: float = 1e-5,
--- a/neuracore/ml/algorithms/cnnmlp/modules.py
+++ b/neuracore/ml/algorithms/cnnmlp/modules.py
@@ -19,6 +19,8 @@ class ImageEncoder(nn.Module):
     vision tasks requiring fixed-size feature representations.
     """
 
+    OUTPUT_CNN_W_H = 7 * 7
+
     def __init__(self, output_dim: int = 512, backbone: str = "resnet18"):
         """Initialize the image encoder.
 
@@ -29,7 +31,8 @@ class ImageEncoder(nn.Module):
         super().__init__()
         # Use pretrained ResNet but remove final layer
         self.backbone = self._build_backbone(backbone)
-        self.proj = nn.Linear(512, output_dim)
+        self.proj_2d = nn.Conv2d(512, output_dim, kernel_size=1)
+        self.linear = nn.Linear(ImageEncoder.OUTPUT_CNN_W_H * output_dim, output_dim)
 
     def _build_backbone(self, backbone_name: str) -> nn.Module:
         """Build backbone CNN by removing classification layers.
@@ -41,7 +44,7 @@ class ImageEncoder(nn.Module):
             nn.Module: ResNet backbone without final classification layers
         """
         resnet = getattr(models, backbone_name)(pretrained=True)
-        return nn.Sequential(*list(resnet.children())[:-1])
+        return nn.Sequential(*list(resnet.children())[:-2])
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """Forward pass through image encoder.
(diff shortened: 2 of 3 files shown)


neuracore/api/endpoints.py
```diff
@@ -41,6 +42,7 @@ def policy(
         train_run_name: Name of the training run to load the model from.
         model_file: Path to the model file to load. If provided, overrides
         output_mapping: Optional mapping of output data types to response fields.
+        device: Model to be loaded on
```
ypang-neuraco: Please change to "Torch device to run the model on (CPU or GPU, or MPS).", same for policy_local_server.

neuracore/core/endpoint.py
```diff
@@ -511,7 +514,7 @@ def policy(
         robot_name: Robot identifier.
         instance: Instance number of the robot.
         output_mapping: Optional mapping of data types to output keys.
-
+        device: Model to be loaded on.
```
ypang-neuraco: Please change to "Torch device to run the model on (CPU or GPU, or MPS)."

neuracore/api/endpoints.py
```diff
@@ -31,6 +31,7 @@ def policy(
     train_run_name: Optional[str] = None,
     model_file: Optional[str] = None,
     output_mapping: Optional[dict[DataType, list[str]]] = None,
+    device: Optional[str] = None,
```
ypang-neuraco, line 34: Should this be changed to torch.device?

neuracore/core/endpoint.py
```diff
@@ -133,6 +133,7 @@ def __init__(
         org_id: str,
         job_id: Optional[str] = None,
         output_mapping: Optional[dict[DataType, list[str]]] = None,
+        device: Optional[str] = None,
```
ypang-neuraco, line 136: Should this be changed to torch.device?

neuracore/core/endpoint.py
```diff
@@ -503,6 +505,7 @@ def policy(
     train_run_name: Optional[str] = None,
     model_file: Optional[str] = None,
     output_mapping: Optional[dict[DataType, list[str]]] = None,
+    device: Optional[str] = None,
```
ypang-neuraco, line 508: Should this be changed to torch.device?

neuracore/ml/utils/validate.py
```diff
@@ -419,7 +422,7 @@ def run_validation(
                     policy = nc.policy_local_server(
                         model_file=str(artifacts_dir / "model.nc.zip"),
                         port=port,
-                        device=device,
+                        device=str(device),
```
ypang-neuraco, line 426: Why not use the torch device directly?

neuracore/ml/datasets/pytorch_synchronized_dataset.py
ypang-neuraco, line 1: I think it is good that we avoid double caching the dataset. Does this affect the training speed much?

neuracore/ml/trainers/distributed_trainer.py
```diff
@@ -78,11 +80,16 @@ def __init__(
...
             world_size: Total number of processes/GPUs
+            device: Optional device to use for training
         """
-        self.device = torch.device(
+        if keep_last_n_checkpoints <= 0:
+            raise ValueError("keep_last_n_checkpoints must be greater than 0")
+
+        self.device = device or torch.device(
             f"cuda:{rank}" if torch.cuda.is_available() else "cpu"
         )
```
ypang-neuraco: We have repeated this logic many times. Should we make this a function somewhere?

neuracore/ml/datasets/pytorch_synchronized_dataset.py
stepjam: I dd some prelim timing and it was very small cost (~0-5% slower perhaps)

neuracore/core/endpoint.py
stepjam: It has to be a str because only /neuracore/ml folder can have any reference to pytroch. Otherwise, people would have to install pytroch even if they only want to use the core logging functionality.

neuracore/ml/utils/validate.py
stepjam: It has to be a str because only /neuracore/ml folder can have any reference to pytroch. Otherwise, people would have to install pytroch even if they only want to use the core logging functionality.

neuracore/core/endpoint.py
stepjam: It has to be a str because only /neuracore/ml folder can have any reference to pytroch. Otherwise, people would have to install pytroch even if they only want to use the core logging functionality.

neuracore/api/endpoints.py
stepjam: It has to be a str because only /neuracore/ml folder can have any reference to pytroch. Otherwise, people would have to install pytroch even if they only want to use the core logging functionality.

neuracore/ml/trainers/distributed_trainer.py
stepjam: done!

PR #130, opened 2025-11-11
Wait for frame to decode when not prefetching videos

ypang-neuraco:
### Bugfixes
Following https://neuracore.atlassian.net/browse/NCRL-10 , there can be a race condition when prefetch_videos=False. One thread may start to download, the cache  dir is crated, but at the same time another thread will check if the dir exists, it does, but the video has not finished decoding on the other thread. This will mean that we get a FileNotFound error. 

Changes:
- Check if frame is decoded before getting the frame, if not we wait and retry
- Set max attempts to check if decoding is progressing

### Items
 - [NCRL-24](https://neuracore.atlassian.net/browse/NCRL-24)

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -3,6 +3,7 @@
 import copy
 import logging
 import tempfile
+import time
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable, Optional, Union, cast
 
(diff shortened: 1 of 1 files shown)


neuracore/core/data/synced_recording.py
```diff
@@ -167,13 +170,36 @@ def _get_frame_from_disk_cache(
...
+                            f"Decoding timed out for recording {self.id}"
+                        )
+                else:
+                    last_num_frames = current_num_frames
+                    attempts_left = MAX_DECODING_ATTEMPTS
+
+                # Wait for decoding to progress and try again
+                time.sleep(5)
+
+            frame = Image.open(frame_file)
```
stepjam: As discussed, wrap in try catch. Worth adding a comment why we are try catching it.

PR #131, opened 2025-11-11
Fix synced data num transition from total length

mark-neuracore:
based on the changes from feat/total_length in both `neuracore` and `neuracore_backend`:
- changes in neuracore adding total_length in DatasetDescription pydantic model
- changes in neuracore_backend summing all transitions in each recording in a dataset into total_length in the Datasetdescription

 We can use total_length from the dataset description to get the total number of transitions in a dataset, which makes initialising a pytorch synchronised dataset much faster (~1-2 mins -> 0.0001 secs)

### DIFF
--- a/neuracore/core/data/synced_dataset.py
+++ b/neuracore/core/data/synced_dataset.py
@@ -79,7 +79,7 @@ class SynchronizedDataset:
     @property
     def num_transitions(self) -> int:
         """Get the number of transitions in the dataset."""
-        return self.dataset_description.total_length
+        return self.dataset_description.total_num_transitions
 
     def __iter__(self) -> "SynchronizedDataset":
         """Initialize iterator over episodes in the dataset.


neuracore/core/nc_types.py
```diff
@@ -524,6 +524,9 @@ class DatasetDescription(BaseModel):
     for determining which data types are present.
     """
 
+    # Total number of sync points
```
aditya-neuraco: I think we should change it to `# Total number of sync points in the dataset"`

neuracore/core/nc_types.py
```diff
@@ -524,6 +524,9 @@ class DatasetDescription(BaseModel):
...
+    total_length: int = 0
```
aditya-neuraco: I think `total_num_sync_points` or `total_num_transitions` or `total_num_samples` will be a better name because IMO length of the dataset is not clear. Someone could interpret the length of dataset as it as `number of episodes or recordings`.

stepjam:
> @stepjam this now requires neuracore_type's DatasetDescription to have total_num_transitions and required neuracore_backend to calculate and populate total number of transitions into this field during synchronisation. We discussed this and said we will have to clear all the cache of synchronised datasets once we merge here.

This has now been done for dev.

PR #135, opened 2025-11-12
fix: remove device parameter in model initialization

aditya-neuraco:
### Updates
- Removed device parameter from the constructor of all model classes.
- Updated tests to explicitly move model to device.
- Updated state statistics to buffers so that they can move automatically to device after using the `.to(device)` function.

### Items
 - [NC-42](https://neuracore.atlassian.net/browse/NC-42)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_install_hook_types: [pre-commit]
-default_stages: [commit]
+default_stages: [pre-commit]
 
 repos:
   - repo: https://github.com/pycqa/isort
@@ -30,7 +30,7 @@ repos:
         args: ["--convention=google"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: ""
+    rev: v1.18.2
     hooks:
       - id: mypy
         files: neuracore/
--- a/examples/example_local_endpoint.py
+++ b/examples/example_local_endpoint.py
@@ -77,6 +77,11 @@ def main():
             if done:
                 print(f"Episode {episode_idx} done")
                 break
+        if reward == 4:
+            print(f"Episode {episode_idx} successful.")
+        else:
+            print(f"Episode {episode_idx} failed.")
+
         plt.close()
 
     policy.disconnect()
--- a/neuracore/core/auth.py
+++ b/neuracore/core/auth.py
@@ -10,6 +10,7 @@ import os
 from typing import Optional
 
 import requests
+from neuracore_types import __version__ as nc_types_version
 
 from neuracore.core.config.config_manager import get_config_manager
 from neuracore.core.config.get_api_key import get_api_key
@@ -108,11 +109,10 @@ class Auth(metaclass=SingletonMetaclass):
                 incompatible versions or server communication issues.
         """
         # Placeholder for version validation logic
-        import neuracore as nc
 
         response = requests.get(
             f"{API_URL}/auth/verify-version",
-            params={"version": nc.__version__},
+            params={"version": nc_types_version},
         )
         if response.status_code != 200:
             raise AuthenticationError(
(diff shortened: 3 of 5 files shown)


neuracore/ml/algorithms/act/act.py
```diff
@@ -164,24 +161,46 @@ def __init__(
...
+            self.register_buffer(
+                "joint_state_mean", self._to_torch_float_tensor(state_means)
+            )
+            self.register_buffer(
+                "joint_state_std", self._to_torch_float_tensor(state_stds)
+            )
+        else:
+            self.joint_state_mean = None
+            self.joint_state_std = None
+        self.register_buffer(
```
kwangneuraco, line 193: would be better if we have action_means and action_stds? Because later we will have ee pose as action as well

neuracore/ml/algorithms/cnnmlp/cnnmlp.py
```diff
@@ -209,8 +207,12 @@ def _setup_normalization_stats(self) -> None:
             state_stds.extend(self.dataset_description.joint_torques.std)
 
         if state_means:
-            self.joint_state_mean = self._to_torch_float_tensor(state_means)
-            self.joint_state_std = self._to_torch_float_tensor(state_stds)
+            self.register_buffer(
```
kwangneuraco, line 210: does cnnmlp not have setup_normalization?

aditya-neuraco: It was already present in CNNMLP.

neuracore/ml/algorithms/act/act.py
aditya-neuraco: Yes, can do. I didn't rename it for consistency.

Can do a separate PR where we change the variable names across all models.

neuracore/ml/core/neuracore_model.py
```diff
@@ -49,12 +47,28 @@ def __init__(
...
 
+    @property
+    def device(self) -> torch.device:
+        """Get the device for the model.
+
+        Returns:
+            torch.device: The device for the model
+        """
+        try:
+            return next(self.parameters()).device
```
kwangneuraco, line 63: Is there a better way to get device? This doesn't seems to me a optimal way

aditya-neuraco: I'll have to see. This sets a default device on which all tensors and buffers are initialized on. `.to(device)` then correctly moves it to the requested device. I have checked with both CPU (by disabling my NVIDIA GPU) and GPU.

kwangneuraco (approved):
LGTM!

neuracore/ml/core/neuracore_model.py
ypang-neuraco: I like this as it get the actual device the parameters are on. Alternatively, is there anyway to save the device when we call model.to(device)?

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/core/neuracore_model.py
aditya-neuraco: > Alternatively, is there anyway to save the device when we call model.to(device)?

Actually we don't need to save it at all. `next(model.parameters()).device` returns the device the model is on.

```python
Python 3.10.18 (main, Sep 18 2025, 19:50:32) [Clang 20.1.4 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> import torchvision
>>> resnet = torchvision.models.get_model("resnet18", weights="DEFAULT")
>>> next(resnet.parameters()).device
device(type='cpu')
>>> resnet = resnet.to("cuda")
>>> next(resnet.parameters()).device
device(type='cuda', index=0)
```

PR #136, opened 2025-11-12
fix: removed use of deprecated `pretrained=True` parameter.

aditya-neuraco:
### Updates
- Removed deprecated parameter when instantiating resnet backbones.

### Items

 - [NC-28](https://neuracore.atlassian.net/browse/NC-28)

### DIFF
--- a/neuracore/ml/algorithm_utils/modules.py
+++ b/neuracore/ml/algorithm_utils/modules.py
@@ -29,7 +29,7 @@ class DepthImageEncoder(nn.Module):
         """
         super().__init__()
         # Modify ResNet for single-channel input
-        resnet = getattr(models, backbone)(pretrained=True)
+        resnet = models.get_model(backbone, weights="DEFAULT")
 
         # Replace first conv layer for single-channel depth input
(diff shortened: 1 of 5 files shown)


neuracore/ml/algorithm_utils/modules.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, output_dim: int = 256, backbone: str = "resnet18"):
         """
         super().__init__()
         # Modify ResNet for single-channel input
-        resnet = getattr(models, backbone)(pretrained=True)
+        resnet = models.get_model(backbone, weights="DEFAULT")
```
kwangneuraco, line 32: I need to double check, i think the weights are not DEFAULT

aditya-neuraco: By DEFAULT do you mean the weights are not same or the keyword is not DEFAULT?

kwangneuraco: After discussion it is good now

kwangneuraco (approved):
LGTM!

PR #141, opened 2025-11-13
changed old neuracore core types to neuracore_types

mark-neuracore:
removing old neuracore types and replacing them with neuracore_types


stepjam (approved):
LGTM

PR #143, opened 2025-11-16
Add min max normalization diffusion policy

kwangneuraco:
### Features

- Add min_max normalization in diffusion policy
- Modified several hyper-parameters to get better performance
- Modified general config.yaml to match training config in frontend and our default training setting up.
- Add output_types check for normalization in other methods

Related to ticket
[NCRL-47](https://neuracore.atlassian.net/browse/NCRL-47?atlOrigin=eyJpIjoiMGM5OGZiMDM3ZWQxNGY5ZGFhNzA1NGFmMTUxMjdlNmMiLCJwIjoiaiJ9)

### DIFF
--- a/neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
+++ b/neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
@@ -36,13 +36,14 @@ class DiffusionPolicy(NeuracoreModel):
         model_init_description: ModelInitDescription,
         hidden_dim: int = 256,
         unet_down_dims: Tuple[int, ...] = (
+            256,
             512,
             1024,
-            2048,
         ),
         unet_kernel_size: int = 5,
         unet_n_groups: int = 8,
         unet_diffusion_step_embed_dim: int = 128,
+        spatial_softmax_num_keypoints: int = 32,
         unet_use_film_scale_modulation: bool = True,
         noise_scheduler_type: str = "DDPM",
         num_train_timesteps: int = 100,
@@ -54,8 +55,9 @@ class DiffusionPolicy(NeuracoreModel):
         clip_sample_range: float = 1.0,
         lr: float = 1e-4,
         lr_backbone: float = 1e-5,
-        weight_decay: float = 1e-4,
+        weight_decay: float = 1e-6,
         prediction_type: str = "epsilon",
+        normalization_type: str = "min_max",
     ):
         """Initialize the Diffusion Policy model.
 
@@ -66,6 +68,7 @@ class DiffusionPolicy(NeuracoreModel):
             unet_kernel_size: Kernel size for UNet convolutions.
             unet_n_groups: Number of groups for group normalization.
             unet_diffusion_step_embed_dim: Dimension of diffusion step embeddings.
+            spatial_softmax_num_keypoints: Number of keypoints for spatial softmax.
             unet_use_film_scale_modulation: Whether to use FiLM scale modulation.
             noise_scheduler_type: Type of noise scheduler ("DDPM" or "DDIM").
             num_train_timesteps: Number of timesteps for training.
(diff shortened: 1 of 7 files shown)


neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
```diff
@@ -183,17 +237,67 @@ def _to_torch_float_tensor(self, data: list[float]) -> torch.FloatTensor:
...
+            range_val = stat2 - stat1
+            # Avoid division by zero
+            range_val = torch.clamp(range_val, min=1e-8)
+            # Scale to [0, 1], then map to [-1, 1]
+            normalized = (joint_state - stat1) / range_val
+            return normalized * 2.0 - 1.0
+        else:
+            raise ValueError(
+                f"Unsupported normalization_type: {self.normalization_type}"
+            )
```
aditya-neuraco, line 277: We can pass the `dataset_description` directly and read stats from there. The code would be cleaner.

neuracore/ml/algorithms/pi0/pi0.py
```diff
@@ -215,18 +215,26 @@ def _setup_normalization_stats(self) -> None:
...
+            )
+            self.register_buffer(
+                "joint_target_std",
+                self._to_torch_float_tensor(
+                    self.dataset_description.joint_target_positions.std
+                ),
+            )
+        else:
+            self.joint_target_mean = None
+            self.joint_target_std = None
```
aditya-neuraco: This could be in a different PR? It's changing files in `Pi0` but this PR is seems mainly about Diffusion Policy.

neuracore/ml/algorithms/simple_vla/simple_vla.py
```diff
@@ -162,18 +162,26 @@ def _setup_normalization_stats(self) -> None:
...
+            )
+            self.register_buffer(
+                "joint_target_std",
+                self._to_torch_float_tensor(
+                    self.dataset_description.joint_target_positions.std
+                ),
+            )
+        else:
+            self.joint_target_mean = None
+            self.joint_target_std = None
```
aditya-neuraco: This could be in a different PR? It's changing files in `SimpleVLA` but this PR is seems mainly about Diffusion Policy.

neuracore/ml/config/algorithm/diffusion_policy.yaml
```diff
@@ -2,20 +2,22 @@
...
   beta_end: 0.02
   beta_schedule: "squaredcos_cap_v2"
   clip_sample: true
   clip_sample_range: 1.0
   lr: 1e-4
   lr_backbone: 1e-5
-  weight_decay: 1e-4
+  weight_decay: 1e-6
   prediction_type: "epsilon"
+  normalization_type: "min_max"  # Options: "mean_std" or "min_max"
```
aditya-neuraco: Don't need a comment here as we are not commenting the choices for other options.

neuracore/ml/config/config.yaml
```diff
@@ -5,8 +5,8 @@ defaults:
 
 # Core training parameters
 seed: 42
-epochs: 100
-output_prediction_horizon: 10
+epochs: 10
+output_prediction_horizon: 100
```
aditya-neuraco: I think we should keep the default number of epochs as `100` because `10` epochs won't lead to acceptable performance on any dataset/algorithm.

neuracore/ml/algorithms/pi0/pi0.py
kwangneuraco: ok sure

neuracore/ml/algorithms/diffusion_policy/diffusion_policy.py
kwangneuraco: Agree, we should have a dataset_stats dictionary with all the elements (mean, std, etc), and setup_normalization function just returns this. I would rather put this into a separate PR in which the normalization and un-normalization are properly handled and as a helper function.

PR #149, opened 2025-11-19
fix: add comprehensive tests for train.py

aditya-neuraco:
- Added tests for train.py

### DIFF
--- a/tests/unit/ml/test_train.py
+++ b/tests/unit/ml/test_train.py
@@ -1,0 +1,1577 @@
+"""Tests for train.py training script.
+
+This module provides comprehensive testing for the training script functionality
+including logging setup, model configuration, data type conversion, batch size
+autotuning, and training execution.
+"""
+
+import gc
+import logging
+import os
+import tempfile
+from pathlib import Path
+from unittest.mock import Mock
+
+import pytest
+import torch
+from neuracore_types import (
+    DataItemStats,
+    DatasetDescription,
+    DataType,
+    ModelInitDescription,
+    ModelPrediction,
+)
+from omegaconf import OmegaConf
+
+from neuracore.ml import BatchedTrainingOutputs, NeuracoreModel
+from neuracore.ml.datasets.pytorch_single_sample_dataset import SingleSampleDataset
+from neuracore.ml.datasets.pytorch_synchronized_dataset import (
+    PytorchSynchronizedDataset,
+)
+from neuracore.ml.train import (
+    convert_data_types,
+    determine_optimal_batch_size,
+    get_model_and_algorithm_config,
+    main,
+    run_training,
+    setup_logging,
+)
+
+SKIP_TEST = os.environ.get("CI", "false").lower() == "true"
+
+
+class MainTestSetup:
+    def __init__(self, monkeypatch, cuda_device_count=1):
+        self.monkeypatch = monkeypatch
+        self.cuda_device_count = cuda_device_count
+
+        self.mock_dataset = Mock()
+        self.mock_synchronized_dataset = Mock()
+        self.mock_pytorch_dataset = Mock(spec=PytorchSynchronizedDataset)
+        self.mock_pytorch_dataset.dataset_description = Mock()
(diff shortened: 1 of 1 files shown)


tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+
+    mock_instantiate = Mock(return_value=mock_model_class(model_init_description))
+    monkeypatch.setattr("neuracore.ml.train.hydra.utils.instantiate", mock_instantiate)
+
+    model, algorithm_config = get_model_and_algorithm_config(
+        cfg, model_init_description
+    )
+
+    assert isinstance(model, NeuracoreModel)
+    mock_instantiate.assert_called_once()
```
ypang-neuraco: Should this raise an error instead? Similar to the logic in main

tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+        model_init_description,
+        mock_model_class,
+    )
+    setup.mock_trainer.train.side_effect = RuntimeError("Training failed")
+    setup.setup_mocks()
+
+    with pytest.raises(RuntimeError, match="Training failed"):
+        setup.call_run_training(mock_cfg_training, mock_dataset)
+
+    setup.mock_cleanup.assert_not_called()
```
ypang-neuraco: Shouldn't cleanup always run?

tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+        "Checkpoint not found"
+    )
+    setup.setup_mocks()
+
+    setup.call_run_training(mock_cfg_training, mock_dataset)
+
+    setup.mock_trainer.load_checkpoint.assert_called_once_with(
+        "/path/to/checkpoint.pth"
+    )
+    setup.mock_trainer.train.assert_called_once_with(start_epoch=0)
```
ypang-neuraco: I think training should not run if load_checkpoint fails. If user expects to continue training from checkpoint, and loading fails but training runs anyway from epoch 0,  that means compute is consumed on something that the user does not intend to do.

tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+    monkeypatch.setattr("neuracore.ml.train.hydra.utils.instantiate", mock_instantiate)
+
+    model, algorithm_config = get_model_and_algorithm_config(
+        cfg, model_init_description
+    )
+
+    assert isinstance(model, NeuracoreModel)
+    assert "_target_" not in algorithm_config
+    assert algorithm_config == {"param1": "value1", "param2": "value2"}
+    mock_instantiate.assert_called_once()
```
ypang-neuraco: Maybe move it earlier where the other tests for this function is.

tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+        "output_prediction_horizon": 5,
+        "frequency": 30,
+    })
+
+    setup = MainTestSetup(monkeypatch)
+    setup.setup_mocks()
+
+    main(cfg)
+
+    setup.mock_get_dataset.assert_called_once()
```
ypang-neuraco: use get_model_and_algorithm_config instead?

tests/unit/ml/test_train.py
```diff
@@ -0,0 +1,1218 @@
...
+        "frequency": 30,
+        "algorithm_params": None,
+    })
+
+    setup = MainTestSetup(monkeypatch)
+    setup.setup_mocks()
+
+    main(cfg)
+
+    setup.mock_get_dataset.assert_called_once()
```
ypang-neuraco: use get_model_and_algorithm_config instead?

ypang-neuraco (approved):
LGTM \w comments

tests/unit/ml/test_train.py
aditya-neuraco: We provide feedback in logs that we failed to load checkpoint so I think this is fine.

PR #151, opened 2025-11-19
feat: move simple VLA to examples and remove simple world model

aditya-neuraco:
### Features
 - Moved SimpleVLA to examples to showcase custom algorithm development.
 - Removed simple world model, configs ans test.

### Items

 - [NCRL-53](https://neuracore.atlassian.net/browse/NCRL-53)


tests/unit/examples/test_simple_vla.py
```diff
@@ -20,11 +21,14 @@
     BatchedTrainingSamples,
     MaskableData,
 )
-from neuracore.ml.algorithms.simple_vla.simple_vla import SimpleVLA
 from neuracore.ml.core.ml_types import BatchedData
 from neuracore.ml.utils.device_utils import get_default_device
 from neuracore.ml.utils.validate import run_validation
 
+sys.path.append(str(Path(__file__).resolve().parent.parent.parent.parent))
```
stepjam: This is really hard to parse

tests/unit/examples/test_simple_vla.py
```diff
@@ -20,11 +21,14 @@
...
+
+from examples.example_custom_algorithm import SimpleVLA  # noqa: E402
```
stepjam: What if we move the the algorithms to a folder examples/ml/

As it would hint that the user needs to pip install the package with the [ml] optional in order for it to work

examples/__init__.py
stepjam, line 1: Should examples be a module?

tests/unit/examples/test_simple_vla.py
aditya-neuraco: I agree. Changing it to `sys.path.append(str(Path(__file__).resolve().parents[3]`

examples/__init__.py
aditya-neuraco: Removing. Tests work without it anyway.

tests/unit/examples/test_simple_vla.py
aditya-neuraco: Good suggestion, I'll make the change.

PR #155, opened 2025-11-24
Split dataset into train and validation sets on the frame level

ypang-neuraco:
### Bugfixes
When splitting the dataset into train/val splits, we get the list of all transitions, and randomly sample. However, when we actually get the sample, we get a random episode and a random frame, which means train/val can contain the same frames.

In this PR:
- Correctly sample on the frame level (no overlap between train/val)
- Always pre-fetch synced meta data (need episode length to correctly map sample idx to episode idx)
- Move __getitem__ to each dataset.py file

### Items
 - [NCRL-56](https://neuracore.atlassian.net/browse/NCRL-56)

### DIFF
--- a/neuracore/ml/datasets/pytorch_synchronized_dataset.py
+++ b/neuracore/ml/datasets/pytorch_synchronized_dataset.py
@@ -77,8 +77,25 @@ class PytorchSynchronizedDataset(PytorchNeuracoreDataset):
             max_ram_utilization=0.8, max_gpu_utilization=1.0, gpu_id=None
         )
         self._mem_check_counter = 0
-        self._num_samples = self.synchronized_dataset.num_transitions
+        self._num_samples_excluding_last = (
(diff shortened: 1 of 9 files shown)


aditya-neuraco (approved):
LGTM!

neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -78,8 +78,23 @@ def __init__(
         )
         self._mem_check_counter = 0
         self._num_samples = self.synchronized_dataset.num_transitions
```
stepjam: Let's stopre the value in the init:
```
self._num_samples = self.synchronized_dataset.num_transitions - len(self.synchronized_dataset)
```
And also, maybe lets choose a better name than num_samples. E.g. `num_samples_excluding_last` or similar

PR #156, opened 2025-11-24
remove the resize of image at the dataloader and pre-processing 

kwangneuraco:
Remove the resize of image to [224, 224] in pytorch dataloader and policy inference processing as it is already done in the algorithm.

Related to [NCRL-91](https://neuracore.atlassian.net/browse/NCRL-91?atlOrigin=eyJpIjoiNmE1MjVlYTdiZGJkNDExNjgwZjU3MDFiNzRlOWY3ZTEiLCJwIjoiaiJ9)

### DIFF
--- a/neuracore/core/utils/image_string_encoder.py
+++ b/neuracore/core/utils/image_string_encoder.py
@@ -7,7 +7,7 @@ This is useful for transmitting image data over networks with other data.
 
 import base64
 from io import BytesIO
-from typing import Union
+from typing import Tuple, Union
 
 import numpy as np
 from PIL import Image
@@ -19,7 +19,11 @@ class ImageStringEncoder:
     """Class for encoding and decoding images as data URIS."""
 
     @staticmethod
-    def encode_image(image: Union[np.ndarray, str], cap_size: bool = False) -> str:
+    def encode_image(
+        image: Union[np.ndarray, str],
+        cap_size: bool = False,
+        resize_shape: Tuple[int, int] = (224, 224),
+    ) -> str:
         """Encode numpy image array to base64 string for transmission.
 
         Converts numpy arrays to PNG format and encodes as base64. For remote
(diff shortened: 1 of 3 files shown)


neuracore/ml/utils/policy_inference.py
```diff
@@ -217,9 +217,9 @@ def _process_image_data(
         Returns:
             MaskableData containing batched image tensors and attention masks.
         """
-        channels = 1 if is_depth else 3
-        values = np.zeros((1, max_len, channels, 224, 224))
-        mask = np.zeros((1, max_len))
+        # Use lists to collect images - no need to know size upfront
+        image_list = []
+        mask_list = []
```
stepjam, line 222: Why remove pre-construction? 
More efficient to pre-constrict the arrays than constantly append to lists

neuracore/core/utils/image_string_encoder.py
```diff
@@ -36,10 +40,9 @@ def encode_image(image: Union[np.ndarray, str], cap_size: bool = False) -> str:
             return image
 
         pil_image = Image.fromarray(image)
-        if cap_size and pil_image.size > (224, 224):
-            # There is a limit on the image size for non-local endpoints
-            # This is OK as almost all algorithms scale to 224x224
-            pil_image = pil_image.resize((224, 224))
+        # resize the image if the video is too large
```
stepjam, line 43: Why not:
`if cap_size and pil_image.size > resize_shape:`

stepjam (approved):
LGTM with comments

neuracore/ml/utils/policy_inference.py
kwangneuraco: Because there maybe the case that there are multiple cameras and the image size are not the same, then we cannot determine the size in advance. Just want to be more general

PR #158, opened 2025-11-26
Feat/add normalization helper

kwangneuraco:
### Features
I wrote a separate normalizer and refactor algorithms so it looks much better for each algorithms in terms of reducing replicated codes and user can choose which normalization method to use. 

- It has the parent class Normalizer and child class MinMaxNormalizer and MeanStdNormalizer, and easy to add more later
- Inside MinMax and MeanStd I clamp the min/std value so that there will be no `nan` value appear after normalization, which happens once (https://github.com/NeuracoreAI/neuracore/pull/158/files#diff-e0ab21290693ef69459c9b30a9924b891c6a6f8ddb9c5585fdc2e72cdfba2f73R122)
- For each algorithm a constant str is passed into and the correct class is initialized from the given class name, this simple approach avoids adding extra code if more normalization methods are added. Also it is easy to understand. I also checked the frontend and it looks simple to understand 
Screenshot from 2025-12-04 11-33-02

- Also added unit tests for normalizer


neuracore/ml/algorithm_utils/normalizer.py
```diff
@@ -0,0 +1,139 @@
...
+with support for PyTorch's register_buffer for proper device handling.
+"""
+
+from typing import Any, List
+
+import torch
+import torch.nn as nn
+
+
+class Normalizer(nn.Module):
```
stepjam, line 14: Please use inheritance and then you can just pass in the normaliser into the algorithm.

```
class Normalizer(nn.Module):
  pass
```


```
class MeanStdNormalizer(Normalizer):
  pass
```


```
class MinMaxNormalizer(Normalizer):
  pass
```

neuracore/ml/algorithms/act/act.py
```diff
@@ -161,85 +164,39 @@ def __init__(
...
-            ),
+            joint_states.append(self.dataset_description.joint_torques)
+
+        if (
+            DataType.JOINT_TARGET_POSITIONS
+            in self.model_init_description.output_data_types
+        ):
+            actions.append(self.dataset_description.joint_target_positions)
+
+        self.normalizer.add_statistics(
```
stepjam: I think it would look nicer to pass in the normalisation class type, and then we can instantiate it here.

```
joint_state_normalizer = normalizer_type(joint_states)
```
And later then use it as:
```
joint_state_normalizer.normalize(my_joint_state_tensor)

File: neuracore/ml/algorithm_utils/normalizer.py
Comment (stepjam) [inline neuracore/ml/algorithm_utils/normalizer.py:1]: This is much better :)

File: examples/ml/example_custom_algorithm.py [hunk 3]
```diff
@@ -219,6 +224,8 @@ def __init__(
         lr: float = 1e-4,
         lr_backbone: float = 1e-5,
         weight_decay: float = 1e-4,
+        joint_state_normalizer: str = "mean_std",
```
stepjam: You can directly pass in the type.
joint_state_normalizer: Type[Normalizer] = MeanStdNormalizer,
...

examples/ml/example_custom_algorithm.py
```diff
@@ -297,59 +308,38 @@ def __init__(
         )
 
         self.max_output_size = self.dataset_description.joint_target_positions.max_len
-
-        # Normalization statistics
-        self._setup_normalization_stats()
-
-    def _setup_normalization_stats(self) -> None:
+        # Convert string normalizer names to classes
+        normalizer_map = {
```
stepjam: This can then go once you pass in the type

examples/ml/example_custom_algorithm.py
stepjam: Same goers for all the other files

neuracore/ml/config/algorithm/act.yaml
```diff
@@ -12,3 +12,5 @@ algorithm:
   weight_decay: 1e-4
   kl_weight: 10.0
   latent_dim: 512
+  joint_state_normalizer: mean_std
```
stepjam: Don't forget to chnage these too

examples/ml/example_custom_algorithm.py
kwangneuraco: All done

PR #159, opened 2025-11-26
add fix for prefetching videos

kwangneuraco:
### Features
Fix prefetching videos when prefetching fails but empty folder is created, the old version will still see it as downloaded and skips it. The fix checks if there is really any thing downloaded


ypang-neuraco (approved):
LGTM. Please also update the unit test.

Perhaps we can also consider collecting a list of missing videos and only prefetch those but maybe that's for a future PR.

PR #476, opened 2026-03-06
fix: updated registration and trace count

StevenJacobs61:
### Features

 - Added registration manager
 - created new flow for registering traces

### Bugfixes

 - Traces were failing to register with flaky expected trace count under high throughput

cc @stepjam


neuracore/data_daemon/upload_management/upload_manager.py
```diff
@@ -94,6 +115,21 @@ async def _on_ready_for_upload(
             data_type_name: Data type name
             bytes_uploaded: Starting offset for resume
         """
+        existing = self._active_uploads.get(trace_id)
```
muneebneura: Suggestion: rename this to active_uploads or exisiting_uploads

neuracore/data_daemon/upload_management/upload_manager.py
```diff
@@ -106,8 +142,14 @@ async def _on_ready_for_upload(
             )
         )
 
-        self._active_uploads.add(task)
-        task.add_done_callback(self._active_uploads.discard)
+        self._active_uploads[trace_id] = task
+
+        def _on_done(done_task: asyncio.Task, *, tid: str = trace_id) -> None:
+            tracked = self._active_uploads.get(tid)
```
muneebneura, line 150: again can be renamed better

muneebneura (approved):
LGTM w/c

Suggestion: Don't use one word variables that are very vague like claimed, etc

PR #489, opened 2026-03-11
chore: cleanup datasets and robots in integration tests

stepjam:
Note: To be merged once https://github.com/NeuracoreAI/neuracore/actions/workflows/integration-platform.yaml goes green.

Not removing the datasets will be handy for debugging

### DIFF
--- a/tests/integration/platform/data_daemon/behavioural_correctness/test_signal_cleanup.py
+++ b/tests/integration/platform/data_daemon/behavioural_correctness/test_signal_cleanup.py
@@ -48,6 +48,10 @@ from tests.integration.platform.data_daemon.shared.test_case.build_test_case_con
 from tests.integration.platform.data_daemon.shared.test_case.constants import (
     STOP_METHOD_SIGKILL,
 )
+from tests.integration.platform.data_daemon.shared.test_infrastructure import (
+    cloud_resource_deleter,
+    cloud_resource_names,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


aditya-neuraco:
cc @stepjam  is this still needed?

stepjam:
> cc @stepjam is this still needed?

Yes. The service account has  6,970 robots(!). So the integration tests need to do cleanup. In fact, all of the integration test need to cleanup their datasets and robots.

neuracore/core/robot.py
```diff
@@ -76,6 +76,8 @@ class Robot:
     URDF and MJCF model formats and handles automatic conversion when needed.
     """
 
+    deleted = False
```
CougarTasker: Issue: Dead code, unused outside of tests.
Issue: Unexpected behavour: Encorages users to check if the robot is dead using this flag, however doing so would raise an error if this robot was actually deleted.

tests/integration/platform/data_daemon/behavioural_correctness/test_signal_cleanup.py
```diff
@@ -351,31 +354,35 @@ def test_sigkill_after_recording_allows_clean_restart(case: DataDaemonTestCase)
...
-        if results:
-            wait_for_all_traces_written(results=results)
-        logger.info("SIGKILL daemon pid=%d after completed recording", pid_first)
-        stop_daemon(method=STOP_METHOD_SIGKILL)
+    specs = build_context_specs(case=case)
 
-        assert not pid_is_running(
-            pid_first
-        ), f"pid_is_running still True for killed daemon pid={pid_first}"
+    try:
```
CougarTasker: Suggestion: for consistency use a `@contextmanager`  instead of wrapping with a try...finally. , or  add this into the `scoped_online_mode` context manager's try...finally so you do not need to repeate this

CougarTasker (approved):
LGTM \w comments

neuracore/core/robot.py
aditya-neuraco: I've changed it similarly for Datasets. https://github.com/NeuracoreAI/neuracore/pull/774

Don't want an user to operate on a deleted dataset or a deleted robot.

PR #522, opened 2026-03-31
feat: add groot

kwangneuraco:
Add groot 1.6, the newest VLA model from Nvidia https://research.nvidia.com/labs/gear/gr00t-n1_6/. 
It has the so-called system1 - system2 architecture, uses a VLM model called Cosmos-2B and a big DiT module as the action expert.

**Main difference from Pi series**:

- VLM and action expert different
- Use the mixture of real and simulation data, while Pi uses only real robot data
- Has several Nvidia-specific features to speed up the training and inference (expected), such as flash-attn, mixed bfloat16 model loading, nvidia-speed up for adam optimizer etc
- Action is mainly trained with delta action, NOT absolute.
- Beside nvidia change, also add detect of flash attn if the package installed, if not fall back to pytorch spda 
- Added fallback solution if there is not rgb image input at function `_prepare_vlm_inputs` in `groot.py` @aditya-neuraco 

**The code structure**

- eagle_config folder contains all the files related to the VLM
- modules.py has all the building blocks of action expert and adaptor, as well as the helper functions
- groot.py has the main policy architecture
- image_processor.py process the image so that it can take any size of image and resize to [448, 448]. @mark-neuracore it can be part of the pre-processing 

**Review Suggestion**

- eagle_config is mostly a copy of original files, i added the fallback of using pytorch spad if flash attention is not available
- Main part is the groot.py, especially it is mainly trained on delta actions, so i add the flag to accept either relative action output or absolute output (also fine). And the normalization is adapted to both as well
- You can have a quick look at module.py and image_processor.py

**Test**
Unit test passed. Training and inference runs no errors and loss goes down. Performance-wise the success rate is low for transfer cube task but shows promise results, i think i need to train longer. If any improvement on the performance will be another PR, this PR is focused on the model architecture.

Hi @sdas-neuraco , i am happy to let your PR on cross-embodiment in and add change to this PR, just let me know.


neuracore/ml/algorithms/groot/requirements.txt
```diff
@@ -0,0 +1,10 @@
+diffusers>=0.35.1
+transformers==4.53.2
+huggingface-hub==0.36.0
+safetensors==0.7.0
+einops
```
aditya-neuraco: Can you pin einops?

neuracore/ml/algorithms/groot/requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+peft>=0.13.0,<1.0.0
+Pillow>=10.0.0,<13.0.0
+
+# flash-attn is optional but preferred; if installed, flash_attention_2 is used automatically.
+# Otherwise falls back to SDPA (PyTorch native efficient kernels).
```
aditya-neuraco, line 9: Pytorch has flash attention 2 by default for scaled dot product attention. So we can avoid installing flash attention. It's enabled by default. 
https://docs.pytorch.org/docs/2.2/generated/torch.nn.functional.scaled_dot_product_attention.html

neuracore/ml/algorithms/groot/eagle_config/configuration_eagle3_vl.py
```diff
@@ -0,0 +1,136 @@
+"""Configuration for the Eagle3-VL model family."""
+
+# ruff: noqa: E501
+# noqa: D
+# cspell:ignore autoset
```
aditya-neuraco: Can you add this to `neuracore-dictionary.txt` instead?

neuracore/ml/algorithms/groot/eagle_config/configuration_eagle3_vl.py
```diff
@@ -0,0 +1,136 @@
...
+
+from transformers.configuration_utils import PretrainedConfig
+from transformers.models.llama.configuration_llama import LlamaConfig
+from transformers.models.phi3.configuration_phi3 import Phi3Config
+from transformers.models.qwen2.configuration_qwen2 import Qwen2Config
+from transformers.models.qwen3.configuration_qwen3 import Qwen3Config
+from transformers.models.siglip.configuration_siglip import SiglipVisionConfig
+from transformers.utils import logging
+
+from .modeling_siglip2 import Siglip2VisionConfig
```
aditya-neuraco, line 15: We can use a version of transformers hat has the `Siglip2VisionConfig`, https://github.com/huggingface/transformers/blob/main/src/transformers/models/siglip2/configuration_siglip2.py

neuracore/ml/algorithms/groot/eagle_config/config.json
aditya-neuraco, line 1: This should be a dataclass https://docs.python.org/3/library/dataclasses.html

neuracore/ml/algorithms/groot/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""GR00T N1.6 Vision-Language-Action model for Neuracore.
+
+Implements NVIDIA's GR00T N1.6 architecture: a Cosmos-Reason-2B VLM backbone
+coupled with a 32-layer Diffusion Transformer (DiT) action head. Uses
+rectified flow matching for action generation.
+
+Reference: NVIDIA GR00T N1.6 technical report.
+"""
```
aditya-neuraco: Ideally, this should go in `groot.py` just like other algorithms, because this description is what get's parsed by algorithm upload service and pushed to firestore.

We already have the description for GROOT in `groot.py`

cspell.json
```diff
@@ -0,0 +1,23 @@
...
+          "path": "./neuracore-dictionary.txt",
+          "addWords": true
+      }
+  ],
+  "language": "en,en-GB",
+  "dictionaries": [
+      "neuracore-dictionary"
+  ],
+  "words": []
+}
```
aditya-neuraco: Please remove this file. Use the one in neuracore root.

tests/unit/ml/algorithms/test_groot.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+    subprocess.check_call([
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "-r",
+        str(requirements_file),
+    ])
```
aditya-neuraco: Can you add instructions for uv as well? I would strongly suggest you import algorithm loader and use that instead.

neuracore/ml/utils/algorithm_loader.py
```diff
@@ -144,18 +149,43 @@ def install_requirements(self) -> bool:
...
-            List of Path objects representing all Python files found.
+            List of Path objects representing all algorithm files found.
         """
         files = []
         for root, _, filenames in os.walk(self.algorithm_dir):
             for filename in filenames:
-                if filename.endswith(".py"):
+                if filename.endswith((".py", ".json")):
                     files.append(Path(root) / filename)
         return files
```
aditya-neuraco, line 235: Please revert this to `*.py` files and use `dataclasses` instead of the `json` config.

neuracore/ml/utils/algorithm_loader.py
```diff
@@ -144,18 +149,43 @@ def install_requirements(self) -> bool:
...
+                ],
+                capture_output=True,
+                text=True,
+            )
+            # If dry-run shows "Would install" then something is missing
+            if "Would install" in result.stdout:
+                return False
+            return result.returncode == 0
+        except Exception:
+            return False
```
aditya-neuraco: Please add a branch to use uv as well.

neuracore/ml/config/config.yaml
```diff
@@ -10,7 +10,7 @@ epochs: 100
 output_prediction_horizon: 100
 validation_split: 0.2
 logging_frequency: 50
-keep_last_n_checkpoints: 5 
+keep_last_n_checkpoints: 1
```
aditya-neuraco: Please revert it to 5.

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1163 @@
...
+
+    @staticmethod
+    def get_supported_output_data_types() -> set[DataType]:
+        """Get the output data types supported by GR00T N1.6."""
+        return {
+            DataType.JOINT_POSITIONS,
+            DataType.JOINT_TARGET_POSITIONS,
+            DataType.PARALLEL_GRIPPER_OPEN_AMOUNTS,
+            DataType.PARALLEL_GRIPPER_TARGET_OPEN_AMOUNTS,
+        }
```
aditya-neuraco, line 1166: Have you done a training run with all these as input and output?

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1163 @@
...
+            if current_step < warmup_steps:
+                return current_step / max(1, warmup_steps)
+            progress = (current_step - warmup_steps) / max(
+                1, num_training_steps - warmup_steps
+            )
+            return max(0.0, 0.5 * (1.0 + math.cos(math.pi * progress)))
+
+        return [LambdaLR(optimizer, lr_lambda) for optimizer in optimizers]
+
+    def load_state_dict(
```
aditya-neuraco, line 1102: Isn't this a builtin pytorch function? I feel we don't need to define our own.

neuracore/core/config/config_manager.py
```diff
@@ -49,8 +49,15 @@ def config(self) -> Config:
...
-                self._config = Config.model_validate_json(f.read())
+                raw = f.read()
+
+            # Treat empty/whitespace-only files as "no config yet".
+            if not raw.strip():
+                self._config = Config()
                 return self._config
+
+            self._config = Config.model_validate_json(raw)
+            return self._config
```
aditya-neuraco: Why is this needed?

neuracore/ml/algorithms/groot/eagle_config/modeling_siglip2.py
```diff
@@ -0,0 +1,1653 @@
...
+deterministic_g = os.environ.get("FLASH_ATTENTION_DETERMINISTIC", "0") == "1"
+
+
+if is_flash_attn_2_available():
+    from flash_attn import flash_attn_func, flash_attn_varlen_func
+    from flash_attn.bert_padding import index_first_axis, pad_input, unpad_input  # noqa
+
+    _flash_supports_window_size = "window_size" in list(
+        inspect.signature(flash_attn_func).parameters
+    )
```
aditya-neuraco, line 69: Out PyTorch version (2.7.1) ships with flash attention 2 and is enabled by default.

aditya-neuraco: Flash attention speedup is also hardware specific.

neuracore/ml/algorithms/groot/requirements.txt
kwangneuraco: The sdpa will automatically use the best kernel available with built-in flash attn. The difference of flash-attn2 from huggingface is that it is optimized for padding and a bit better than the built-in.

neuracore/ml/algorithms/groot/eagle_config/configuration_eagle3_vl.py
kwangneuraco: This is similar to the pi series, they have forked the original siglip2 model from transformer and made significant changes, and using the original transformer is not compatible

neuracore/ml/config/config.yaml
kwangneuraco: Solve this in another PR

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: The rgb is required and other combination is tested, as we discussed this is another PR

neuracore/ml/algorithms/groot/eagle_config/modeling_siglip2.py
kwangneuraco: Explained in another comment

neuracore/core/config/config_manager.py
kwangneuraco: This is a fallback solution if the config is empty, and I had this issue before

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: No we need to

aditya-neuraco (approved):
LGTM w/ comments

neuracore/core/config/config_manager.py
mark-neuracore: I think this should be a separate PR and further inspected. it is a separate issue

neuracore/ml/utils/algorithm_loader.py
kwangneuraco: Yes did it

neuracore/ml/utils/algorithm_loader.py
```diff
@@ -144,13 +162,70 @@ def install_requirements(self) -> bool:
...
+
+            # uv writes to stderr; pip writes to stdout. Check both.
+            combined_output = (result.stdout or "") + (result.stderr or "")
+            # If dry-run shows "Would install" then something is missing
+            if "Would install" in combined_output:
+                return False
+            return result.returncode == 0
+
+        return False
+
```
mark-neuracore, line 220: Do we have tests for this new method?

neuracore/ml/algorithms/groot/modules.py
```diff
@@ -0,0 +1,1284 @@
...
+        raise ValueError(
+            f"Action horizon {action_horizon} exceeds "
+            f"max_action_pos_embeddings={max_action_pos_embeddings}"
+        )
+
+    position_ids = torch.arange(action_horizon, device=action_tokens.device)
+    position_embs = action_position_embedding(position_ids).unsqueeze(0)
+    return action_tokens + position_embs.to(dtype=action_tokens.dtype)
+
+
```
mark-neuracore, line 338: I think for readability, these functions above here could move into a utils file. modules is already a big file with all the module classes in it.

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+    VLMBackbone,
+    _add_action_step_positional_encoding,
+    _load_pretrained_state_dict,
+)
+
+logger = logging.getLogger(__name__)
+
+# Normalizer types for proprioception and actions
+PROPRIO_NORMALIZER = MeanStdNormalizer
+ACTION_NORMALIZER = MeanStdNormalizer
```
mark-neuracore, line 75: Confusing styling. Capital snake-case is for constants. Doesn't read well when we see ACTION_NORMALIZER(...)

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+                UserWarning,
+                stacklevel=2,
+            )
+
+        # Build input proprioceptive data
+        self.proprio_dims: dict[DataType, tuple[int, int]] = {}
+        proprio_stats = []
+        current_dim = 0
+
+        for data_type in [
```
mark-neuracore, line 276: we need to discuss a better way of handling this.  If we introduce a new proprio type we will then have to modify this here to include it (and probably other algorithms too)

tests/unit/ml/algorithms/test_groot.py
kwangneuraco: Yep done

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+            DataType.JOINT_POSITIONS,
+            DataType.JOINT_VELOCITIES,
+            DataType.JOINT_TORQUES,
+            DataType.PARALLEL_GRIPPER_OPEN_AMOUNTS,
+        ]:
+            if data_type not in self.data_types:
+                continue
+            if data_type == DataType.PARALLEL_GRIPPER_OPEN_AMOUNTS:
```
mark-neuracore, line 284: we also might want to consider that. it is a minor thing, but because we have jointdatastat and parallel gripper stats we have to have a condition here to consider them differently. what if we started adding another type of grippers for example?

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+            self.add_action_step_positional_encoding
+            and self.output_prediction_horizon > self.max_action_pos_embeddings
+        ):
+            raise ValueError(
+                "output_prediction_horizon="
+                f"{self.output_prediction_horizon} exceeds "
+                f"max_action_pos_embeddings={self.max_action_pos_embeddings}"
+            )
+
+        # State encoder: (B, max_state_dim) -> (B, input_embedding_dim)
```
mark-neuracore, line 452: question: shouldn't these be in modules.py? I know they are just small mlp heads but im asking more for structure consistency

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+    def _combine_proprio(self, batch: BatchedInferenceInputs) -> torch.Tensor:
+        """Combine, normalize, and pad proprioceptive state.
+
+        Args:
+            batch: Input batch containing proprioceptive observations.
+
+        Returns:
+            Normalized and padded state tensor (B, max_state_dim).
+        """
+        proprio_list = []
```
mark-neuracore, line 604: same comment about proprio types as before

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+        """Prepare pixel_values, input_ids, and attention_mask for the VLM.
+
+        Args:
+            batch: Input batch containing RGB_IMAGES and optionally LANGUAGE.
+
+        Returns:
+            Tuple of (pixel_values, input_ids, attention_mask).
+        """
+        if DataType.RGB_IMAGES in batch.inputs:
+            rgb_data = cast(list[BatchedRGBData], batch.inputs[DataType.RGB_IMAGES])
```
mark-neuracore, line 641: can depth images be considered?

neuracore/ml/algorithms/groot/groot.py
```diff
@@ -0,0 +1,1166 @@
...
+        value (e.g. JOINT_TARGET_POSITIONS uses JOINT_POSITIONS as base).
+
+        Args:
+            batch: Input batch.
+
+        Returns:
+            Current output state tensor (B, actual_action_dim).
+        """
+        # Map each output type to its corresponding input type
+        output_to_input: dict[DataType, DataType] = {
```
mark-neuracore, line 688: Same as before for data types

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: OK this is valid

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: I agree! A possible way is having constants in the neuracore types that contains all proprioceptive datatypes and check with allowed input datatypes with each algorithm. Let's discuss it!

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: Same as the previous comment

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: depth image is not trained in groot unfortunately

neuracore/ml/algorithms/groot/groot.py
kwangneuraco: Yes it can be moved into modules, but because it is pretty small and the code is simple, we can just put it here

PR #527, opened 2026-04-07
feat: robust auto batch size

gtiboni-neuraco:
### Auto tuning of batch size improved with features:
- Uses real training set instead of SingleSampleDataset mockup (to mimic real RAM usage)
- Uses both train and validation loaders (to mimic real RAM usage)
- Uses the same num_workers for dataloader as in  training (to mimic real RAM usage)
- optimized gpu usage at training time with `.zero_grad()` and `del batch, loss`
- `granularity` parameter: avoids a full binary search up to the unit of batch size; instead, stop when the search range is narrow enough
- tests updated accordingly: `test_train.py` and `test_batch_autotuner.py`
- it takes around 30-200 seconds, depending on GPU and dataset size.
- tested on diffusion_policy and cnnmlp algorithms, different datasets, and GPUs T4 and V100.

#### Next in the future:
- Make it faster via power of 2s only
- Make it faster via interpolation and prediction


sdas-neuraco:
LGTM

sdas-neuraco (approved):
LGTM

PR #528, opened 2026-04-07
feat: add preprocessing configs

mark-neuracore:
This is a first version of having preprocessing pipelines. The structure of the pydantic models used is defined in neuracore_types with its own pr.

For this stage as you see, we are:
- not exposing any pipeline setup to the user, but rather having our fixed pipeline. (mainly to make sure all plumping is working as intended). Note that defining a preprocessing pipeline (in python and with yaml) is doable with this version.
- only introducing one built-in preprocessing function for rgb and depth images called resize_pad.
-  removed hard-coded reshaping inside each algorithm so they fully depend on the preprocessing.
- introduced unit tests for added functionality

Next stages should:
- add user documentation
- remove fixed preprocessing config
- add algorithm-specific restrictions (restriction on the allowed resizing size, normalisation methods etc)
- add more methods (gaussian noise, blur, jittering etc for images, normalisation, language embedding etc for other modalities)
- add automatic generation of output inference postprocessing generation. Note that this will need to be added in the same pr when normalisation gets added.
- add gui and cli interfaces to create and inspect preprocessing configs.

### DIFF
--- a/neuracore/core/endpoint.py
+++ b/neuracore/core/endpoint.py
@@ -34,6 +34,10 @@ from neuracore.core.get_latest_sync_point import get_latest_sync_point
 from neuracore.core.utils.download import download_with_progress
 from neuracore.ml.logging.endpoint_log_streamer import EndpointLogStreamer
 from neuracore.ml.utils.endpoint_storage_handler import EndpointStorageHandler
+from neuracore.ml.utils.preprocessing_utils import (
+    PreprocessingConfiguration,
+    serialize_preprocessing_config,
+)
 
 from .auth import get_auth
 from .const import API_URL, PING_ENDPOINT, PREDICT_ENDPOINT, SET_CHECKPOINT_ENDPOINT
@@ -147,6 +151,7 @@ class DirectPolicy(Policy):
         org_id: str,
         input_embodiment_description: EmbodimentDescription | None = None,
         output_embodiment_description: EmbodimentDescription | None = None,
+        input_preprocessing_config: PreprocessingConfiguration | None = None,
         job_id: str | None = None,
         device: str | None = None,
         robot_id: str | None = None,
@@ -159,6 +164,7 @@ class DirectPolicy(Policy):
         self._policy = PolicyInference(
             input_embodiment_description=input_embodiment_description,
             output_embodiment_description=output_embodiment_description,
+            input_preprocessing_config=input_preprocessing_config,
             org_id=org_id,
             job_id=job_id,
             model_file=model_path,
@@ -363,6 +369,7 @@ class LocalServerPolicy(ServerPolicy):
         model_path: Path,
         input_embodiment_description: EmbodimentDescription | None = None,
         output_embodiment_description: EmbodimentDescription | None = None,
+        input_preprocessing_config: PreprocessingConfiguration | None = None,
         device: str | None = None,
         job_id: str | None = None,
         port: int = 8080,
@@ -386,6 +393,8 @@ class LocalServerPolicy(ServerPolicy):
             endpoint_id: Optional deployed endpoint ID used for cloud log uploads.
             robot_id: Optional robot ID used to resolve embodiments from model
                 metadata when explicit embodiments are omitted.
+            input_preprocessing_config: Preprocessing configuration for the input
+                data.
         """
         super().__init__(
             f"http://{host}:{port}",
@@ -393,6 +402,7 @@ class LocalServerPolicy(ServerPolicy):
         )
         self.input_embodiment_description = input_embodiment_description
         self.output_embodiment_description = output_embodiment_description
+        self.input_preprocessing_config = input_preprocessing_config
         self.robot_id = robot_id
         self.org_id = org_id
         self.job_id = job_id
@@ -460,6 +470,14 @@ class LocalServerPolicy(ServerPolicy):
             cmd.extend([
                 "--output-embodiment-description",
                 f"{output_embodiment_description_str}",
+            ])
+        if self.input_preprocessing_config is not None:
+            input_preprocessing_config_serialized = serialize_preprocessing_config(
+                self.input_preprocessing_config
+            )
+            cmd.extend([
+                "--input-preprocessing-config",
+                json.dumps(input_preprocessing_config_serialized),
             ])
         if self.robot_id is not None:
             cmd.extend(["--robot-id", self.robot_id])
@@ -613,6 +631,7 @@ class RemoteServerPolicy(ServerPolicy):
 def policy(
     input_embodiment_description: EmbodimentDescription | None = None,
     output_embodiment_description: EmbodimentDescription | None = None,
+    input_preprocessing_config: PreprocessingConfiguration | None = None,
     train_run_name: str | None = None,
     model_file: str | None = None,
     device: str | None = None,
@@ -625,6 +644,7 @@ def policy(
             be fed into the model
         output_embodiment_description: Specification of the order that will
             be output from the model
+        input_preprocessing_config: Preprocessing configuration for the input data.
         train_run_name: Name of the training run to load the model from.
         model_file: Path to the model file to load.
         device: Torch device to run the model on (CPU or GPU, or MPS).
@@ -647,6 +667,7 @@ def policy(
     return DirectPolicy(
         input_embodiment_description=input_embodiment_description,
         output_embodiment_description=output_embodiment_description,
+        input_preprocessing_config=input_preprocessing_config,
         org_id=org_id,
         job_id=job_id,
         model_path=model_path,
@@ -658,6 +679,7 @@ def policy(
 def policy_local_server(
     input_embodiment_description: EmbodimentDescription | None = None,
     output_embodiment_description: EmbodimentDescription | None = None,
+    input_preprocessing_config: PreprocessingConfiguration | None = None,
     train_run_name: str | None = None,
     model_file: str | None = None,
     device: str | None = None,
@@ -674,6 +696,7 @@ def policy_local_server(
             will be fed into the model
         output_embodiment_description: Specification of the order that
             will be output from the model
+        input_preprocessing_config: Preprocessing configuration for the input data.
         train_run_name: Name of the training run to load the model from.
         model_file: Path to the model file to load.
         device: Device model to be loaded on.
@@ -709,6 +732,7 @@ def policy_local_server(
         model_path=model_path,
         input_embodiment_description=input_embodiment_description,
         output_embodiment_description=output_embodiment_description,
+        input_preprocessing_config=input_preprocessing_config,
         device=device,
         job_id=job_id,
         port=port,
--- a/neuracore/core/endpoint.py
+++ b/neuracore/core/endpoint.py
@@ -34,10 +34,7 @@ from neuracore.core.get_latest_sync_point import get_latest_sync_point
 from neuracore.core.utils.download import download_with_progress
 from neuracore.ml.logging.endpoint_log_streamer import EndpointLogStreamer
 from neuracore.ml.utils.endpoint_storage_handler import EndpointStorageHandler
-from neuracore.ml.utils.preprocessing_utils import (
-    PreprocessingConfiguration,
-    serialize_preprocessing_config,
-)
+from neuracore.ml.utils.preprocessing_utils import PreprocessingConfiguration
 
 from .auth import get_auth
 from .const import API_URL, PING_ENDPOINT, PREDICT_ENDPOINT, SET_CHECKPOINT_ENDPOINT
@@ -472,9 +469,10 @@ class LocalServerPolicy(ServerPolicy):
                 f"{output_embodiment_description_str}",
             ])
         if self.input_preprocessing_config is not None:
-            input_preprocessing_config_serialized = serialize_preprocessing_config(
-                self.input_preprocessing_config
-            )
+            input_preprocessing_config_serialized = {
+                data_type.value: [m.to_dict() for m in methods]
+                for data_type, methods in self.input_preprocessing_config.items()
+            }
             cmd.extend([
                 "--input-preprocessing-config",
                 json.dumps(input_preprocessing_config_serialized),
--- a/neuracore/ml/algorithms/act/act.py
+++ b/neuracore/ml/algorithms/act/act.py
@@ -212,6 +212,7 @@ class ACT(NeuracoreModel):
                 self.input_dataset_statistics[DataType.RGB_IMAGES],
             )
             max_cameras = len(stats)
+            self.image_normalizers = nn.ModuleList()
             self.image_encoders = nn.ModuleList()
             for i in range(max_cameras):
                 if use_resnet_stats:
@@ -220,15 +221,8 @@ class ACT(NeuracoreModel):
                     mean_c_h_w, std_c_h_w = stats[i].frame.mean, stats[i].frame.std
                     mean = mean_c_h_w.mean(axis=(1, 2)).tolist()
                     std = std_c_h_w.mean(axis=(1, 2)).tolist()
-
-                encoder = nn.ModuleDict({
-                    "transform": torch.nn.Sequential(
-                        T.Resize((224, 224)),
-                        T.Normalize(mean=mean, std=std),
-                    ),
-                    "encoder": ACTImageEncoder(output_dim=hidden_dim),
-                })
-                self.image_encoders.append(encoder)
+                self.image_normalizers.append(T.Normalize(mean=mean, std=std))
+                self.image_encoders.append(ACTImageEncoder(output_dim=hidden_dim))
 
         # CLS token embedding for latent encoder
         self.cls_embed = nn.Parameter(torch.randn(1, 1, hidden_dim))
@@ -473,12 +467,12 @@ class ACT(NeuracoreModel):
         # Process images
         image_features = []
         image_pos = []
-        for cam_id, (encoder_dict, input_rgb) in enumerate(
-            zip(self.image_encoders, batched_rgb_data)
+        for cam_id, (normalizer, encoder, input_rgb) in enumerate(
+            zip(self.image_normalizers, self.image_encoders, batched_rgb_data)
         ):
             last_frame = input_rgb.frame[:, -1, :, :, :]  # (B, 3, H, W)
-            transformed = encoder_dict["transform"](last_frame)
-            features, pos = encoder_dict["encoder"](
+            transformed = normalizer(last_frame)
+            features, pos = encoder(
                 transformed
             )  # Vision backbone provides features and pos
             features *= camera_images_mask[:, cam_id].view(batch_size, 1, 1, 1)
--- a/neuracore/ml/config/config.yaml
+++ b/neuracore/ml/config/config.yaml
@@ -48,6 +48,24 @@ output_cross_embodiment_description: {
 
 }
 
+# Dict[str, Dict[DataType, List[MethodConfig]]], e.g.:
+# for images, we must have a resize step to reshape all images from different sources to the same size.
+preprocessing:
+  input:
+    RGB_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+    DEPTH_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+  output:
+    RGB_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+    DEPTH_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+
 # Paths and IDs (for cloud vs local training)
 algorithm_id: null    # For cloud training
 training_id: null     # For cloud training
--- a/neuracore/ml/datasets/pytorch_synchronized_dataset.py
+++ b/neuracore/ml/datasets/pytorch_synchronized_dataset.py
@@ -13,6 +13,7 @@ from neuracore_types import (
     CrossEmbodimentDescription,
     DataType,
     EmbodimentDescription,
+    EmbodimentUnion,
     NCDataStats,
     SynchronizedDatasetStatistics,
     SynchronizedPoint,
@@ -29,6 +30,10 @@ from neuracore.core.utils.training_input_args_validation import (
 from neuracore.ml import BatchedTrainingSamples
 from neuracore.ml.datasets.pytorch_neuracore_dataset import PytorchNeuracoreDataset
 from neuracore.ml.utils.memory_monitor import MemoryMonitor
+from neuracore.ml.utils.preprocessing_utils import (
+    PreprocessingConfiguration,
+    apply_preprocessing_methods,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -57,6 +62,8 @@ class PytorchSynchronizedDataset(PytorchNeuracoreDataset):
         synchronized_dataset: SynchronizedDataset,
         input_cross_embodiment_description: CrossEmbodimentDescription,
         output_cross_embodiment_description: CrossEmbodimentDescription,
+        input_preprocessing_config: PreprocessingConfiguration,
+        output_preprocessing_config: PreprocessingConfiguration,
         output_prediction_horizon: int,
     ):
         """Initialize the dataset.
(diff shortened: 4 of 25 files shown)


neuracore/ml/preprocessing/methods.py
```diff
@@ -0,0 +1,61 @@
+"""Preprocessing method implementations."""
+
+from __future__ import annotations
+
+import torch
+from neuracore_types import BatchedDepthData, BatchedNCData, BatchedRGBData
+
+
+def resize_pad(
```
ypang-neuraco: Do you plan to reimplement each transform? Why not reuse the transforms in torchvision?

neuracore/ml/preprocessing/methods.py
ypang-neuraco, line 1: Maybe methods should be in neuracore_types

neuracore/ml/preprocessing/runtime.py
```diff
@@ -0,0 +1,76 @@
...
+
+from .methods import resize_pad
+
+# Single source of truth for available runtime methods and compatibility.
+_METHOD_REGISTRY: dict[str, MethodSpec] = {
+    "resize_pad": MethodSpec(
+        handler=resize_pad,
+        allowed_data_types={DataType.RGB_IMAGES, DataType.DEPTH_IMAGES},
+    ),
+}
```
ypang-neuraco: Maybe this should be in neuracore_types

neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -141,6 +148,26 @@ def __init__(
...
+        if preprocessing_config is None:
+            return batched_nc_data
+        methods = preprocessing_config.steps.get(data_type, {}).get(slot_idx, [])
+        if not methods:
+            return batched_nc_data
+        return apply_methods_for_slot(
+            data_type=data_type,
+            batched_data=batched_nc_data,
+            methods=methods,
+        )
```
ypang-neuraco: This can be in the runtime

neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -141,6 +148,26 @@ def __init__(
...
+    @staticmethod
+    def _apply_slot_preprocessing(
+        data_type: DataType,
+        slot_idx: int,
+        batched_nc_data: BatchedNCData,
+        preprocessing_config: PreProcessingConfiguration | None,
+    ) -> BatchedNCData:
+        if preprocessing_config is None:
+            return batched_nc_data
+        methods = preprocessing_config.steps.get(data_type, {}).get(slot_idx, [])
```
ypang-neuraco: This is not very readable, this should be a function in PreProcessingConfiguration (e.g. `get_step`)

neuracore/ml/utils/policy_inference.py
```diff
@@ -97,10 +102,22 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
+                if input_preprocessing_config is not None:
+                    methods = input_preprocessing_config.steps.get(data_type, {}).get(
+                        slot_idx, []
+                    )
+                    if methods:
+                        tensor = apply_methods_for_slot(
+                            data_type=data_type,
+                            batched_data=tensor,
+                            methods=methods,
+                        )
```
ypang-neuraco: Reuse `_apply_slot_preprocessing` after moving to `runtime.py`

neuracore/ml/preprocessing/methods.py
mark-neuracore: neuracore_types, to me seems like a place to define pydantic models and structures not implementation details. that's why i have the methods here.

neuracore/ml/algorithms/act/act.py
```diff
@@ -212,15 +213,8 @@ def __init__(
...
-
-                encoder = nn.ModuleDict({
-                    "transform": torch.nn.Sequential(
-                        T.Resize((224, 224)),
-                        T.Normalize(mean=mean, std=std),
-                    ),
-                    "encoder": ACTImageEncoder(output_dim=hidden_dim),
-                })
-                self.image_encoders.append(encoder)
+                self.image_normalizers.append(T.Normalize(mean=mean, std=std))
```
kwangneuraco, line 224: Do the mean and std also the parameters considered during the pre-processing? Or it is better sitted inside the algorithm? My opinion as it is more closely related to visual encoder you use, so sitted in algorithm maybe better

neuracore/ml/preprocessing/methods.py
kwangneuraco: Also the filename `methods.py` is too general for me. Given we will have other functionalities like augmentation, normalization etc

kwangneuraco:
LGTM/w comments. Also in the future we need to move the state/action normalization here as well

neuracore/ml/algorithms/act/act.py
mark-neuracore: normalisers arent going to move to the preprocessing pipelines for now, as explained in the plan above. so they are staying here ( in the exact same format as they were already).

mark-neuracore: Also considering the comment on the whole PR, as mentioned in the plan, normalisers will come on a later stage.

neuracore/ml/preprocessing/methods.py
mark-neuracore: can be split into different files later down the line for sure

neuracore/ml/preprocessing/runtime.py
mark-neuracore: i think the registry should live in neuracore to have the handles from the methods.py where it gets defined (inside neuracore).

i was thinking about it if you have added a new method to our built-in methods, we would:
1- define it in methods.py
2- go to the registry and add it there as another key

and that means we have modified one package to add the new method.   also makes more sense to me that  to have the manifests and dataclasses and pydantic models etc defined in neuracore_types, and any actual implementation or functionality to be in neuracore itself
  
 Happy to discuss another alternative.

neuracore/ml/preprocessing/methods.py
```diff
@@ -0,0 +1,61 @@
...
+    batched_data: BatchedRGBData | BatchedDepthData,
+    size: list[int] | tuple[int, int] = (224, 224),
+) -> BatchedNCData:
+    """Resize and pad a batched RGB/depth item to a fixed spatial size."""
```
stepjam: Missing param docstring; and in a few other places

stepjam (changes requested):
I've kept my comments in one place rather than spreading across neuracore_types and here. 

The string-keyed `_METHOD_REGISTRY` and `_build_fixed_preprocessing_config` in `train.py` are the main things I'd like to see rethought as this stage before I go through for another pass. 

- Registering methods by string name is fragile; typos fail silently until runtime, there's no IDE support, and the registry becomes a maintenance burden as methods grow.

- Rather than encoding method metadata as a Pydantic model (PreProcessingMethod), we should use a **proper base class for preprocessing**. Something like:

  ```python
  # preprocessing/base.py
  class PreprocessingMethod(ABC):
      @staticmethod
      @abstractmethod
      def allowed_data_types() -> frozenset[DataType]: ...

      @abstractmethod
      def __call__(self, data: BatchedNCData,) -> BatchedNCData: ...
  ```
  
    ```python
  # preprocessing/resize_pad.py
  class ResizePad(PreprocessingMethod):
  
      def __init__(resize_shape: tuple[int, int], ...): ...
      
      @staticmethod
      def allowed_data_types() -> frozenset[DataType]:
        return {DataType.RGB, ...}
      
      def __call__(self, data: BatchedNCData,) -> BatchedNCData: ...
  ```

Each concrete implementation (e.g. `ResizePad`) declares its own `allowed_data_types()` statically. No string keys and no external registry needed. As a bonus, because these are now proper classes, Hydra can construct them directly via config:

```yaml
preprocessing:
  - _target_: preprocessing.resize_pad.ResizePad
    resize_shape: [224, 224]
  - _target_: preprocessing.normalize.NormalizeResnet
```

And then on the Python side:

```python
steps = hydra.utils.instantiate(cfg.preprocessing)  # returns list[PreprocessingMethod]
```

- Please drop PreprocessingPhase. It seems premature at this phase, and can be added in much later if needed. 

- I would challenge you to evaluate whether PreProcessingConfiguration can come in a followup PR. Perhaps this PR can be about parity to what we already have. And then a follow up can go beyond parity. (e.g. preprocessing per slot).

After the above has been address, this should simplify and reduce some of the code. After that ping me again and I'll take another look.

neuracore/ml/preprocessing/methods.py
stepjam, line 1: In my overview comment this file becomes unnecessary, but for future ref, methods.py is very vague and ideally would have been more descriptive. Same goes for the runtime.py file.

neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -58,6 +60,8 @@ def __init__(
         input_cross_embodiment_description: CrossEmbodimentDescription,
         output_cross_embodiment_description: CrossEmbodimentDescription,
         output_prediction_horizon: int,
+        input_preprocessing_config: dict[DataType, list[PreprocessingMethod]],
+        output_preprocessing_config: dict[DataType, list[PreprocessingMethod]],
```
ypang-neuraco: Maybe we should have a class for PreprocessingConfig? Similar to CrossEmbodimentDescription in neuracore_types

neuracore/ml/preprocessing/base.py
ypang-neuraco, line 1: Perhaps the base class should be in neuracore_types

neuracore/ml/preprocessing/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        allowed_types = method.allowed_data_types()
+        if data_type not in allowed_types:
+            allowed_list = ", ".join(sorted(dt.value for dt in allowed_types))
+            raise ValueError(
+                f"Preprocessing method '{type(method).__name__}' "
+                "is not allowed for data type "
+                f"{data_type.value}. Allowed data types: [{allowed_list}]"
+            )
+        batched_data = method(batched_data)
+    return batched_data
```
ypang-neuraco: Instead of a function it can be a class that stores the preprocessing configs and apply them based on the datatype and whether it is input/output.

neuracore/ml/preprocessing/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    methods = preprocessing_configs.get(data_type, [])
+    for method in methods:
+        allowed_types = method.allowed_data_types()
+        if data_type not in allowed_types:
+            allowed_list = ", ".join(sorted(dt.value for dt in allowed_types))
+            raise ValueError(
+                f"Preprocessing method '{type(method).__name__}' "
+                "is not allowed for data type "
+                f"{data_type.value}. Allowed data types: [{allowed_list}]"
+            )
```
ypang-neuraco: This validation should be done before the dataset is constructed. Otherwise the user has to wait for the prefetch to finish before getting this feedback.

neuracore/ml/config/config.yaml
```diff
@@ -48,6 +48,24 @@ output_cross_embodiment_description: {
 
 }
 
+# Dict[str, Dict[DataType, List[MethodConfig]]], e.g.:
+# for images, we must have a resize step to reshape all images from different sources to the same size.
+preprocessing:
+  input:
+    RGB_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
```
ypang-neuraco, line 56: I understand we are trying to move away from method registry but it seems cumbersome to have to repeat the path for every step. Maybe we can hardcode "neuracore.ml.preprocessing.methods" in train.py and keep only "method_file.method_class" in the config?

neuracore/ml/train.py
```diff
@@ -503,6 +552,8 @@ def determine_optimal_batch_size(
         input_data_types=extract_data_types(input_cross_embodiment_description),
         output_data_types=extract_data_types(output_cross_embodiment_description),
         output_prediction_horizon=cfg.output_prediction_horizon,
+        input_preprocessing_config=input_preprocessing_config,
+        output_preprocessing_config=output_preprocessing_config,
```
ypang-neuraco: Why is this passed in here? I don't see it being used inside the model

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/preprocessing/base.py
mark-neuracore: As discussed offline, it will be kept here

neuracore/ml/preprocessing/methods.py
mark-neuracore: whenever possible we will just wrap available transforms in our class.

mark-neuracore:
@stepjam @kwangneuraco @sdas-neuraco @ypang-neuraco 

Hello everyone,

Thanks for taking the time to review. This PR has gone through several changes based on all your comments. 

Here is an overview of the current state:

1. **Base Class:** Added a base class for preprocessing methods.
2. **First Method:** Implemented resizing pad as our first preprocessing method, which extends the base class in `preprocessing/methods/`.
3. **Pipeline Split:** Preprocessing is split separately for inputs and outputs.
4. **Sequential Application:** Defined a list of methods to be applied in sequence on each data stream of each data type. `Dict[DataType, list[PreprocessingMethods]]`
5. **Saving:** Preprocessing gets serialized and saved in the `nc.zip` file as a JSON, following our cross-embodiment saving method.
6. **Inference:** Similar to cross-embodiment, all policy inference classes (direct, remote, local server) can now read preprocessing configs from the saved JSON files.  
7. **Config Overrides:** Policy classes can also take preprocessing parameters to override the saved configurations. If neither is available, a `ValueError` is thrown.
8. **Hydra Integration:** As discussed with @stepjam, preprocessing defaults are client-facing in the Hydra config for training and can be modified there or through the CLI (via Hydra).
9. **Validation Updates:** `run_validation()` makes up a dummy dataset and synthesizes cross-embodiment. It now also synthesizes preprocessing configs to prevent shape mismatches (between training on dummy dataset batches and inference on NC batch data samples).
10. **Testing:** Added tests to cover the majority of the changes mentioned above.

Let me know if you have any further feedback!

neuracore/ml/utils/policy_inference.py
```diff
@@ -94,6 +107,32 @@ def __init__(
...
+                "Please provide a input preprocessing configuration."
+            )
+        if effective_output_preprocessing_config:
+            self.output_preprocessing_config = effective_output_preprocessing_config
+        else:
+            raise ValueError(
+                "Output preprocessing configuration is missing "
+                "from policy initialization and not found in the model archive! "
+                "Please provide a output preprocessing configuration."
+            )
```
ypang-neuraco, line 120: Please simplify for readability e.g.
```suggestion
        self.input_preprocessing_config = (
            input_preprocessing_config or archive_input_preprocessing_config
        )
        self.output_preprocessing_config = (
            output_preprocessing_config or archive_output_preprocessing_config
        )
        if not self.input_preprocessing_config:
            raise ValueError(
                "Input preprocessing configuration is missing "
                "from policy initialization and not found in the model archive! "
                "Please provide a input preprocessing configuration."
            )
        if not self.output_preprocessing_config:
            raise ValueError(
                "Output preprocessing configuration is missing "
                "from policy initialization and not found in the model archive! "
                "Please provide a output preprocessing configuration."
            )
```

neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -392,15 +405,25 @@ def load_sample(
...
-                    inputs[data_type].append(
-                        batched_nc_data_class.sample(batch_size=1, time_steps=1)
+                    batched_nc_data = batched_nc_data_class.sample(
+                        batch_size=1, time_steps=1
+                    )
+                    batched_nc_data = apply_preprocessing_configs(
+                        data_type=data_type,
+                        batched_data=batched_nc_data,
+                        preprocessing_configs=self._input_preprocessing_config,
                     )
```
ypang-neuraco: I'm not sure we want to do this here for all preprocessing. e.g. Should Gaussian noise be applied to missing data? Same with output

neuracore/core/endpoint.py
ypang-neuraco, line 1: We should also update the endpoint APIs but maybe it is out of scope for this PR

neuracore/ml/utils/policy_inference.py
```diff
@@ -69,13 +75,20 @@ def __init__(
...
+                When None, values are loaded from the model archive.
         """
         self.org_id = org_id
         self.job_id = job_id
         (
             self.model,
             input_cross_embodiment_description,
             output_cross_embodiment_description,
+            archive_input_preprocessing_config,
+            archive_output_preprocessing_config,
```
ypang-neuraco: I think output config is not required for inference

neuracore/ml/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    methods = preprocessing_configs.get(data_type, [])
+    for method in methods:
+        allowed_types = method.allowed_data_types()
+        if data_type not in allowed_types:
+            allowed_list = ", ".join(sorted(dt.value for dt in allowed_types))
+            raise ValueError(
+                f"Preprocessing method '{type(method).__name__}' "
+                "is not allowed for data type "
+                f"{data_type.value}. Allowed data types: [{allowed_list}]"
+            )
```
ypang-neuraco: This valiation should also be done in resolve_preprocessing_config to provide early warnings/feedback to the user

neuracore/ml/train.py
```diff
@@ -983,6 +991,27 @@ def _main(cfg: DictConfig) -> None:
         else:
             device = get_default_device()
 
+        preprocessing_dictconfig = cfg.get("preprocessing", {})
+        if not preprocessing_dictconfig:
+            raise ValueError(
+                "Preprocessing configuration is missing ! Please provide a "
+                "preprocessing configuration."
+            )
```
sdas-neuraco, line 999: why does the user have to provide a pre-processing config?

neuracore/ml/train.py
```diff
@@ -983,6 +991,27 @@ def _main(cfg: DictConfig) -> None:
         else:
             device = get_default_device()
 
+        preprocessing_dictconfig = cfg.get("preprocessing", {})
```
sdas-neuraco, line 994: This statement makes it so the var will default to {} if the pre-rpcessing key is not found, making the if statement below redundent?

neuracore/ml/utils/policy_inference.py
```diff
@@ -94,6 +107,32 @@ def __init__(
             input_cross_embodiment_description=input_cross_embodiment_description,
             output_cross_embodiment_description=output_cross_embodiment_description,
         )
```
sdas-neuraco, line 106: not too sure whats going on here? maybe some comments might help?

neuracore/ml/utils/policy_inference.py
sdas-neuraco: agree

neuracore/ml/utils/policy_inference.py
```diff
@@ -137,10 +176,17 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
+            for slot_idx, (name, nc_data) in enumerate(
+                sync_point.data[data_type].items()
+            ):
                 tensor = DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type].from_nc_data(
                     nc_data
                 )
+                tensor = apply_preprocessing_configs(
+                    data_type=data_type,
+                    batched_data=tensor,
+                    preprocessing_configs=self.input_preprocessing_config,
```
sdas-neuraco: do we need to pass in the entire pre-processing config? can we not just pass in the relevent details to make the function more isolated? rather then passing in excess data?

neuracore/ml/preprocessing/methods/resize_pad.py
```diff
@@ -0,0 +1,72 @@
...
+from neuracore_types import BatchedDepthData, BatchedNCData, BatchedRGBData, DataType
+
+from ..base import PreprocessingMethod
+
+
+class ResizePad(PreprocessingMethod):
+    """Resize and pad RGB/depth tensors to fixed spatial size."""
+
+    def __init__(self, size: list[int] | tuple[int, int] = (224, 224)) -> None:
+        """Initialize resize target as (height, width)."""
```
sdas-neuraco, line 19: doc string could help, especially when we want to expose this to the web front end and can parse it to give hit suggestions

neuracore/ml/preprocessing/methods/resize_pad.py
```diff
@@ -0,0 +1,72 @@
...
+        self.size = size
+
+    @staticmethod
+    def allowed_data_types() -> frozenset[DataType]:
```
sdas-neuraco, line 23: curious why frozenset?

mark-neuracore: fixed set of allowed types

neuracore/ml/datasets/pytorch_synchronized_dataset.py
mark-neuracore: we need to reshape sample() methods output because they have hardcoded shapes.

neuracore/ml/train.py
mark-neuracore: They come from hydra configs as defaults anyway. so if user dont provide them, AND remove them from hydra this will be triggered

neuracore/ml/train.py
mark-neuracore: Not really. Here is why


Python 3.10.19 (main, Oct 21 2025, 16:43:05) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> x= {}
>>> if not x: print("i am a poor empty dict")
... 
i am a poor empty dict
>>>

neuracore/ml/utils/policy_inference.py
mark-neuracore: I thought about this. we need to pass data_type to check if it is in allowed methods. we cannot just infer the data type from the batched_data (it is many-to-one; velocity and position and torque data types all become bached joint data). so the only saving will be to get the data types methods: list[PreprocessingMethods] and pass that. happy to do that but it sounded nicer to send the preprocessing configs.

neuracore/ml/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Any
+
+from neuracore_types import BatchedNCData, DataType
+
+from neuracore.ml.preprocessing.base import PreprocessingMethod
+
+PreprocessingConfiguration = dict[DataType, list[PreprocessingMethod]]
+
+
+def resolve_preprocessing_config(
```
stepjam, line 29: This function makes me feel a littler uneasy. 
Right now it does a lot of manual work: parsing DataType keys, checking for _target_ fields, and calling instantiate per entry. Most of that is just working around receiving a plain dict rather than a proper DictConfig.

If the archive restore goes through OmegaConf.create before hitting resolve_preprocessing_config, the function can drop most of that boilerplate and keep only the allowed_data_types validation.

```
# before: resolve_preprocessing_config receives a plain dict and does everything manually
for data_type_key, method_entries in preprocessing_config_dict.items():
    data_type = DataType(str(data_type_key).split(".")[-1].strip())
    for method_entry in method_entries:
        if "_target_" not in method_entry:
            raise ValueError(...)
        method = instantiate(method_entry, _convert_="object")
        ...

# after: receives a DictConfig
def resolve_preprocessing_config(cfg: DictConfig) -> PreprocessingConfiguration:
    resolved = {}
    for data_type_key, methods_cfg in cfg.items():
        data_type = DataType(data_type_key)
        methods = [instantiate(m) for m in methods_cfg]
        for method in methods:
            if data_type not in method.allowed_data_types():
                raise ValueError(...)
        resolved[data_type] = methods
    return resolved

# archive restore; The callers just wrap before passing in:
cfg = OmegaConf.create(json.load(f))
config = resolve_preprocessing_config(cfg)
```
This also means the manual OmegaConf.to_container call in train.py before passing to resolve_preprocessing_config can go away, since it's currently converting a DictConfig to a plain dict just to have resolve_preprocessing_config treat it as one.

stepjam (approved):
LGTM w/c.

Also some additional minor points:

1. Unless I am mistaken, apply_preprocessing_configs re-validates allowed_data_types on every call.
This check already happens in resolve_preprocessing_config at startup. Running it again inside load_sample (which is called for every training sample in the hot path) is redundant overhead. The validation should stay in resolve_preprocessing_config only.

2. Only input_preprocessing_config is wired up at inference time.
In policy_inference.py the output preprocessing config from the archive is explicitly discarded with _. I know this is to come later, so please add a TODO.

neuracore/ml/utils/preprocessing_utils.py
ypang-neuraco: Echoing this and @sdas-neuraco 's comment. You can use pydantic to construct a preprocessing config class (see importer config as an example https://github.com/NeuracoreAI/neuracore_types/blob/d6073ab87d1d2507077c70eb77a61b56288ad760/neuracore_types/nc_data/__init__.py#L197)

You can use model validator to enforce the allowed data types and configs can be stored in the class so they don't need to be passed to a function every time you apply the preprocessing methods.

mark-neuracore: output preprocessing will not be passed to the inference at any time. my plan was to include output post processing in the next stage. it will invlolve a step to invert the used output preprocessing methods and a lot of other changes.

PR #529, opened 2026-04-08
feat: add recovery coverage for registration and reconnect flows

favour-neuraco:
### Features
 - Added coverage for registration retry/recovery and reconnect recovery behaviour

### Items
 - [NC-854](https://neuracore.atlassian.net/browse/NCP-854)


muneebneura (changes requested):
All good tests, can you also write a test that covers after we mark a trace registration as failed, those traces should re-appear in claim traces for registration call in the state manager.

muneebneura (approved):
LGTM

PR #536, opened 2026-04-13
fix: ensure daemon has bound its socket before signalling ready

muneebneura:
### Bugfixes

 - To avoid race conditions, 404 in get active data traces should be treated as empty list rather than a runtime error because we remove the traces subcollection and pending recording doc as soon as a recording is complete.
 - When start recording was called, the daemon subprocess was started but given 100ms before the producer immediately tried to send OPEN_RING_BUFFER over ZMQ, and since the daemon's auth HTTP call could take some time during init, sometimes its ZMQ socket wasn't bound yet, so zmq silently dropped the message and the daemon never knew a recording had started.
 
- Daemon Readiness is now determined by the appearance of the ZMQ socket file, Polling instead of sleeping a fixed interval avoids this race condition.

### Items

 - [NCP-886](https://neuracore.atlassian.net/browse/NCP-886)
 - [NCP-888](https://neuracore.atlassian.net/browse/NCP-888)

### Related PRs

 - None

### DIFF
--- a/neuracore/core/utils/backend_utils.py
+++ b/neuracore/core/utils/backend_utils.py
@@ -37,6 +37,8 @@ def get_active_data_traces(recording_id: str) -> list[RecordingDataTrace]:
         f"{API_URL}/org/{org_id}/recording/{recording_id}/traces/active",
         headers=get_auth().get_headers(),
     )
+    if response.status_code == 404:
+        return []
     response.raise_for_status()
     data = response.json() or []
     return [RecordingDataTrace.model_validate(item) for item in data]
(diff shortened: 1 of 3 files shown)


StevenJacobs61 (approved):
So long as this is fixing the issue (i.e. the race condition is absolutely unavoidable) then it is fine, though if there is a way to prevent the race condition for occurring then lets do that fix instead.

muneebneura:
> So long as this is fixing the issue (i.e. the race condition is absolutely unavoidable) then it is fine, though if there is a way to prevent the race condition for occurring then lets do that fix instead.

This does prevent the race condition case, because previously ensure daemon running returned before it was actually ready to receive messages. The race only existed because we handed control back to the caller too early.

PR #537, opened 2026-04-14
feat: error capture and log to bucket for endpoint

ypang-neuraco:
### Features
- Error capture for server startup similar to training job error capture
- Send local server policy logs to bucket if running in cloud

### Items
 - [NCRL-443](https://neuracore.atlassian.net/browse/NCRL-443)

### Related PRs
- https://github.com/NeuracoreAI/neuracore_backend/pull/349

### DIFF
--- a/neuracore/core/cli/launch_server.py
+++ b/neuracore/core/cli/launch_server.py
@@ -1,21 +1,43 @@
-"""Algorithm validation script for neuracore ML algorithms.
+"""CLI entrypoint for launching a local policy server.
 
-This module provides a command-line tool for validating ML algorithms in an
-isolated virtual environment. It creates a temporary venv, installs dependencies,
-and runs validation to ensure algorithms meet neuracore requirements.
+This module parses embodiment descriptions from JSON, authenticates with
+Neuracore, starts the local endpoint server process, and forwards startup
+failures to cloud endpoint error reporting when an endpoint ID is provided.
 """
 
 import json
 import logging
+import traceback
 
 import typer
 from neuracore_types import DataType
 
 import neuracore as nc
 from neuracore.core.endpoint import policy_local_server
+from neuracore.ml.utils.endpoint_storage_handler import EndpointStorageHandler
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+
+
+def _try_report_error_to_cloud(
+    endpoint_id: str, org_id: str | None, error_msg: str
+) -> None:
+    """Report endpoint startup failures to cloud storage without blocking shutdown.
+
+    This helper is called from the local launch error path after the stack trace
+    has already been logged. Any failure in this reporting flow is logged and
+    suppressed so the original launch exception still propagates.
+    """
+    try:
+        nc.login()
+        if org_id is not None:
+            nc.set_organization(org_id)
+        storage_handler = EndpointStorageHandler(endpoint_id=endpoint_id)
+        storage_handler.report_endpoint_error(error_msg)
+        logger.info("Successfully reported endpoint error to cloud.")
+    except Exception:
+        logger.error("Failed to report endpoint error to cloud.", exc_info=True)
 
 
 def run(
@@ -36,6 +58,9 @@ def run(
         ),
     ),
     job_id: str | None = typer.Option(None, "--job_id", help="Job ID to run"),
+    endpoint_id: str | None = typer.Option(
+        None, "--endpoint_id", help="Endpoint ID for log streaming"
+    ),
     org_id: str | None = typer.Option(None, "--org_id", help="Organization ID"),
     host: str = typer.Option("0.0.0.0", "--host", help="Host to bind the server"),
     port: int = typer.Option(8080, "--port", help="Port to bind the server"),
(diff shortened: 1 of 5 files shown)


neuracore/core/cli/launch_server.py
```diff
@@ -7,17 +7,34 @@
...
+from neuracore.ml.utils.endpoint_storage_handler import EndpointStorageHandler
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+def _try_report_error_to_cloud(
+    endpoint_id: str, org_id: str | None, error_msg: str
+) -> None:
+    """Best-effort attempt to report an endpoint error to the cloud."""
```
aditya-neuraco: Docstring can me made more formal.

neuracore/core/cli/launch_server.py
```diff
@@ -36,41 +53,54 @@ def run(
         ),
     ),
     job_id: str | None = typer.Option(None, "--job_id", help="Job ID to run"),
+    endpoint_id: str | None = typer.Option(
+        None, "--endpoint_id", help="Endpoint ID for log streaming"
+    ),
```
aditya-neuraco, line 63: We use endpoint name in policy server, but endpoint_id here. Is this intentional?

neuracore/core/cli/launch_server.py
```diff
@@ -36,41 +53,54 @@ def run(
...
-        input_embodiment_description=input_embodiment_description_map,
-        output_embodiment_description=output_embodiment_description_map,
-        train_run_name="",  # Use job id instead
-        port=port,
-        host=host,
-        job_id=job_id,
-    )
-    assert policy.server_process is not None
-    policy.server_process.wait()
+        try:
```
aditya-neuraco: We have a try inside an try? We can just raise in the original try block itself.

neuracore/core/endpoint.py
```diff
@@ -599,6 +655,7 @@ def policy_local_server(
         model_path=model_path,
         device=device,
         job_id=job_id,
+        endpoint_id=endpoint_id,
```
aditya-neuraco: This breaks existing behaviour. Can you use keyword args instead?

neuracore/core/utils/server.py
```diff
@@ -36,6 +38,37 @@ class CheckpointRequest(BaseModel):
     epoch: int
 
 
+def setup_server_logging(log_level: str, log_file: str | None = None) -> None:
+    """Configure structured logging for the server process."""
+    handlers: list[logging.Handler] = [logging.StreamHandler()]
+    if log_file is not None:
```
aditya-neuraco: Logfile is a file or the Path? We should type it and name it appropriately.

neuracore/core/utils/server.py
```diff
@@ -36,6 +38,37 @@ class CheckpointRequest(BaseModel):
...
+) -> None:
+    """Persist startup status so parent process can report errors."""
+    if startup_status_file is None:
+        return
+    payload: dict[str, str] = {"status": status}
+    if error is not None:
+        payload["error"] = error
+    path = Path(startup_status_file)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(payload), encoding="utf-8")
```
aditya-neuraco, line 69: What's this function for?

neuracore/core/utils/server.py
```diff
@@ -207,26 +259,46 @@ def start_server(
     parser.add_argument("--host", default="0.0.0.0", help="Host to bind to")
     parser.add_argument("--port", type=int, default=8080, help="Port to bind to")
     parser.add_argument("--log-level", default="info", help="Logging level")
+    parser.add_argument(
+        "--log-file",
+        required=False,
+        help="Optional path to write structured server logs.",
```
aditya-neuraco, line 265: Add this info in the function above that takes this argument.

neuracore/core/utils/server.py
```diff
@@ -207,26 +259,46 @@ def start_server(
...
+            DataType(k): v
+            for k, v in json.loads(str(args.input_embodiment_description)).items()
+        }
+        output_embodiment_description = {
+            DataType(k): v
+            for k, v in json.loads(str(args.output_embodiment_description)).items()
+        }
+        start_server(
+            input_embodiment_description=input_embodiment_description,
+            output_embodiment_description=output_embodiment_description,
```
aditya-neuraco, line 287: This needs to be cross embodiment right?

aditya-neuraco (approved):
LGTM w/ comments

neuracore/core/cli/launch_server.py
ypang-neuraco: Do you mean in policy_remote_server? This is used by the backend and not user facing so ID is more specific.

neuracore/core/utils/server.py
ypang-neuraco: Inference requires EmbodimentDescription instead of CrossEmbodimentDescription
https://github.com/NeuracoreAI/neuracore/blob/main/docs/model_construction.md

neuracore/core/cli/launch_server.py
aditya-neuraco: Oh okay. Then it's correct.

neuracore/core/utils/server.py
aditya-neuraco: It should be `CrossEmbodimentDescription` based right? There could be many non-identical robots in a scene,

ypang-neuraco: Currently we only support a single robot for inference. If we want to support `CrossEmbodimentDescription` for inference, this will require more thinking and is out of scope for this PR

neuracore/core/endpoint.py
ypang-neuraco: what do you mean? it is a kwarg

neuracore/core/utils/server.py
ypang-neuraco: We separate the error during endpoint startup and running. Since the server is launched in a subprocess, errors during startup will be saved to a file and the main process will read the file and send the content to the backend for reporting.

neuracore/core/utils/server.py
ypang-neuraco: renamed to log_file_path

neuracore/core/endpoint.py
ypang-neuraco: discussed offline, moved endpoint_id to the end

PR #540, opened 2026-04-15
fix: launch-server CLI docs

sdas-neuraco:
Fixes the broken launch-server JSON example in the CLI docs.

Also updates the CLI help text and adds a small test so the documented JSON shape keeps working.


ypang-neuraco (approved):
LGTM

aditya-neuraco (approved):
LGTM!

PR #543, opened 2026-04-15
feat: add guarded cache clear command

sdas-neuraco:
## Summary
- Add `neuracore cache clear` for safe local cache cleanup
- Show users exactly which disposable cache directories will be cleared before deletion
- Require confirmation unless `--yes` is passed
- Support `--dry-run` so users can preview disk cleanup without deleting files
- Keep training runs, auth config, and daemon state out of cache clearing

## User Experience
The command is built around a preview-first flow. Users see what will be deleted, how much disk space is involved, and what is explicitly protected before they confirm.

Default flow:
```bash
neuracore cache clear
```

Example prompt:
```text
The following disposable cache directories will be cleared:
  - recording cache: ~/.neuracore/training/recording_cache (3.8 GiB)
  - dataset statistics cache: ~/.neuracore/training/dataset_cache (24.0 KiB)
Total: 3.8 GiB
Training runs, auth config, and daemon state will not be deleted.
Clear these cache files? [y/N]
```

Preview-only flow:
```bash
neuracore cache clear --dry-run
```

Automation flow:
```bash
neuracore cache clear --yes
```

Scoped cleanup:
```bash
neuracore cache clear --recording-cache
neuracore cache clear --dataset-stats
```

### DIFF
--- a/neuracore/core/cli/cache_commands.py
+++ b/neuracore/core/cli/cache_commands.py
@@ -34,6 +34,8 @@ def _directory_size(path: Path) -> int:
     total = 0
     for entry in path.rglob("*"):
         try:
+            if entry.is_symlink():
+                continue
             if entry.is_file():
                 total += entry.stat().st_size
         except OSError:
--- a/neuracore/core/cli/cache_commands.py
+++ b/neuracore/core/cli/cache_commands.py
@@ -1,0 +1,131 @@
+"""Cache management CLI commands."""
+
+from __future__ import annotations
+
+import shutil
+from pathlib import Path
+
+import typer
+
+from neuracore.core.const import DEFAULT_CACHE_DIR, DEFAULT_RECORDING_CACHE_DIR
+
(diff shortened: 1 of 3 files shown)


neuracore/core/cli/cache_commands.py
```diff
@@ -0,0 +1,131 @@
...
+
+def _directory_size(path: Path) -> int:
+    """Return the total size of files under a directory."""
+    if not path.exists():
+        return 0
+
+    total = 0
+    for entry in path.rglob("*"):
+        try:
+            if entry.is_file():
```
favour-neuraco, line 39: _directory_size() may count symlinked files via is_file(). Might be worth skipping symlinks explicitly so the reported cache size only reflects files actually stored under the cache directory.

something like : `if entry.is_file() and not entry.is_symlink():`

favour-neuraco (approved):

favour-neuraco (approved):
LGTM

PR #544, opened 2026-04-15
fix: training spec resolution from data types

sdas-neuraco:
## Summary
- Treat empty input/output cross-embodiment configs as unset so data type configs can generate specs.
- Preserve dataset indexed sensor-name mappings when generating cross-embodiment descriptions from data types.
- Add unit coverage for explicit specs and data-type fallback behavior.

## Tests
- `pytest tests/unit/ml/test_train.py`
- Commit pre-commit hooks: pyupgrade, isort, black, ruff, pydocstyle, mypy, cspell


gtiboni-neuraco (approved):
LGTM

PR #545, opened 2026-04-15
fix: created dynamic ring buffer sizes per datatype

StevenJacobs61:
### Features

 -Created different chunk and ring buffer sizes for video and non video data types. This allows larger chunk sizes for video data making the system faster.


neuracore/data_daemon/communications_management/producer_channel.py
```diff
@@ -193,19 +256,21 @@ def _send(self, command: CommandType, payload: dict | None = None) -> None:
...
+                sequence_number = self._next_sequence_number
+                self._next_sequence_number += 1
+                self._last_enqueued_sequence_number = sequence_number
+            envelope = MessageEnvelope(
+                producer_id=self.channel_id,
+                command=command,
+                payload=payload or {},
+                sequence_number=sequence_number,
+            )
+            self._send_queue.put(envelope)
```
CougarTasker:
```suggestion
        with self._sequence_cv:
            sequence_number = self._next_sequence_number
            self._next_sequence_number += 1
            self._last_enqueued_sequence_number = sequence_number
            envelope = MessageEnvelope(
                producer_id=self.channel_id,
                command=command,
                payload=payload or {},
                sequence_number=sequence_number,
            )
            self._send_queue.put(envelope)
```

neuracore/data_daemon/communications_management/producer_channel.py
```diff
@@ -7,43 +7,109 @@
...
-        self._send_queue: queue.Queue[MessageEnvelope | None] = queue.Queue()
+        self._send_queue: queue.Queue[MessageEnvelope | None] = queue.Queue(
+            maxsize=self.send_queue_maxsize
+        )
         self._sender_thread: threading.Thread | None = None
         self._next_sequence_number = 1
         self._last_enqueued_sequence_number = 0
         self._last_socket_sent_sequence_number = 0
         self._sequence_cv = threading.Condition()
+        self._enqueue_lock = threading.Lock()
```
CougarTasker: issue: duplicate of `_sequence_cv`

CougarTasker (approved):
LGTM \w comment

neuracore/data_daemon/communications_management/producer_channel.py
StevenJacobs61: the first lock protect enqueuing order, the second protect the sequence number concurrency. this is making sequence number and queue allocation thread safe.

PR #547, opened 2026-04-16
fix: remove stale nc-data-daemon entry point

favour-neuraco:
remove stale nc-data-daemon entry point


damonhayhurst-neuraco (approved):
LGTM

PR #548, opened 2026-04-16
fix: capture cloud training setup logs

sdas-neuraco:
## Summary
- Configure training file logging immediately after Hydra config resolution so setup logs such as the run directory and resolved config are written to `train.log`.
- Move `CloudLogStreamer` ownership to `_main()` so cloud runs start uploading logs before dataset loading, validation, synchronization, and other pre-training setup work.
- Keep `run_training()` focused on training execution and metric logging, avoiding a second later log streamer.
- Add regression tests for logging-before-login, cloud log streaming before dataset loading, and streamer cleanup after pre-training failures.

### DIFF
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -777,16 +777,14 @@ def _main(cfg: DictConfig) -> None:
                 "Please specify only one."
             )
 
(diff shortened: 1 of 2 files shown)


ypang-neuraco (approved):
LGTM

PR #550, opened 2026-04-16
fix: align production integration tests with release tags

sdas-neuraco:
### Summary

This PR updates integration workflows so production tests run from the released source of truth, while staging continues to test `main`.

### Changes

- Production integration jobs now resolve the latest `v*` release tag and check out that tag for the test harness.
- Production integration jobs now install Neuracore from the checked-out release tag with `pip install .[...]`, instead of mixing a PyPI package with tests from `main`.
- Staging integration jobs continue to check out and install from `main`.
- Production jobs now verify that the checked-out commit exactly matches the expected release tag before running tests.
- Each integration workflow logs the environment, harness ref, checkout commit, test path, and installed `neuracore.__version__` for easier debugging.
- The split ML algorithm performance workflow now caches both the training job ID and the test harness ref, so the evaluate workflow uses the exact same release tag that started the training job.

### DIFF
--- a/.github/workflows/integration-ml-algorithm-performance-evaluate.yaml
+++ b/.github/workflows/integration-ml-algorithm-performance-evaluate.yaml
@@ -9,18 +9,71 @@ jobs:
   discover-algorithms:
     runs-on: ubuntu-latest
     outputs:
-      algorithms: ${{ steps.list.outputs.algorithms }}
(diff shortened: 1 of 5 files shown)


CougarTasker:
@damonhayhurst-neuraco Can you have a look at this?

CougarTasker (approved):
LGTM 👍

PR #551, opened 2026-04-16
chore: update neuracore readme badges

aditya-neuraco:
### Features

 - Made all shield follow the same format
 - Updated banner on readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,17 @@
+<!-- fullWidth: false tocVisible: false tableWrap: true -->
 <div align="center">
   <img src="./docs/assets/neuracore_logo.jpg" alt="Neuracore Logo" width="100%">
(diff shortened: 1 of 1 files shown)


docs/assets/neuracore_logo.png
sdas-neuraco, line 1: This image is repeated below, I dont think its worth it here, makes it seem very cluttered, I would just put the text in the middle of the screen and have that as the image

sdas-neuraco (dismissed):

docs/assets/neuracore_logo.png
aditya-neuraco: Removing the banner change. Only changing the shield badges.

damonhayhurst-neuraco (approved):

PR #554, opened 2026-04-16
fix: preserve local server embodiment ordering

sdas-neuraco:
## Summary
- Restore integer indices when the local model server parses JSON embodiment descriptions so joint/action ordering stays numeric after subprocess launch.
- Filter local server prediction sync-point data types to the declared input embodiment, matching direct `nc.policy` behavior.
- Add a regression test covering 12 indexed joints so string sorting would fail.


aditya-neuraco (approved):
LGTM!

PR #557, opened 2026-04-17
feat: check batch size is valid before training

gtiboni-neuraco:
### Features
 - When the user selects a specific batch size, this value is now checked using the BatchAutoTuner to make sure if fits into memory (GPU and RAM), before even starting training
 - This way the user gets a well described error
 
image

 - Perform batch size validation in subprocess to prevent CUDA OOM on main training process due to left over artefacts
 - Restore logging to console on train.py

 ### Notes:
 - In the future, we can run this check before even downloading the entire dataset, to save even more time for the user before finding out that the batch size was too high. Still, we need to download at least one sample on local disk to mimic the RAM and GPU memory usage.

### DIFF
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -339,25 +339,20 @@ def setup_logging(output_dir: str, rank: int = 0) -> None:
     """Setup logging configuration."""
     output_path = Path(output_dir)
     output_path.mkdir(parents=True, exist_ok=True)
+    stream_handler = logging.StreamHandler()
 
     if rank == 0:
         file_handler = logging.FileHandler(output_path / "train.log")
-        file_handler.setFormatter(JsonLineLogFormatter())
-        handlers: list[logging.Handler] = [file_handler]
-        logging.basicConfig(
-            level=logging.INFO,
-            handlers=handlers,
-            force=True,
-        )
     else:
         file_handler = logging.FileHandler(output_path / f"train-rank{rank}.log")
-        file_handler.setFormatter(JsonLineLogFormatter())
-        handlers = [file_handler]
-        logging.basicConfig(
-            level=logging.INFO,
-            handlers=handlers,
-            force=True,
-        )
+    file_handler.setFormatter(JsonLineLogFormatter())
+
+    handlers: list[logging.Handler] = [file_handler, stream_handler]
+    logging.basicConfig(
+        level=logging.INFO,
+        handlers=handlers,
+        force=True,
+    )
 
 
 def get_model_and_algorithm_config(
@@ -451,6 +446,9 @@ def assert_valid_batch_size(
             batch_size=batch_size,
             device=device,
         )
+    except Exception:
+        logger.error("Batch size validation failed", exc_info=True)
+        raise
     finally:
         del model
         del assert_dataset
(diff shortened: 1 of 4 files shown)


neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -421,3 +434,70 @@ def find_optimal_batch_size(
...
+            max_batch_size=batch_size,
+        )
+
+        valid = autotuner.is_valid_batch_size(batch_size)
+
+        return valid
+
+    except Exception:
+        logger.error("Batch size validation failed", exc_info=True)
+        raise
```
ypang-neuraco: We should abstract out the batch size checking logic into a new class e.g. BatchSizeValidator which can be reused by the BatchSizeAutotuner and the batch size checking in train.py

neuracore/ml/train.py
```diff
@@ -398,6 +401,74 @@ def get_model_and_algorithm_config(
...
+            torch.cuda.empty_cache()
+        gc.collect()
+
+    if not valid:
+        raise ValueError(
+            f"Batch size {batch_size} is not valid: it does not fit in "
+            "memory for the current algorithm, dataset, and GPU type. "
+            "Try a smaller batch size, or use batch_size='auto' to automatically "
+            "find the largest batch size that fits."
+        )
```
ypang-neuraco, line 465: raise exception inside BatchSizeValidator

ypang-neuraco (changes requested):

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -421,3 +434,70 @@ def find_optimal_batch_size(
...
+    )
+
+    try:
+        if batch_size > len(train_dataset):
+            logger.info(
+                "Batch size %d exceeds train dataset size %d; treating as invalid.",
+                batch_size,
+                len(train_dataset),
+            )
+            return False
```
ypang-neuraco: if batch size is greater than training set size, we can default to the training set size and validate it

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -19,6 +21,346 @@
...
+            e,
+            exc_info=True,
+        )
+        return False
+
+    except OutOfMemoryError as e:
+        logger.info("Batch size %s failed due to RAM OOM error: %s", batch_size, e)
+        return False
+
+    except Exception as e:  # noqa: BLE001
```
sdas-neuraco, line 273: Why do we return false here? if something goes wrong at this step we should crash and throw an error?

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -19,6 +21,346 @@
...
+        logger.error(
+            "RuntimeError while probing batch size %s: %s",
+            batch_size,
+            e,
+            exc_info=True,
+        )
+        return False
+
+    except OutOfMemoryError as e:
+        logger.info("Batch size %s failed due to RAM OOM error: %s", batch_size, e)
```
sdas-neuraco: logger should be an error? rather then info

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -19,6 +21,346 @@
...
+        return False
+
+    except Exception as e:  # noqa: BLE001
+        logger.error(
+            "Unexpected exception while probing batch size %s: %s",
+            batch_size,
+            e,
+            exc_info=True,
+        )
+        return False
```
sdas-neuraco: If i understand correctly, this will return false, and then tells the developer to reduce the batch_size or use auto, but then if the error is related to somewhere else like shape miss match or anything else it will till flag an incorrect next step for the user

sdas-neuraco (approved):
LGTM /w c

neuracore/ml/trainers/batch_autotuner.py
ypang-neuraco: updated to raise error and stop training if not failed due to OOM

ypang-neuraco (approved):

PR #561, opened 2026-04-20
fix: provide recording name instead of id in error when training on uneven sensor names

aditya-neuraco:
### Bugfixes

- `_format_failure_summary`: resolve recording IDs to human-readable names — check cache first, fetch via API on miss, fall back to ID on error
- `_raise_sync_failure`: centralize duplicate error-raise logic; `processed`/`total` default to `None` so 409 path doesn't pass explicit `None, None`
- 409 handler: parse failed recording IDs from error string and resolve names via `_raise_sync_failure`
- Error message now includes dataset name and lists `name (id)` per failed recording
- `synced_dataset.py`: gate `_perform_synced_data_prefetch` behind `_prefetch_videos_needed` to fix `ValueError: max_workers must be greater than 0` when `prefetch_videos=False`
- Tests: 4 new tests covering name-from-cache, name-from-API, ID-fallback, and 409 name resolution

### Items

 - [NCRL-420](https://neuracore.atlassian.net/browse/NCRL-420)

### DIFF
--- a/neuracore/core/data/dataset.py
+++ b/neuracore/core/data/dataset.py
@@ -389,12 +389,44 @@ class Dataset:
         )
         response.raise_for_status()
 
-    @staticmethod
-    def _format_failure_summary(failed_recording_ids: list[str]) -> str:
-        summary = ", ".join(failed_recording_ids[:3])
-        if len(failed_recording_ids) > 3:
-            summary += ", ..."
-        return summary
+    def _format_failure_summary(self, failed_recording_ids: list[str]) -> str:
+        id_to_name = {r.id: r.name for r in self._recordings_cache}
+        auth_headers = get_auth().get_headers()
+        for recording_id in failed_recording_ids:
+            if recording_id not in id_to_name:
+                try:
+                    response = requests.get(
+                        f"{API_URL}/org/{self.org_id}/recording/{recording_id}",
+                        headers=auth_headers,
+                    )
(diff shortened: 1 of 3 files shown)


neuracore/core/data/dataset.py
```diff
@@ -389,12 +389,46 @@ def delete(self) -> None:
...
+        processed: int | None = None,
+        total: int | None = None,
+    ) -> None:
+        failure_summary = self._format_failure_summary(failed_recording_ids)
+        progress_line = (
+            f"({processed}/{total} recordings synchronized).\n"
+            if processed is not None and total is not None
+            else ""
+        )
+        raise DatasetError(
```
sdas-neuraco, line 424: Maybe shape the error message like this: 

`Synchronization failed for dataset 'test_dataset'.

Problem recordings:
  - recording1 (rec1)

These recordings may have missing or extra sensor data, or the synchronization parameters may not match the data.
`

so it clearer and more legible to the user

sdas-neuraco: I would also completely hide the ID as this is not relevant and the user never knows the recording ID.

sdas-neuraco:
I would also set the entire sync to fail is any of the recordings cause problems, rather then continuing. But happy to be pushed back on this.

sdas-neuraco (approved):
LGTM w/ C

PR #564, opened 2026-04-21
fix: cancel recording and improvements

damonhayhurst-neuraco:
### Features

 - Better abstraction for cancel recording test case

### Bugfixes

 - Cancel recording was not stopping streams


favour-neuraco (approved):
LGTM

PR #566, opened 2026-04-21
fix: improve signal handling

damonhayhurst-neuraco:
### Features

 - remove oversompensation of tests on sigkill
 - add stop daemon log timing


StevenJacobs61 (approved):
LGTM

PR #567, opened 2026-04-21
fix: max streams 1000 and per data type

damonhayhurst-neuraco:
### Features

 - change max streams to 1000 and make it per data type


favour-neuraco (approved):
LGTM

PR #569, opened 2026-04-22
fix: looser assertions for synchronisation scenarios

damonhayhurst-neuraco:
### Features

 - loose assertions for sync cases in data daemon


aditya-neuraco (approved):
LGTM!

PR #571, opened 2026-04-23
test: stochastic tests for daemon and other improvements

damonhayhurst-neuraco:
### Features

Stochastic timestamp mode

- New TIMESTAMP_MODE_STOCHASTIC with ±50ms jitter applied per-frame using seeded random
- Timer.merge_stats — merges timer stats from subprocess workers
- Unexpected joints/cameras/markers now asserted
- nc.login() calls removed from individual tests, moved to fixture
- log_run_analysis consolidated with build_isolation_run_analysis
- run_case_contexts consolidated with assert and verify cases

### DIFF
--- a/.github/workflows/integration-platform.yaml
+++ b/.github/workflows/integration-platform.yaml
@@ -99,13 +99,6 @@ jobs:
         print(f"Installed neuracore version: {neuracore.__version__}")
         EOF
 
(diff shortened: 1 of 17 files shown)


aditya-neuraco (approved):
LGTM!

PR #573, opened 2026-04-23
test: pass all supported data types in training flow test

aditya-neuraco:
### Features

 - Pass all supported data types in training flow integration test.

### Items

 - [NCRL-390](https://neuracore.atlassian.net/browse/NCRL-390)

### DIFF
--- a/tests/integration/ml/test_training_flow.py
+++ b/tests/integration/ml/test_training_flow.py
@@ -7,13 +7,21 @@ inference, and remote endpoint deployment — all sharing one training job.
 
 import logging
 import os
+import pprint
 import sys
 import time
 import uuid
 
 import numpy as np
 import pytest
-from neuracore_types import DataType, JointData, RGBCameraData, SynchronizedPoint
+from neuracore_types import (
+    DataType,
+    EmbodimentDescription,
+    JointData,
+    LanguageData,
+    RGBCameraData,
+    SynchronizedPoint,
+)
 
 import neuracore as nc
 from neuracore.core.data.dataset import Dataset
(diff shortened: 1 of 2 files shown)


tests/integration/ml/test_training_flow.py
```diff
@@ -313,9 +399,10 @@ def test_training_flow():
...
-                assert (
-                    DataType.JOINT_POSITIONS in embodiment_description
-                ), f"JOINT_POSITIONS missing from robot {robot_id} data spec"
+                assert DataType.JOINT_POSITIONS in embodiment_description, (
+                    f"JOINT_POSITIONS missing from robot {robot_id}"
+                    " embodiment_description"
+                )
 
                 input_cross_embodiment_description[robot_id] = {
                     DataType.JOINT_POSITIONS: embodiment_description[
```
sdas-neuraco: We should assert all datatypes are logged correctly and available

tests/integration/ml/test_training_flow.py
sdas-neuraco, line 515: Training run only trains on a subset of data. IT should train on all the data.

sdas-neuraco (changes requested):
i think this should be a separate test in its self. Training flow test should be focused on one or two particular things, this now makes it very convoluted. The goal of the ticket was to check all data types can be logged and be used to train a model on which this currently does not do.

tests/integration/ml/test_training_flow.py
aditya-neuraco: Agrred, made the changes.

tests/integration/ml/test_training_flow.py
aditya-neuraco: Fixed now. Thanks for pointing thatout.

stepjam (approved):
LGTM

sdas-neuraco (approved):
LGTM

PR #574, opened 2026-04-23
fix: define output ordering for all-types validation fixture

sdas-neuraco:
## Summary
This fixes the `AllTypesModel` integration fixture so validation can initialize it when the model advertises all supported output data types.


stepjam (approved):
LGTM

PR #577, opened 2026-04-24
fix: daemon is stopped before next test run

muneebneura:
### Bugfixes

 - makes sure daemon is stopped before next test run

### Related PRs

 - None


damonhayhurst-neuraco (approved):
LGTM

PR #579, opened 2026-04-26
fix: pass model factory to batch size validator subprocess instead of…

stepjam:
## fix: resolve pickling error in batch size validator for transformer models

Fixes: https://github.com/NeuracoreAI/neuracore/pull/557

Fixes a `_pickle.PicklingError` that occurred when `BatchSizeValidator` tried to serialize transformer models (e.g. Pi0/Pi05) into a `spawn` subprocess. These models store `GemmaConfig` instances as attributes, which fail pickling due to a class identity mismatch across import paths in the transformers library.

`BatchSizeValidator._run_in_subprocess` passed the model instance directly as a `ctx.Process` arg. The `spawn` start method pickles all args before sending them to the subprocess, causing the error for any model with non-picklable attributes.                                                                                                                                 
                                                            
### Fix                                                                                                                                                                                                           
                                                            
Instead of passing a model instance, callers now pass a `model_factory:  Callable[[], NeuracoreModel]` + explicit `device`.  The subprocess constructs the model fresh by calling `model_factory()`. No model instance ever crosses the process boundary.

In `train.py`, the factory is `functools.partial(_create_model_for_batch_validation, cfg, model_init_description)`, a module-level function with picklable args, which uses the same Hydra instantiation path as the main training loop, including all algorithm-specific kwargs.

### DIFF
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -9,6 +9,7 @@ import re
 import sys
 import time
 import traceback
+from functools import partial
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


ypang-neuraco (approved):
LGTM

PR #584, opened 2026-04-29
fix: reduce fds for daemon db connections

damonhayhurst-neuraco:
### Features

 - Reduce number of loose file descriptors that access sqlite. 

### Justification

 - all read/writes are atomic

### DIFF
--- a/neuracore/data_daemon/state_management/state_store_sqlite.py
+++ b/neuracore/data_daemon/state_management/state_store_sqlite.py
@@ -59,6 +59,8 @@ class SqliteStateStore(StateStore):
         self._engine: AsyncEngine = create_async_engine(
             f"sqlite+aiosqlite:///{db_path}",
             future=True,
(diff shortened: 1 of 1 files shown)


neuracore/data_daemon/state_management/state_store_sqlite.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, db_path: Path) -> None:
         self._engine: AsyncEngine = create_async_engine(
             f"sqlite+aiosqlite:///{db_path}",
             future=True,
+            pool_size=1,
```
CougarTasker, line 62: Question: can you check this pool size and how it effects performance, try a few values and see how it effects perfomace timings

CougarTasker (approved):
LGTM \w comment

damonhayhurst-neuraco:
Addressed in #623

PR #587, opened 2026-04-29
fix: stop polling finalized recording durations

muneebneura:
### Bugfixes

 - This fixes the cloud data integrity wait logic. Before, the test kept polling until a recording cloud duration matched the expected test duration. But once the backend sets end_time, that value does not change, so waiting longer cannot fix a bad duration.
 
The test now only waits until each recording has an end_time. After that, the normal duration check runs and fails right away if the duration is wrong

### Items

 - [NCP-928](https://neuracore.atlassian.net/browse/NCP-928)


damonhayhurst-neuraco (approved):
LGTM

PR #590, opened 2026-04-29
fix: retry stale keep-alive connections via shared HTTP session

damonhayhurst-neuraco:
### Features

 - Session backed by a HTTPAdapter so that stale keep-alive sockets are transparently retried instead of surfacing as an SSLError

### Bugfixes
- Data daemon tests are failing from depleted connection pool


CougarTasker (approved):
LGTM 👍

PR #591, opened 2026-04-30
refactor: move helper functions into utils.py in groot

kwangneuraco:
move helper functions into utils.py in groot.
Fix comments from Mark

### DIFF
--- a/neuracore/ml/algorithms/groot/modules.py
+++ b/neuracore/ml/algorithms/groot/modules.py
@@ -20,11 +20,7 @@ Reference: NVIDIA GR00T N1.6 technical report.
 
 from __future__ import annotations
 
-import json
 import logging
-import os
-from contextlib import AbstractContextManager, nullcontext
-from typing import Any
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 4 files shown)


neuracore/ml/algorithms/groot/modules.py
```diff
@@ -35,46 +31,18 @@
...
-# Re-export dataclass so callers can customise the bundled Eagle config
-# without touching JSON files.
-from .eagle_config.eagle3_vl_dataconfig import Eagle3VLDataConfig  # noqa: E402
+from .utils import (
+    _build_eagle_config_from_dataclass,
+    _get_attn_implementation,
+    _load_config_json,
+    _load_pretrained_state_dict,
+    _sdpa_context,
+)
```
aditya-neuraco, line 41: usually we want helpers to start with `_` if they are in the same file. if they are imported from a utils file, we can remove the `_`

kwangneuraco: Good spot

mark-neuracore (approved):
LGTM. Makes navigating through files much easier

mark-neuracore:
Something we will need to consider later in another refactor is unified parameters configurations. They here use dataclasses.

It would be cleaner but it means more refactor specifically when migrating implemented algorithms from other repos

kwangneuraco:
> Something we will need to consider later in another refactor is unified parameters configurations. They here use dataclasses.
> 
> It would be cleaner but it means more refactor specifically when migrating implemented algorithms from other repos



> Something we will need to consider later in another refactor is unified parameters configurations. They here use dataclasses.
> 
> It would be cleaner but it means more refactor specifically when migrating implemented algorithms from other repos

On the way!

PR #592, opened 2026-04-30
test: improved timer for daemon tests

damonhayhurst-neuraco:
### Features

 - New timer specify which function calls to assert limit on
- Overall wall clock time of test added
- Improved logging

### DIFF
--- a/tests/integration/platform/data_daemon/shared/timer.py
+++ b/tests/integration/platform/data_daemon/shared/timer.py
@@ -1,0 +1,231 @@
+"""Timer primitives for daemon integration tests.
+
+Provides :class:`TimerSpec`, :class:`Timer`, the per-label registry, and the
+shared timing constants used across the suite.  Sits at the bottom of the
+import graph so any other shared module may depend on it without cycles.
+"""
+
+from __future__ import annotations
+
+import logging
+import time
+from dataclasses import dataclass
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Timing constants
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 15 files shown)


tests/integration/platform/data_daemon/shared/timer.py
```diff
@@ -0,0 +1,231 @@
...
+        "nc.log_custom_1d": TimerSpec(max_time=MAX_TIME_TO_LOG_S),
+        "nc.log_rgb": TimerSpec(max_time=MAX_TIME_TO_LOG_S),
+        # Daemon control
+        "stop_daemon_escalated": TimerSpec(max_time=5.0),
+        "stop_daemon[cli]": TimerSpec(max_time=15.0),
+        "stop_daemon[sigterm]": TimerSpec(max_time=15.0),
+        "stop_daemon[sigint]": TimerSpec(max_time=15.0),
+        "stop_daemon[sigkill]": TimerSpec(max_time=15.0),
+        # DB readiness — tests pass an explicit max_time= override
+        "daemon.offline_db_ready": TimerSpec(max_time=HIGH_TIME_TO_DATASET_READY_S),
```
CougarTasker, line 133: Question: why do we need this?
Issue: this seems un-necessarily complex de-coupling this seems to just make it more difficult to debug and analyse the tests

before you would just need to look at the timer to understand its configuration now:
 - how you label it will change its functionality
 - you need to look at multiple files:  what is the registry config have to hunt and remember any assert_labels changes

CougarTasker (changes requested):
This just seems to complicate the test while providing no improvement.

CougarTasker:
These changes were not required.

PR #595, opened 2026-05-01
fix: shared ring buffer transport

StevenJacobs61:
### Features
 - Refactors the daemon shared-memory transport into clearer producer, consumer, and shared-transport modules.
 - Introduces daemon-owned fixed shared-slot transport for producer/daemon communication.
 - Adds shared-memory budgeting, registry, and runtime helpers to support the new transport flow.
 -  Adds an RGB spool for the recording disk manager to read directly from and encoding can happen without building full frames into bytes to be queued and copied.-


neuracore/data_daemon/communications_management/consumer/completion_worker.py
```diff
@@ -0,0 +1,269 @@
...
+        producer_id: str,
+        trace_id: str,
+        recording_id: str,
+        data_type: DataType,
+        metadata: dict[str, str | int | None],
+        data: bytes,
+        final_chunk: bool = False,
+    ) -> None:
+        robot_instance = int(metadata.get("robot_instance") or 0)
+        time.monotonic()
```
favour-neuraco: `time.monotonic()`  is not assigned or used, also remove the import

neuracore/api/logging.py
```diff
@@ -165,32 +188,81 @@ def _log_single_joint_data(
...
     StreamManagerOrchestrator().get_provider_manager(
         robot.id, robot.instance
-    ).get_json_source(str_id, data_type, sensor_key=str_id).publish(
-        data.model_dump(mode="json")
+    ).get_json_source(
+        joint_stream_binding.stream_id,
+        data_type,
+        sensor_key=joint_stream_binding.stream_id,
+    ).publish(data.model_dump(mode="json"))
+    _publish_json_to_p2p(robot, joint_stream_binding.stream_id, data_type, data)
```
favour-neuraco, line 246: This looks like a duplicate P2P publish. Both this manual `get_json_source(...).publish(...)` call and `_publish_json_to_p2p(...)` use the same stream id as `sensor_key`, so they resolve to the same cached JSON source and publish the same joint payload twice.

Can we remove the manual publish block and keep `_publish_json_to_p2p(...)`?

neuracore/data_daemon/communications_management/consumer/completion_worker.py
favour-neuraco: `time.monotonic()` is not assigned or used here. Can we remove this leftover debug line and the `time` import if nothing else uses it?

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,650 @@
...
+            raise RuntimeError("Daemon is already running")
+
+        self._running = True
+        self.comm.start_consumer()
+
+        logger.info("Daemon started and ready to receive messages...")
+        try:
+            last_receive_log_at = datetime.now(timezone.utc)
+            datetime.now(timezone.utc)
+            loop_count = 0
```
favour-neuraco: `datetime.now(timezone.utc)` is called without using the result, and `loop_count` is incremented but never read.

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,650 @@
...
+        :param trace_id: The trace ID that the message belongs to.
+        :param data_type: The data type of the message payload.
+        :param data: The message data.
+        :param recording_id: The recording ID (from immutable _trace_recordings).
+        :param final_chunk: Whether this is the final chunk for the trace.
+        """
+        metadata = self._trace_lifecycle.get_trace_metadata(trace_id)
+        robot_instance = int(metadata.get("robot_instance") or 0)
+        try:
+            time.monotonic()
```
favour-neuraco: time.monotonic() is called but not assigned or used.

neuracore/core/robot.py
favour-neuraco: Small nit: this looks off by one. Because the count is checked before incrementing, `> MAX_DATA_STREAMS` allows `MAX_DATA_STREAMS + 1` streams.

Should this be `>= MAX_DATA_STREAMS`?

neuracore/core/streaming/data_stream.py
```diff
@@ -108,28 +108,27 @@ def _handle_ensure_producer_channel(self, context: DataRecordingContext) -> None
...
+    def stop_recording(self, *, wait_for_drain: bool = True) -> None:
+        """Stop recording data and tear down the active producer, if any."""
         self._recording = False
-        if (
-            isinstance(self._producer_channel, ProducerChannel)
-            and self._producer_channel.trace_id
-        ):
-            self._producer_channel.cleanup_producer_channel()
+        producer_channel = self._producer_channel
+        self._producer_channel = None
```
favour-neuraco, line 121: After `stop_recording()`, `_recording` is set to `False` and `_producer_channel` is cleared, but `_context` still points at the previous recording.

Should we also clear `_context` here so `get_recording_context()` does not return stale recording metadata after the stream has stopped?

neuracore/data_daemon/communications_management/shared_transport/shared_slot_transport.py
```diff
@@ -0,0 +1,504 @@
...
+    """Read a float override from the environment, falling back safely."""
+    raw = os.getenv(name)
+    if raw is None:
+        return float(default)
+    try:
+        return float(raw)
+    except ValueError:
+        logger.warning("Invalid %s=%r; using default %s", name, raw, default)
+        return float(default)
+
```
muneebneura: I have seen this in registry as well. You can move this to the shared transport runtime and import from there in both places.

neuracore/data_daemon/communications_management/consumer/bridge_chunk_spool.py
```diff
@@ -0,0 +1,132 @@
...
+        with self._lock:
+            if (
+                self._current_segment_size > 0
+                and self._current_segment_size + chunk_len > self._segment_max_bytes
+            ):
+                self._rotate_segment_locked()
+
+            offset = self._current_segment_size
+            self._current_segment_path.parent.mkdir(parents=True, exist_ok=True)
+            with self._current_segment_path.open("ab") as handle:
```
CougarTasker: Issue: performance, we open the fille for every chunk, The file handle should be kept open for the segment's lifetime.

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,650 @@
...
+
+        :param trace_id: ID of the trace to clean up
+        :param bytes_written: total number of bytes written for the trace (unused)
+        """
+        self._trace_lifecycle.cleanup_trace_written(trace_id)
+
+        channel = next(
+            (ch for ch in self.channels.values() if ch.trace_id == trace_id),
+            None,
+        )
```
CougarTasker: Issue: performance, a scan here is not ideal it would be best if `ChannelRegistry` also tracked by `trace_id` not just producer

neuracore/data_daemon/recording_encoding_disk_manager/core/rgb_trace_spool.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def append_frame(
+        self,
+        *,
+        trace_key: _TraceKey,
+        trace_dir: Path,
+        frame_bytes: bytes,
+    ) -> _RGBFrameRef:
+        """Append one raw frame to the trace's spool file and return its ref."""
+        del trace_key
```
muneebneura: What is the purpose of doing this? param dicarded. I think we can remove this as the path is determined by trace dir.

tests/integration/platform/test_recording_wait_false.py
```diff
@@ -0,0 +1,809 @@
...
+        subprocess.run(["bash", str(cleanup_script)], cwd=REPO_ROOT, check=True)
+
+        for key, previous_value in previous_env.items():
+            if previous_value is None:
+                os.environ.pop(key, None)
+            else:
+                os.environ[key] = previous_value
+
+
+def _run_wait_false_rgb_payload_test(
```
damonhayhurst-neuraco: wait false tests are already covered. This file repeats a lot of functionality.

neuracore/core/streaming/data_stream.py
```diff
@@ -108,28 +108,27 @@ def _handle_ensure_producer_channel(self, context: DataRecordingContext) -> None
...
+        if not isinstance(producer_channel, ProducerChannel):
+            return
+
+        try:
+            if producer_channel.trace_id:
+                producer_channel.cleanup_producer_channel(
+                    wait_for_slot_drain=wait_for_drain
+                )
+        finally:
+            producer_channel.stop_producer_channel()
```
favour-neuraco: Question: Is `wait=False` expected to avoid waiting for RGB shared-slot drain?

`cleanup_producer_channel(wait_for_slot_drain=False)` skips one `wait_until_drained()` call, but `DataStream.stop_recording()` always calls `stop_producer_channel()` afterwards.

For RGB streams, `_shared_slot_transport` should still exist at that point, so `stop_producer_channel()` still calls `wait_until_drained(timeout_s=30.0)`. Is that intentional?

neuracore/data_daemon/communications_management/producer/producer_channel.py
```diff
@@ -167,13 +207,34 @@ def end_trace(self) -> None:
...
+            command,
+            payload,
+            sequence_number=sequence_number,
+            on_failed_send=(
+                self._shared_slot_transport.notify_sender_failure
+                if self._use_shared_slot_transport
+                and self._shared_slot_transport is not None
+                else None
+            ),
+        )
```
muneebneura: Can we avoid repeating the same conditions here?

neuracore/data_daemon/communications_management/producer/producer_transport_debug_models.py
muneebneura, line 1: Why do we need this file which just re exports models?

neuracore/data_daemon/recording_encoding_disk_manager/workers/raw_batch_writer.py
```diff
@@ -16,7 +17,15 @@
 )
 
 from ..core.trace_filesystem import _TraceFilesystem
-from ..core.types import _BatchJob, _TraceKey, _WriteState
+from ..core.types import (
+    _BatchJob,
```
damonhayhurst-neuraco: why are these types indicated private and then you are importing them here?

neuracore/data_daemon/communications_management/consumer/spool_worker.py
```diff
@@ -0,0 +1,266 @@
...
+                chunk_spool_ref = None
+                self._mark_sequence_completed(
+                    SharedSlotSequenceProgressRequest(
+                        producer_id=work.channel.producer_id,
+                        sequence_number=descriptor.sequence_id,
+                    )
+                )
+                return
+
+            work.channel.set_trace_id(trace_id)
```
CougarTasker: Question: why do we need this?
Issue: does this not conflict with the `trace_id` coming from the data bridge. e.g. with back to back traces if the spool processor is slower than the data bridge the channel could have moved onto the next trace but the spool would still be processing the previous trace?

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
```diff
@@ -0,0 +1,360 @@
...
+            )
+        except Exception:
+            logger.exception(
+                "Shared-slot copy failed " "producer_id=%s sequence_id=%s slot_id=%s",
+                channel.producer_id,
+                descriptor.sequence_id,
+                descriptor.slot_id,
+            )
+            raise
+        self._send_slot_credit_return(channel, descriptor)
```
CougarTasker: Question: should we not release the slot credit on failure too? 

Suggestion: move this to a finally block

neuracore/api/logging.py
StevenJacobs61: Yes removed

neuracore/data_daemon/communications_management/consumer/completion_worker.py
StevenJacobs61: Yes removed

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
```diff
@@ -0,0 +1,360 @@
...
+            )
+            chunk_view = packet_view[chunk_start:chunk_end]
+            try:
+                chunk_spool_ref = chunk_spool.append(chunk_view)
+            finally:
+                chunk_view.release()
+            return metadata, chunk_spool_ref
+        finally:
+            packet_view.release()
+
```
muneebneura, line 290: The credit return is blocked by this file IO here which will result in slot exhaustion under disk pressure. Do we need to wait for disk write here bcz the shm slot only needs to be held for the copy, not disk write. You don't have to address this if we strictly need to wait for write before credit return

neuracore/data_daemon/communications_management/consumer/data_bridge.py
StevenJacobs61: yes removed

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,650 @@
...
+
+        logger.info("Daemon started and ready to receive messages...")
+        try:
+            last_receive_log_at = datetime.now(timezone.utc)
+            datetime.now(timezone.utc)
+            loop_count = 0
+
+            while self._running:
+                loop_count += 1
+                self._trace_lifecycle.finalize_closing_recordings()
```
CougarTasker: Suggestion: doing this every loop is inefficient this should be event driven

neuracore/data_daemon/communications_management/consumer/data_bridge.py
StevenJacobs61: yes removed, sorry yes leftovers from debugging

neuracore/data_daemon/communications_management/shared_transport/registry.py
```diff
@@ -0,0 +1,720 @@
...
+def _env_float(name: str, default: float) -> float:
+    """Read a float override from the environment, falling back safely."""
+    raw = os.getenv(name)
+    if raw is None:
+        return float(default)
+    try:
+        return float(raw)
+    except ValueError:
+        logger.warning("Invalid %s=%r; using default %s", name, raw, default)
+        return float(default)
```
CougarTasker: Issue: Duplicate code 

shared_transport/shared_slot_transport.py:64

neuracore/data_daemon/communications_management/shared_transport/registry.py
```diff
@@ -0,0 +1,720 @@
...
+                            if allocate_timeout_s is None
+                            else allocate_timeout_s
+                        ),
+                    ),
+                )
+            cls._refcount += 1
+            return cls._instance
+
+    @classmethod
+    def release_shared_instance(cls) -> None:
```
CougarTasker: Issue: dead code

neuracore/data_daemon/communications_management/shared_transport/registry.py
```diff
@@ -0,0 +1,720 @@
...
+                return
+            cls._refcount = max(0, cls._refcount - 1)
+            if cls._refcount > 0:
+                return
+            instance = cls._instance
+            cls._instance = None
+        instance.close()
+
+    @classmethod
+    def reset_shared_instance_for_tests(cls) -> None:
```
CougarTasker: Issue: this is not an ideal pattern please use fixtures instead, I don't like test fixtures leaking into the codebase

neuracore/core/robot.py
StevenJacobs61: Potentially, i see your thought, it would 1 more than the max in this case as being equal to the max would allow it to increment, this is not a change i have made in this PR the code is much older, but I can add that update also

tests/integration/platform/test_recording_wait_false.py
StevenJacobs61: Sorry file shouldn't be here I will remove it, you are right to point this out

neuracore/data_daemon/communications_management/shared_transport/shared_slot_transport.py
```diff
@@ -0,0 +1,504 @@
...
+    def acquire(cls, registry: SharedSlotRegistry) -> SharedSlotVideoWorker:
+        """Acquire a singleton worker instance for isolated unit tests."""
+        with cls._instance_lock:
+            if cls._instance is None:
+                cls._instance = cls(registry)
+            cls._refcount += 1
+            return cls._instance
+
+    @classmethod
+    def release_shared_instance(cls) -> None:
```
CougarTasker, line 135: Issue: dead code

neuracore/data_daemon/communications_management/shared_transport/shared_slot_transport.py
```diff
@@ -0,0 +1,504 @@
...
+                return
+            cls._refcount = max(0, cls._refcount - 1)
+            if cls._refcount > 0:
+                return
+            instance = cls._instance
+            cls._instance = None
+        instance.close()
+
+    @classmethod
+    def reset_shared_instance_for_tests(cls) -> None:
```
CougarTasker, line 148: Issue: this is not an ideal pattern please use fixtures instead, I don't like test fixtures leaking into the codebase

neuracore/data_daemon/communications_management/consumer/models.py
```diff
@@ -0,0 +1,743 @@
...
+    last_sequence_number: int = 0
+    opened_at: datetime | None = None
+    heartbeat_expired_at: datetime | None = None
+    transport_mode: TransportMode = TransportMode.NONE
+    socket_pending_messages: dict[str, PartialMessage] = field(default_factory=dict)
+    shared_slot: SharedSlotTransportState = field(
+        default_factory=SharedSlotTransportState
+    )
+
+    def is_opened(self) -> bool:
```
CougarTasker: Issue: dead code, duplicates `is_open`

neuracore/data_daemon/communications_management/consumer/models.py
```diff
@@ -0,0 +1,743 @@
...
+        """Return whether any transport for this channel has been opened."""
+        return self.transport_mode is not TransportMode.NONE
+
+    def touch(self) -> None:
+        """Refresh heartbeat tracking for this channel."""
+        self.last_heartbeat = datetime.now(timezone.utc)
+        self.heartbeat_expired_at = None
+
+    def is_open(self) -> bool:
```
CougarTasker, line 439: Issue: dead code only used in two tests
Suggestion: this is very simple please remove

neuracore/data_daemon/communications_management/shared_transport/shared_slot_transport.py
```diff
@@ -0,0 +1,504 @@
...
+            in_flight_slot_count=_int_snapshot_value(snapshot, "in_flight_slot_count"),
+            max_in_flight_slot_count=_int_snapshot_value(
+                snapshot, "max_in_flight_slot_count"
+            ),
+            acked_sequence_count=_int_snapshot_value(snapshot, "acked_sequence_count"),
+            ack_timeout_count=_int_snapshot_value(snapshot, "ack_timeout_count"),
+            worker_queue_qsize=self._worker._queue.qsize(),
+            worker_queue_maxsize=self._worker._queue.maxsize,
+            worker_thread_alive=self._worker._thread.is_alive(),
+            worker_error=(
```
CougarTasker: Issue: accessing private fields `_queue` or `_thread` from outside the class

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,637 @@
...
+        :return: None
+        """
+        if self._running:
+            raise RuntimeError("Daemon is already running")
+
+        self._running = True
+        self.comm.start_consumer()
+
+        logger.info("Daemon started and ready to receive messages...")
+        try:
```
CougarTasker, line 141: Issue: dead code `last_receive_log_at` and `datetime.now(timezone.utc)` are not used
```suggestion
```

neuracore/data_daemon/const.py
muneebneura, line 1: Remove the orphaned constants from here which were previously used by channel reader

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,637 @@
...
+            self._closed_producers.contains(producer_id)
+            and cmd != CommandType.OPEN_FIXED_SHARED_SLOTS
+        ):
+            return
+
+        if (
+            cmd == CommandType.OPEN_FIXED_SHARED_SLOTS
+            and self._closed_producers.contains(producer_id)
+        ):
+            self._closed_producers.discard(producer_id)
```
CougarTasker: Issue: dead code the guard condition above is the same so this is never reached
Question: did you intend to discard this producer?

neuracore/data_daemon/communications_management/consumer/data_bridge.py
```diff
@@ -0,0 +1,637 @@
...
+                    state.heartbeat_expired_at = now
+                continue
+
+            to_remove.append(producer_id)
+
+        for producer_id in to_remove:
+            channel = self.channels.get(producer_id)
+            if channel is None:
+                continue
+            if channel is not None and channel.trace_id is not None:
```
CougarTasker: Suggestion: We skip if the channel is none 

```suggestion
            if channel.trace_id is not None:
```

neuracore/data_daemon/communications_management/shared_transport/registry.py
```diff
@@ -0,0 +1,720 @@
...
+    SharedSlotRegistryState,
+)
+from .runtime import create_control_socket_path
+
+logger = logging.getLogger(__name__)
+
+
+def _should_log_credit_event(sequence_id: int) -> bool:
+    """Sample per-sequence investigation logs without hiding early activity."""
+    return sequence_id <= 8 or sequence_id % 25 == 0
```
CougarTasker: Issue: magic numbers, do we really need to sample the first 8?
Suggestion: move this inline and use a constant

CougarTasker (commented):
Some comments so far

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
```diff
@@ -0,0 +1,360 @@
...
+    def _shared_slot_packet_view(self, descriptor: SharedSlotDescriptor) -> memoryview:
+        """Return one packet view out of cached shared memory."""
+        shm = self._shared_memory_cache.get(descriptor.shm_name)
+        if shm is None:
+            shm = SharedMemory(name=descriptor.shm_name, create=False)
```
CougarTasker: Question: We don't call reserve on this so does that not mean the budget will be out of sync?

neuracore/core/streaming/data_stream.py
StevenJacobs61: Yes, you caught it still hard waiting even if we have explicitly passed false previously, so I have passed that var down to the stop_producer_channel now also to reflect this intended not waiting behaviour, as we do not need to wait for the slot to be drained if wait=False as the shared slot memory is owned by the daemon process now.

neuracore/data_daemon/recording_encoding_disk_manager/workers/raw_batch_writer.py
StevenJacobs61: Yes will make them public as moved them there to clean up the file but didn't make them public again

neuracore/data_daemon/communications_management/shared_transport/runtime.py
```diff
@@ -0,0 +1,20 @@
...
+from __future__ import annotations
+
+import os
+import uuid
+from pathlib import Path
+
+from neuracore.data_daemon.const import ACK_BASE_DIR
+
+
+def create_control_socket_path(base_dir: Path = ACK_BASE_DIR) -> Path:
```
CougarTasker: Question: Does this need to be its own file can this be a method in the `SharedSlotRegistry` as its only used once there.

neuracore/data_daemon/communications_management/producer/producer_channel.py
```diff
@@ -167,13 +207,34 @@ def end_trace(self) -> None:
     def stop_producer_channel(self) -> None:
         """Stops the producer channel and cleans up any associated resources."""
         self._stop_heartbeat_service()
+
+        # Ensure all descriptors were actually sent
+        cutoff_sequence = self.get_last_enqueued_sequence_number()
+        self.wait_until_sequence_sent(cutoff_sequence)
```
favour-neuraco: Question: Should we check the return value from `wait_until_sequence_sent()` during producer cleanup?

In `cleanup_producer_channel()` and `stop_producer_channel()`, we call `wait_until_sequence_sent(...)` but ignore the boolean result. Since this returns `False` if the sender thread failed/stopped before sending up to that sequence, should cleanup raise/log/mark the transport unhealthy instead of continuing as if the sequence was sent?

This seems important because shutdown/finalisation relies on `TRACE_END` and the stop cutoff sequence actually reaching the daemon.

neuracore/data_daemon/communications_management/producer/producer_channel.py
```diff
@@ -351,24 +481,47 @@ def send_data_parts(
...
+        When ``wait_for_slot_drain`` is False, this returns after the producer has
+        pushed all currently queued shared-slot descriptors and the TRACE_END
+        control message onto the socket. Slot credits may still return
+        asynchronously while the channel remains alive for later cleanup.
+        """
+        if self._shared_slot_transport is not None:
+            self._shared_slot_transport.wait_until_payload_handed_off(timeout_s=30.0)
+
+        cutoff_sequence = self.get_last_enqueued_sequence_number()
+        self.wait_until_sequence_sent(cutoff_sequence)
```
favour-neuraco: Same as the earlier comment: should we check the boolean return value here rather than ignoring a failed `wait_until_sequence_sent()`?

neuracore/data_daemon/communications_management/consumer/bridge_chunk_spool.py
StevenJacobs61: Indeed!

neuracore/data_daemon/lifecycle/runtime_recovery.py
```diff
@@ -44,22 +147,30 @@ def validate_or_recover_sqlite(db_path: Path, *, recover: bool = True) -> bool:
...
+    )
+
     return False
 
 
 def checkpoint_sqlite(db_path: Path) -> None:
     """Checkpoint SQLite WAL to disk."""
     if not db_path.exists():
         return
+
```
favour-neuraco, line 173: nit:  empty white lines

neuracore/data_daemon/communications_management/consumer/trace_lifecycle_coordinator.py
```diff
@@ -0,0 +1,441 @@
...
+        self._closing_recordings.mark_closing(
+            recording_id,
+            RecordingClosingState(
+                producer_stop_sequence_numbers=producer_stop_sequence_numbers,
+                stop_requested_at=utc_now(),
+            ),
+        )
+        self._emitter.emit(Emitter.STOP_RECORDING_REQUESTED, recording_id)
+
+    def finalize_closing_recordings(self) -> None:
```
damonhayhurst-neuraco, line 375: Issue: This function is not unit tested

muneebneura (commented):
Comments so far

neuracore/data_daemon/communications_management/consumer/spool_worker.py
```diff
@@ -0,0 +1,270 @@
...
+        self._chunk_spool.cleanup()
+
+    def _ensure_running(self) -> None:
+        with self._error_lock:
+            if self._error is not None:
+                raise RuntimeError("Daemon spool shard failed") from self._error
+        if not self._thread.is_alive():
+            raise RuntimeError("Daemon spool shard is not running")
+
+    def _worker_loop(self) -> None:
```
damonhayhurst-neuraco, line 89: Issue: There are no unit tests for this

neuracore/data_daemon/communications_management/consumer/spool_worker.py
StevenJacobs61: No this does not corrupt the data or cause an conflict as the flow continues to use the trace id send from the producer, so this is internal bookkeeping essentially regarding 'last seen trace id in this producer channel', and it propagates that trace ID. Technically we could keep a track of in flight traces for that producer if say we have some irregularity in the ordering or timing, for example we get trace A, then trace B then trace A again and then trace B, but we don't need to maintain this state for any behavioural correctness, so keeping this state consistent in terms of what trace we had ( not yet finished) vs what we are currently on would be more explicit, but maybe juts unnecessary overhead as there is not a risk of conflict between the two traces as we always use the one coming from the prod channel. We could make it consistent and explicit, but this might just be more buggy and ultimately perhaps pointless to manage as it doesn't affect behaviour at all. Let me know if you disagree.

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
StevenJacobs61: Yes agreed

neuracore/data_daemon/communications_management/consumer/data_bridge.py
StevenJacobs61: Good point, this actually might save a lot of computing. So now it is only done after the stop recording event comes though, cleanup after trace written or trace end. We still have to check if we have reached sequence cutoff to call this function each sequence number after the stop is recorded but that is much cheaper and event driven compared to the current polling.

StevenJacobs61: Let me know if the new solution is good and i will resolve this issue.

neuracore/data_daemon/communications_management/shared_transport/registry.py
StevenJacobs61: Agreed, leftover, removed.

neuracore/data_daemon/communications_management/consumer/data_bridge.py
StevenJacobs61: They are not dublicates one is == the other is !=

neuracore/data_daemon/communications_management/shared_transport/registry.py
StevenJacobs61: debug leftover, removed

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
StevenJacobs61: Reserve is called in the open function, this just attaches to it

neuracore/data_daemon/communications_management/producer/producer_channel.py
StevenJacobs61: YEs, I have updated this so cleanup_producer_channel now raises if the cutoff/trace_end is not actually sent, stop_producer_channel logs and raises after teardown if the sender dies before the stop cutoff is flushed, and end_trace only clears trace_id/recording_id aftertrace_end is confirmed sent.

neuracore/data_daemon/communications_management/producer/producer_channel.py
StevenJacobs61: Addressed previously

neuracore/data_daemon/lifecycle/runtime_recovery.py
StevenJacobs61: That's my favourite empty time

neuracore/data_daemon/communications_management/consumer/spool_worker.py
StevenJacobs61: Added

neuracore/data_daemon/communications_management/consumer/trace_lifecycle_coordinator.py
StevenJacobs61: added

neuracore/data_daemon/recording_encoding_disk_manager/core/rgb_trace_spool.py
StevenJacobs61: Yes was an older interface i have cleaned it up

neuracore/data_daemon/communications_management/producer/producer_channel.py
StevenJacobs61: Yes i have cleaned it up to avoid repeating the conditional logic

neuracore/data_daemon/communications_management/producer/producer_transport_debug_models.py
StevenJacobs61: I have removed

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
StevenJacobs61: Yes i agree with this, the only concern once moving this is for the ram to not blow up if we can read from the slot faster than disk write, so i have implemented a bounded queue with memory checks to ensure this doesn't happen, but have implemented this refactor. This did speed up RGB logging speed significantly.

CougarTasker (commented):
> No this does not corrupt the data or cause an conflict as the flow continues to use the trace id send from the producer, so this is internal bookkeeping essentially regarding 'last seen trace id in this producer channel', and it propagates that trace ID. Technically we could keep a track of in flight traces for that producer if say we have some irregularity in the ordering or timing, for example we get trace A, then trace B then trace A again and then trace B, but we don't need to maintain this state for any behavioural correctness, so keeping this state consistent in terms of what trace we had ( not yet finished) vs what we are currently on would be more explicit, but maybe juts unnecessary overhead as there is not a risk of conflict between the two traces as we always use the one coming from the prod channel. We could make it consistent and explicit, but this might just be more buggy and ultimately perhaps pointless to manage as it doesn't affect behaviour at all. Let me know if you disagree.   

Thank you for removing this :)

CougarTasker (approved):
Looks good for now. While this pr brings some genunine improvments it introduces a lot of complexity not all of which I beleive is neccissary, in future prs we can refine this and improve documentation. lets make this the last giant pr

muneebneura (approved):
This pr serves as a good base for this architectural change. There are some refinement needed in the code which we will need to address with future prs.  Overall, great work @StevenJacobs61.

LGTM.

PR #598, opened 2026-05-05
feat: surface daemon premature exit errors

damonhayhurst-neuraco:
### Features

 - Premature exit of the daemon shows errors such as when auth errors occur


CougarTasker (approved):
LGTM 👍

PR #599, opened 2026-05-05
fix: call create dataset once per daemon test

damonhayhurst-neuraco:
### Bugfixes

 - Per daemon test, create dataset was getting called once per worker. Due to the shared http session introduced recently, workers were in a race condition to read from the keep alive socket

### DIFF
--- a/tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
+++ b/tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
@@ -643,13 +643,8 @@ def log_frames(
     return [marker_name]
 
 
-def _bind_worker_dataset(*, dataset_name: str, create_dataset: bool) -> None:
-    """Ensure a worker is bound to the shared dataset before recording."""
-    if create_dataset:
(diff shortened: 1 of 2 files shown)


tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
```diff
@@ -643,13 +643,8 @@ def log_frames(
...
 
-def _bind_worker_dataset(*, dataset_name: str, create_dataset: bool) -> None:
-    """Ensure a worker is bound to the shared dataset before recording."""
-    if create_dataset:
-        with Timer(MAX_TIME_TO_START_S, label="nc.create_dataset", always_log=True):
-            nc.create_dataset(dataset_name)
-        return
-
+def _bind_worker_dataset(*, dataset_name: str) -> None:
+    """Poll until the shared dataset is visible to this worker."""
```
muneebneura: can you clarify or rename what shared dataset means here, because it looks like we are referring to the shared org datasets.

muneebneura (approved):
LGTM!

PR #602, opened 2026-05-05
test: improved logging for test timer

damonhayhurst-neuraco:
### Features

 - Wall clock timer for overall test time
 - Print case id with prefix

### DIFF
--- a/tests/integration/platform/data_daemon/conftest.py
+++ b/tests/integration/platform/data_daemon/conftest.py
@@ -124,13 +124,13 @@ def log_run_analysis_on_teardown(
 
 
 def pytest_terminal_summary(terminalreporter, exitstatus, config):
-    """Print a session summary at the end of the test run."""
     del exitstatus, config
     if not SESSION_RUNS:
         return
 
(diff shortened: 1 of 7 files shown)


tests/integration/platform/data_daemon/conftest.py
```diff
@@ -126,38 +126,6 @@ def finalizer() -> None:
...
-    if infra_labels:
-        lines.append("\n  Infrastructure timings:")
-        for label in infra_labels:
-            lines.append(_format_timer_stats_line(label, Timer._stats[label]))
-
-    lines.append(separator)
-    terminalreporter.write_line("\n".join(lines))
+    summary = build_terminal_summary(SESSION_RUNS, Timer._stats)
+    if summary is not None:
+        terminalreporter.write_line(summary)
```
CougarTasker: Question: do we need a seperate method for this?
Note: there is nothing wrong with this code on its own but. un-neccsarry indirection makes diffs harder to compare

CougarTasker (approved):
LGTM 👍

PR #603, opened 2026-05-06
fix: add set -o pipefail on git workflow to avoid slient unit test failure

kwangneuraco:
By adding this line, the pipeline returns the first non-zero exit code in the chain instead of the last, this fixes the silent fail from the non-ml unit test


muneebneura (approved):
LGTM but don't merge it in as discussed offline

CougarTasker (approved):
LGTM 👍

PR #604, opened 2026-05-06
test: daemon data integrity test bug and improvements

damonhayhurst-neuraco:
### Features

 - Explicit test cases rather than programmatic
 - Context number hidden in id unless more than one
 - Separated concepts of deadlines for perf tests and scheduled log times for stochastic perf tests

### Bugfixes

 - Data integrity cases asserting deadlines when they shouldn't be
 - Fixed bug where stochastic timestamps are only creating jitter constant once per worker
 - Removed overly strict assertion for sequential mode where we checking the end times were within a tolerance (Thanks to @muneebneura for revealing this bug)

### DIFF
--- a/tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
+++ b/tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
@@ -23,6 +23,7 @@ from tests.integration.platform.data_daemon.shared.assertions import assert_cont
 from tests.integration.platform.data_daemon.shared.process_control import (
     MAX_TIME_TO_LOG_S,
     Timer,
+    assert_on_schedule,
 )
 from tests.integration.platform.data_daemon.shared.test_case.build_test_case import (
     DataDaemonTestCase,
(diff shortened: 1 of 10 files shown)


tests/integration/platform/data_daemon/shared/test_case/build_test_case_context.py
```diff
@@ -524,6 +530,7 @@ def worker(role_spec: dict[str, object]) -> None:
                             label="nc.log_joint_positions",
                             deadline=frame_deadline if assert_deadline else None,
                             timing_tolerance=timing_tolerance,
+                            assert_limit=assert_deadline,
```
CougarTasker: note: disscussed offline lets use one name for this to avoid confusion, agreeing on `assert_deadline`

CougarTasker (approved):
LGTM 👍

CougarTasker (approved):
Discussed offline: the assert on schedule may open up a new kind of failure @damonhayhurst-neuraco confirmed he wants to take on the fallout of introducing this.

PR #607, opened 2026-05-08
fix: update large model defaults for less memory usage

aditya-neuraco:
### Bugfixes

 - Updated the default data type from `float32` to `bfloat16` and enabled `gradient_checkpointing` by default.
 - Validations process went out of memory when using the defaults as `algorithm_config` wasn't getting passed to the script. So it defaulted to `float32` and `gradient_checkpointing` being `False` which spiked memory a lot and Pi0 didn't fit on A100 40GB.

### Items

 - [NCRL-447](https://neuracore.atlassian.net/browse/NCRL-447)


kwangneuraco (approved):
LGTM!

PR #609, opened 2026-05-11
fix: training config 

sdas-neuraco:
### Features
- Added support for composing user training YAML files via Hydra's optional `user` config group, so training arguments can live in reusable config files instead of long CLI commands.
- Introduced `algorithm_name` as packaged-algorithm shorthand, resolving it to the matching bundled algorithm config while preserving support for cloud `algorithm_id` workflows.
- Renamed local run naming config from `run_name` / `run_name_auto_increment` to `training_name` / `training_name_auto_increment`, including output directory resolution and documentation updates.
- Added training config normalization utilities in `neuracore/ml/utils/training_config.py` for default merging, algorithm resolution, dataset lookup, data type shorthand expansion, robot name/id conversion, and validation.
- Added `get_robot_name_from_id` to support converting resolved robot IDs back to user-facing robot names for saved training metadata.
- Moved all validation and completion checks outside of train.py

### DIFF
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -23,23 +23,14 @@ from neuracore_types import (
     CrossEmbodimentUnion,
     ModelInitDescription,
 )
-from neuracore_types.nc_data import DataType
 from omegaconf import DictConfig, OmegaConf
 from torch.utils.data import DataLoader, DistributedSampler, random_split
 
 import neuracore as nc
-from neuracore.api.training import _get_algorithms
 from neuracore.core.const import DEFAULT_CACHE_DIR, DEFAULT_RECORDING_CACHE_DIR
 from neuracore.core.utils.robot_data_spec_utils import (
-    convert_cross_embodiment_description_names_to_ids,
-    convert_omegaconf_to_cross_embodiment_description,
     extract_data_types,
     merge_cross_embodiment_description,
-)
-from neuracore.core.utils.training_input_args_validation import (
-    _get_data_types_for_algorithms,
-    get_algorithm_name,
-    validate_training_params,
 )
 from neuracore.ml import BatchedTrainingSamples, NeuracoreModel
 from neuracore.ml.datasets.pytorch_synchronized_dataset import (
@@ -65,6 +56,11 @@ from neuracore.ml.utils.preprocessing_utils import (
     PreprocessingConfiguration,
     resolve_preprocessing_config,
 )
+from neuracore.ml.utils.training_config import (
+    resolve_to_complete_config,
+    resolve_user_input_config,
+    validate_complete_config,
+)
 from neuracore.ml.utils.training_storage_handler import TrainingStorageHandler
 
 # Environment setup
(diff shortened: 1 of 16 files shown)


neuracore/ml/train.py
```diff
@@ -878,38 +743,27 @@ def _main(cfg: DictConfig) -> None:
...
-                input_cross_embodiment_description
+                cfg.input_cross_embodiment_description
             )
         )
         output_cross_embodiment_description = (
             convert_cross_embodiment_description_names_to_ids(
-                output_cross_embodiment_description
+                cfg.output_cross_embodiment_description
             )
         )
```
ypang-neuraco: convert_cross_embodiment_description_names_to_ids is already called in resolve_to_complete_config. Can we save that in the cfg returned?

neuracore/ml/train.py
```diff
@@ -878,38 +743,27 @@ def _main(cfg: DictConfig) -> None:
...
-            raise ValueError("Either 'dataset_id' or 'dataset_name' must be provided.")
+            dataset = nc.get_dataset(id=cfg.dataset_id)
+
+        cfg = resolve_to_complete_config(cfg, dataset=dataset)
+        logger.info("Training configuration:")
+        logger.info(OmegaConf.to_yaml(cfg, resolve=False))
+        logger.info(f"Training run directory: {cfg.local_output_dir}")
+
         dataset.cache_dir = _resolve_recording_cache_dir(cfg)
         dataset.cache_dir.mkdir(parents=True, exist_ok=True)
```
ypang-neuraco, line 764: Can we move this inside resolve_to_complete_config?

neuracore/ml/train.py
ypang-neuraco: discussed offline, please remove convert_omegaconf_to_cross_embodiment_description if it is not needed anymore

neuracore/ml/utils/training_config.py
```diff
@@ -0,0 +1,381 @@
...
+        cfg.dataset_name = _primitive_attr(dataset, "name", cfg.get("dataset_name"))
+
+    algorithms_jsons = _get_algorithms()
+    # Populate algorithm name/id from whichever selector the caller supplied.
+    if cfg.algorithm_id is not None:
+        algorithm_name = get_algorithm_name(
+            algorithm_id=cfg.algorithm_id,
+            algorithm_jsons=algorithms_jsons,
+        )
+        cfg.algorithm_name = algorithm_name
```
ypang-neuraco, line 343: There is the get_algorithm api available (although it prioritize shared algorithms)

neuracore/ml/utils/training_config.py
```diff
@@ -0,0 +1,381 @@
...
+        shared_alg_req = executor.submit(
+            get_session().get,
+            f"{API_URL}/org/{org_id}/algorithms",
+            headers=auth.get_headers(),
+            params={"shared": True},
+        )
+        org_alg, shared_alg = org_alg_req.result(), shared_alg_req.result()
+    org_alg.raise_for_status()
+    shared_alg.raise_for_status()
+    return org_alg.json() + shared_alg.json()
```
ypang-neuraco: Instead of making a concurrent call here, maybe we should add an option to the api to return both shared and org specific algos

neuracore/core/robot.py
```diff
@@ -928,3 +928,37 @@ def get_robot_id_from_name(robot_name: str, org_id: str | None = None) -> str:
...
+    ) + list_organization_robots(org_id, is_shared=True)
+
+    for robot in robots:
+        if robot["id"] == robot_id:
+            return robot["name"]
+
+    raise RobotError(
+        f"Robot with ID '{robot_id}' not found in organization '{org_id}' "
+        "or shared robots."
+    )
```
ypang-neuraco: We have an api get_robot_names for this

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/train.py
sarthakdas: this is just to get the config value as a variable to use directly rather then passing the entire config everywhere

neuracore/ml/utils/training_config.py
sdas-neuraco: Created a Ticket to adress the backend change later

PR #612, opened 2026-05-12
feat: integration test fail on error in logs

sdas-neuraco:
### Features

 - When we check when training is completed if any of the logs from the VM are error messages this will cause the integration tests to fail.

### DIFF
--- a/tests/integration/ml/test_training_flow.py
+++ b/tests/integration/ml/test_training_flow.py
@@ -100,6 +100,7 @@ CNNMLP_CONFIG = {
 }
 
 TERMINAL_STATES = {"COMPLETED", "FAILED", "CANCELLED", "ERROR"}
+TRAINING_ERROR_LOG_MAX_ENTRIES = 100
 
 # A batch_size value that is not "auto" and not parseable as an integer.
 # It passes client-side validation (which only checks data types / algorithm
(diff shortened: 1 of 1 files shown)


tests/integration/ml/test_training_flow.py
```diff
@@ -306,6 +306,44 @@ def _wait_for_training(
...
+    return "\n\n".join(formatted_entries)
+
+
+def _assert_no_training_log_errors(job_id: str, context: str) -> None:
+    """Fail if a training VM emitted ERROR or CRITICAL log entries."""
+    offending_entries: list[dict] = []
+    for severity in ("ERROR", "CRITICAL"):
+        logs = nc.get_training_job_logs(
+            job_id,
+            max_entries=100,
```
mark-neuracore: a bit unfamiliar with the nature of the logs, why only 100 entries?

tests/integration/ml/test_training_flow.py
```diff
@@ -574,6 +612,10 @@ def test_training_flow():
                     job_id, max_entries=10, severity_filter="ERROR"
                 )
                 assert "logs" in filtered
+                assert not filtered["logs"], (
+                    "Training VM emitted ERROR logs while RUNNING:\n\n"
+                    f"{_format_training_log_entries(filtered['logs'])}"
+                )
```
mark-neuracore: i feel there is a bit of repetition here since we introduced this function `_assert_no_training_log_errors` and it seemed like it does something similar yet not being used which caused some confusion to me. if it cant be used please leave a comment why, and if can maybe it's better to refactor to use it

sdas-neuraco: agree, this was mainly to chek only errors in log for max_entries=10  while the other uses critial and error, with better formatting

mark-neuracore (approved):
LGTM w/c 

left my comments earlier

aditya-neuraco (changes requested):
The structure of the training flow test has changed. It's now modular and more verbose. Check: https://github.com/NeuracoreAI/neuracore/commit/312c1db7f25bc8ab2a26fd2bd37fec21d42e6fe6 

Please make these changes in the newer version.

aditya-neuraco (approved):
LGTM!

PR #617, opened 2026-05-13
fix: importer overwrite should wait for dataset to complete deletion

ypang-neuraco:
### Bugfixes
- Importer has overwrite option to delete existing dataset with the same name and create a new one
- Dataset deletion request is sent but does not wait for deletion to complete before creating new one causing error on dataset creation
- This fix tries to create the new dataset at a fixed time interval until it is successfully created

### DIFF
--- a/neuracore/importer/importer.py
+++ b/neuracore/importer/importer.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import os
 import sys
+import time
 from pathlib import Path
 from types import SimpleNamespace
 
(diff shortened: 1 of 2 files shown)


tests/unit/importer/test_importer.py
aditya-neuraco, line 1: I think it would be more appropriate to name this file `test_importer.py` since we might add mode functions later on.

neuracore/importer/importer.py
```diff
@@ -376,6 +385,59 @@ def _run_import(
     logger.info("Finished importing dataset.")
 
 
+def _create_dataset_with_overwrite_guard(
```
aditya-neuraco, line 388: I feel this change should be part of the `create_dataset` function.

aditya-neuraco (approved):
LGTM! w/ comments

neuracore/importer/importer.py
ypang-neuraco: I agree but with the timeout I think the change will need some planning so not including in this PR

PR #618, opened 2026-05-13
feat: importer option to log target ee pose

ypang-neuraco:
### Features
- Debug option to use forward kinematics to convert joint target positions to EE poses with the specified frame name and log them


sdas-neuraco (approved):
LGTM

PR #619, opened 2026-05-13
fix: correct resolution of preprocessing omegaconf

aditya-neuraco:
### Bugfixes
- Fix `TypeError: Object of type ListConfig is not JSON serializable` during cloud training runs. Hydra's `instantiate` now called with `_convert_="all"` in `resolve_preprocessing_config`, ensuring OmegaConf containers (e.g. `ListConfig` for `ResizePad.size`) are converted to plain Python types before use.

### Items
- [NCRL-567](https://neuracore.atlassian.net/browse/NCRL-567)


mark-neuracore (approved):
LGTM

PR #620, opened 2026-05-14
feat: reduce the pr unit test time 

kwangneuraco:
reduce the pr unit test time by moving the time-consuming test_run_validation test for large models such as pi0, pi 05, groot etc from every pr commit into test when merge.

### DIFF
--- a/.github/workflows/pr-merge-test.yaml
+++ b/.github/workflows/pr-merge-test.yaml
@@ -1,0 +1,46 @@
+name: PR Merge Test
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'neuracore/ml/**'
+      - 'neuracore/core/**'
+      - 'tests/unit/ml/**'
+      - 'tests/unit/conftest.py'
+      - 'neuracore/__init__.py'
+      - 'setup.py'
+      - '.github/workflows/pr-merge-test.yaml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  test-ml-slow:
+    runs-on: ubuntu-latest
+    env:
+      UV_VENV_SEED: "true"
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: ["3.10", "3.11"]
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-merge-test.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - 'setup.py'
+      - '.github/workflows/pr-merge-test.yaml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  test-ml-slow:
+    runs-on: ubuntu-22.04
```
aditya-neuraco: Use `ubuntu-latest`

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -91,9 +88,7 @@ jobs:
 
       - name: Run non-ML unit tests
         if: env.SHOULD_RUN == 'true'
-        shell: bash
         run: |
-          set -euo pipefail
```
aditya-neuraco, line 96: We need this line so that intermediate exit codes fail CI tests.

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -156,4 +151,4 @@ jobs:
         if: env.SHOULD_RUN == 'true'
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
-        run: uv run pytest tests/unit/ml
\ No newline at end of file
+        run: uv run pytest tests/unit/ml -m "not slow"
```
aditya-neuraco, line 158: we can use `-k "not test_run_validation"` as well.

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -4,9 +4,6 @@ on:
   pull_request:
     branches:
       - main
-  push:
-    branches:
-      - main
```
aditya-neuraco, line 9: This will also stop non-ml tests from running on push to main, which we don't want.

aditya-neuraco (approved):
LGTM w/ comments.

PR #622, opened 2026-05-15
feat: set-up unit tests to run with different environments

muneebneura:
### Features

 - Non ML unit tests setup to run with different environments
 - The importer tests are skipped on macOS for now as pinnochio has no pre-built wheels on pypi and needs to be installed from source. As dicussed with @aditya-neuraco, the RL team will be looking into using a cross platform library and we can than enable running importer tests on mac.

Non ml tests will run on:

1. ubuntu-22.04	3.10	uv
2. ubuntu-22.04	3.10	pip
3. ubuntu-22.04	3.11	uv
4. ubuntu-22.04	3.11	pip
5. macos-latest	3.10	uv
6. macos-latest	3.10	pip
7. macos-latest	3.11	uv
8. macos-latest	3.11	pip

**Note: We will need to change the branch protection rule for main and replace the old two non ml checks with test-non-ml-complete which checks for all jobs in the new matrix.**

### Fixes
- Moved test_server.py to ml tests

### Items

 - [NCP-946](https://neuracore.atlassian.net/browse/NCP-946)


.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -56,24 +56,32 @@ jobs:
 
   test-non-ml:
     needs: [detect-changes, pre-commit]
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
+        os: [ubuntu-22.04, macos-latest]
```
aditya-neuraco: We can use ubuntu-latest instead of ubuntu-22.04

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -56,24 +56,32 @@ jobs:
...
         python-version: ["3.10", "3.11"]
+        package-manager: [uv, pip]
```
aditya-neuraco, line 65: Not testing with conda?

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -83,26 +91,48 @@ jobs:
...
-            uv pip install pytest-cov diff-cover
+          if [ "${{ matrix.os }}" = "macos-latest" ]; then
+            EXTRAS=".[dev]"
+          else
+            EXTRAS=".[dev,import]"
+          fi
+          if [ "${{ matrix.package-manager }}" = "uv" ]; then
+            uv pip install "${EXTRAS}" --extra-index-url https://download.pytorch.org/whl/cpu --index-strategy unsafe-best-match
+            if [ "${SHOULD_RUN_COVERAGE}" = "true" ]; then
+              uv pip install pytest-cov diff-cover
```
aditya-neuraco: We should make these part of dev subgroup?

.github/workflows/pr-pre-commit-and-test.yaml
```diff
@@ -124,6 +154,17 @@ jobs:
...
+  test-non-ml-complete:
+    needs: [test-non-ml]
+    runs-on: ubuntu-22.04
+    if: always()
+    steps:
+      - name: Check all matrix jobs passed
+        run: |
+          if [ "${{ needs.test-non-ml.result }}" != "success" ]; then
+            exit 1
+          fi
```
aditya-neuraco, line 162: Why is this needed? if we are always running this, no need to have an if condition.

.github/workflows/pr-pre-commit-and-test.yaml
aditya-neuraco, line 1: Can you also add these options to pytest `-sv -o log_cli=true --log-cli-level=INFO`? 

It will help us better understand what went wrong. This prints `logging` module and `print()` outputs.

aditya-neuraco (approved):
LGTM w/ comments.

.github/workflows/pr-pre-commit-and-test.yaml
CougarTasker: Not yet ... let this version land and we can have a look at this.

CougarTasker (approved):
LGTM \w @aditya-neuraco commments

.github/workflows/pr-pre-commit-and-test.yaml
muneebneura: Okay good idea!

.github/workflows/pr-pre-commit-and-test.yaml
muneebneura: Without the exit 1 check, test-non-ml-complete would always report success to branch protection, even when matrix jobs failed.

.github/workflows/pr-pre-commit-and-test.yaml
aditya-neuraco: Fair enough.

aditya-neuraco:
@muneebneura We should also add windows tests.

muneebneura:
> @muneebneura We should also add windows tests.

We are not targeting windows, atleast for now as discussed with @CougarTasker. Windows users can be directed to use wsl

PR #626, opened 2026-05-18
ci: use cpu torch tf in non ml tests

aditya-neuraco:
### Bugfixes

 - We don't use any GPU operations in the importer, so removing tensorflow gpu dependency.
 - Also added torch CPU index url to download CPU torch in importer as CI doesn't have a GPU.

### Items


muneebneura (approved):
LGTM!

PR #627, opened 2026-05-18
fix: resolve macOS file name too long errors in data daemon

muneebneura:
macOS enforces stricter limits than Linux: shared memory names max out at 31 chars and socket paths at 104 bytes. Both were exceeded by the daemon's naming scheme and the test socket paths.

### Bugfixes

- Renamed shm for what is safe for all platforms.
 - Fixed maing issues with tests for macOS (we will soon have unit tests running on diff envs)

### Items

 - [NCP-962](https://neuracore.atlassian.net/browse/NCP-962)

### DIFF
--- a/neuracore/data_daemon/communications_management/shared_transport/shared_memory_budget.py
+++ b/neuracore/data_daemon/communications_management/shared_transport/shared_memory_budget.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass, field
 from neuracore.data_daemon.communications_management.shared_transport.models import (
     SharedSlotReservation,
 )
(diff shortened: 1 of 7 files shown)


neuracore/data_daemon/communications_management/shared_transport/shared_memory_budget.py
```diff
@@ -42,11 +43,11 @@ class SharedMemoryBudget:
 
     def __init__(
         self,
-        shm_path: str = "/dev/shm",
+        shm_path: str = "",
```
StevenJacobs61:
```suggestion
        shm_path: str | None = None,
```

StevenJacobs61 (approved):
LGTM w/comment

PR #628, opened 2026-05-18
fix: shared-slot-transport-teardown

StevenJacobs61:
### Features

 - Wait for slot to drain for valid slots before logging and teardown.
 - Log debug warning for  shared slot acks for torn down slot memory

### Bugfixes

 - Producer was failing if receiving acks for slot which has already been torn down when starting and stopping recording in quick succession .


neuracore/data_daemon/communications_management/consumer/spool_worker.py
```diff
@@ -104,18 +109,49 @@ def _worker_loop(self) -> None:
...
+                        work.descriptor_payload
+                    )
+                if descriptor is not None:
+                    self._shared_slot_handler.mark_descriptor_completed(
+                        work.channel.producer_id,
+                        descriptor,
+                    )
 
-        try:
             descriptor = transport_result.descriptor
```
muneebneura, line 168: This descriptor is already defined in the try block above now, remove this.

neuracore/data_daemon/communications_management/shared_transport/shared_slot_daemon_handler.py
```diff
@@ -208,6 +239,132 @@ def handle_descriptor(
...
+                    self._pending_condition.notify_all()
+                    logger.warning(
+                        "Abandoning stalled shared-slot session before reopen "
+                        "producer_id=%s shm_name=%s pending_sequences=%s "
+                        "timeout=%.3fs",
+                        channel.producer_id,
+                        shm_name,
+                        abandoned,
+                        self._reopen_drain_timeout_s,
+                    )
```
muneebneura: Always good to move logs outside the lock. You can break here and move this log outside the with block.

muneebneura (approved):
LGTM w/c

neuracore/data_daemon/lifecycle/runtime_recovery.py
```diff
@@ -247,25 +281,26 @@ async def reconcile_state_with_filesystem(
...
-    "cleanup_stale_shared_memory_buffers",
+    "cleanup_stale_runtime_state",
+    "cleanup_stale_shared_slot_control_sockets",
     "cleanup_stale_shared_slot_segments",
     "cleanup_socket_files",
     "ensure_shared_memory_capacity",
     "reconcile_state_with_filesystem",
     "SharedMemoryCapacityError",
+    "StaleRuntimeCleanupResult",
     "shared_memory_free_bytes",
```
muneebneura, line 294: See my comments above regarding this unused code

neuracore/data_daemon/lifecycle/runtime_recovery.py
```diff
@@ -33,6 +34,14 @@ class SharedMemoryCapacityError(RuntimeError):
...
 
+@dataclass(frozen=True)
+class StaleRuntimeCleanupResult:
+    """Counts for stale daemon runtime state cleanup."""
+
+    shared_slot_segments: int = 0
+    shared_slot_control_sockets: int = 0
+
+
 def _format_bytes(value: int) -> str:
```
muneebneura: I couldn't find this being used anywhere, maybe you intended to use this somewhere?

neuracore/data_daemon/lifecycle/runtime_recovery.py
```diff
@@ -138,6 +163,15 @@ def cleanup_socket_files(paths: Iterable[Path]) -> None:
...
+def cleanup_stale_runtime_state(
+    socket_paths: Iterable[Path],
+) -> None:
+    """Clean stale daemon-owned runtime files from previous processes."""
+    cleanup_stale_shared_slot_segments()
+    cleanup_stale_shared_slot_control_sockets()
+    cleanup_socket_files(socket_paths)
+
+
 def validate_or_recover_sqlite(db_path: Path, *, recover: bool = True) -> bool:
```
muneebneura, line 166: I think this is why you defined the data class above?

```suggestion
return StaleRuntimeCleanupResult(
        shared_slot_segments=segments,
        shared_slot_control_sockets=sockets,
    )

```

muneebneura (commented):
LGTM w/c
Don't merge this in before we do the release.

neuracore/data_daemon/communications_management/producer/producer_channel.py
```diff
@@ -430,6 +439,103 @@ def _iter_chunk_views(
...
+        normalised_parts: Sequence[memoryview],
+        total_chunks: int,
+        trace_metadata: TraceTransportMetadata,
+    ) -> None:
+        """Send one logical payload over the shared-slot transport."""
+        self.open_fixed_shared_slots()
+        shared_slot_transport = self._shared_slot_transport
+        if shared_slot_transport is None:
+            raise RuntimeError("Shared-slot transport is not available")
+
```
muneebneura, line 503: If none does appear here, during the stop producer channels teardown why not raise this instead of a runtime error so a recovery should be attempted.
```suggestion
        shared_slot_transport = self._shared_slot_transport
        if shared_slot_transport is None:
                raise SharedSlotUnhealthyError("Shared-slot transport is not available")
```

PR #631, opened 2026-05-20
ci: fix pink import issues caused by cmeel-boost update

aditya-neuraco:
### Bugfixes

- **Fix CI importer test failures caused by `cmeel-boost 1.90.0` release**

On 2026-05-20, `cmeel-boost 1.90.0` was published to PyPI. This broke the dependency resolution for the `[import]` extras in a non-obvious way:

- `pin-pink==4.2.0` requires `pin>=2.6.3` (no upper bound on major version)
- `pin 3.9.0` (latest 3.x) requires `cmeel-boost~=1.89.0`, which excludes `1.90.0`
- `uv`'s resolver prefers maximising overall version freshness across all packages
- With `1.90.0` as the newest `cmeel-boost`, `uv` backtracks from `pin 3.x` entirely
- It instead resolves:
  - `pin 2.6.18`
  - `hpp-fcl 2.3.4` (no `cmeel-boost` constraint)
  - `cmeel-boost 1.90.0`
- `pin 2.6.18` was compiled against Boost `1.82`, but `cmeel-boost 1.90.0` provides `libboost_python311.so.1.90.0`
- This creates an ABI mismatch, resulting in an `ImportError` at runtime

Before 2026-05-20, `cmeel-boost 1.89.0` was the latest version. `pin 3.x` (`~=1.89.0`) resolved correctly, and its wheels already included the required RPATH pointing to `cmeel.prefix/lib`, so no linker configuration was needed.

The failure was first exposed because this PR modifies `tests/unit/conftest.py`, which matches the `non-ml` path filter, causing `SHOULD_RUN=true` and triggering the importer tests on CI.

#### Fix

Explicitly pin the following packages in `setup.py`'s `[import]` extras:

- `pin==3.9.0`
- `pin-pink==4.2.0`

This forces `uv` to resolve `pin 3.9.0`, which constrains `cmeel-boost~=1.89.0` and therefore resolves to `1.89.0`, the version that `pin 3.9.0` was compiled against.

### Items
   - [NCRL-577](https://neuracore.atlassian.net/browse/NCRL-577)


mark-neuracore (approved):
LGTM

PR #635, opened 2026-05-21
fix: pin dependency versions

aditya-neuraco:
### Bugfixes
- Fixes `ImportError` when importing neuracore after `pip install neuracore[import]` on a clean install**

`pin 3.9.0` (pinocchio) declares loose constraints on its cmeel native dependencies (e.g. `cmeel-urdfdom>=4`, `cmeel-tinyxml2` unbounded) but was compiled against specific soname versions. When newer cmeel packages were published — `cmeel-urdfdom 6.0.0` (soname 6) and `cmeel-tinyxml2 11.0.0` (soname 11) — uv resolved them, causing runtime `ImportError: cannot open shared object file` failures because the compiled pinocchio extension searched for `liburdfdom_sensor.so.4.0` and `libtinyxml2.so.10`, which no longer existed in the installed packages.

Similarly, `pin-pink>=4` without a lower bound on the pinocchio major version caused uv to resolve `pin 2.6.18` (compiled against boost 1.82) alongside `cmeel-boost 1.90.0` (released 2026-05-20), causing `libboost_python311.so.1.82.0: cannot open shared object file`.

**Fix:** Pin the full native ABI surface in `[import]` extras to the exact versions that `pin 3.9.0` was compiled against, verified by installing in a clean venv and confirming `from neuracore.importer.core.robot_utils import RobotUtils` succeeds without any system-level linker configuration.

### Items
- [NCRL-577](https://neuracore.atlassian.net/browse/NCRL-577)


sdas-neuraco (approved):
LGTM

PR #638, opened 2026-05-21
fix: ensure db cleanup

StevenJacobs61:
### Bugfixes

 - The test was clearing the .wal and .shm files with the wrong prefix


damonhayhurst-neuraco (approved):
LGTM

PR #640, opened 2026-05-22
feat: Preflight daemon auth before socket readiness polling

favour-neuraco:
### Features
- Added authentication preflight before daemon subprocess startup, so interactive login completes before daemon socket readiness polling begins.

### Items
 - [NC-974](https://neuracore.atlassian.net/browse/NCP-974)

### DIFF
--- a/neuracore/data_daemon/lifecycle/auth_preflight.py
+++ b/neuracore/data_daemon/lifecycle/auth_preflight.py
@@ -1,0 +1,28 @@
+"""Authentication preflight for daemon startup."""
+
+import os
+
+from neuracore.core.auth import login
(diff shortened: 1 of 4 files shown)


neuracore/data_daemon/lifecycle/auth_preflight.py
```diff
@@ -0,0 +1,8 @@
+"""Authentication preflight for daemon startup."""
+
+from neuracore.core.auth import login
+
+
+def ensure_daemon_auth_ready() -> None:
```
CougarTasker: blocker: we need to account for offline mode

CougarTasker (approved):
LGTM with comment

PR #641, opened 2026-05-22
fix: notify daemon when recording is stopped remotely

muneebneura:
When a recording is stopped from the web frontend, we were never draining data streams or notifying the data daemon, so recordings were never finalized or uploaded. The bug was introduced when the data daemon was updated to use a shared ring buffer transport that requires an explicit stop IPC with sequence number cutoffs in this pr: [NCP-695](https://github.com/NeuracoreAI/neuracore/pull/595)

### Bugfixes

 - Fixed by registering a remote stop handler on connect that drains streams and notifies the daemon the same way the local stop path does, but is only triggered by the SSE stop event.
 - Added a test for this remote path which checks if drain happend after the sse

### Items

 - [NCP-970](https://neuracore.atlassian.net/browse/NCP-970)

### DIFF
--- a/neuracore/core/robot.py
+++ b/neuracore/core/robot.py
@@ -6,6 +6,7 @@ URDF/MJCF model uploads, data stream management, and coordinates recording
 state across multiple robot instances.
 """
 
+import functools
 import io
 import logging
 import os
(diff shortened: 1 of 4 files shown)


neuracore/core/robot.py
```diff
@@ -333,17 +334,12 @@ def stop_recording(
         if not self.id:
             raise RobotError("Robot not initialized. Call init() first.")
 
+        self._drain_streams_and_notify_daemon(
+            recording_id, wait_for_producer_drain=wait_for_producer_drain
+        )
```
StevenJacobs61: Lets keep this below the .recording_stopped function

StevenJacobs61 (approved):
LGTM W/comments. Lest see f we can remove those RECORDING_STOPPED / RECORDING_STARTED / RECORDING_SAVED events emitting also

neuracore/core/robot.py
muneebneura: Done

PR #642, opened 2026-05-26
fix: create default profile when checking daemon auth

favour-neuraco:
create default profile when checking daemon auth


muneebneura (approved):
LGTM

PR #644, opened 2026-05-26
fix: update robot list request

ypang-neuraco:
### Bugfixes
- Get robot id from name should use the new list method since the old one is now deprecated in (https://github.com/NeuracoreAI/neuracore_backend/pull/379)

### Items
 - [NCRL-593](https://neuracore.atlassian.net/browse/NCRL-593)

### Related PRs
- https://github.com/NeuracoreAI/neuracore_backend/pull/379


favour-neuraco (approved):
LGTM. Since list_organization_robots is now paginated, could we add one direct test covering multiple pages?

PR #645, opened 2026-05-26
fix: select org for cli training run inspections

sdas-neuraco:
### Features

 - Bug notices by Mark/Halid/Ke for when they selected an org with select org and then viewed their traning runs it used the primary org, added extra param to handle requests for different organisations.

### DIFF
--- a/neuracore/api/training.py
+++ b/neuracore/api/training.py
@@ -336,11 +336,12 @@ def get_training_job_logs(
         raise ValueError(f"Error getting training job logs: {e}")
 
 
-def delete_training_job(job_id: str) -> None:
+def delete_training_job(job_id: str, org_id: str | None = None) -> None:
     """Delete a training job and free its resources.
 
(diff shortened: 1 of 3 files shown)


ypang-neuraco (commented):
I'm not sure this is the right fix. Maybe when you select org the environment variable should also be set?

neuracore/core/config/get_current_org.py
```diff
@@ -13,6 +13,14 @@
...
 
 
+def get_selected_org() -> str:
+    """Get the persisted organization selected by `neuracore select-org`."""
+    org_id = get_config_manager().config.current_org_id
+    if org_id:
+        return org_id
+    return get_current_org()
+
+
```
aditya-neuraco: We have get_current_org already. Please update the other references too.

aditya-neuraco (approved):
LGTM!

PR #648, opened 2026-05-26
fix: shared slot unhealthy fix

StevenJacobs61:
### Features

 - Moved the boundary for agreeing that shared slot memory is healthy to use.
 - Printed error message from daemon to the user.

### Bugfixes

 - We were getting shared slot memory unhealthy without any useful error message propagated to the user. 
 - We were trying to open new slot memory before we had processed existing valid data

### DIFF
--- a/tests/integration/ml/test_training_flow.py
+++ b/tests/integration/ml/test_training_flow.py
@@ -26,6 +26,8 @@ from neuracore_types import (
 import neuracore as nc
 from neuracore.core.data.dataset import Dataset
 from neuracore.core.endpoint import Policy
+from neuracore.core.robot import get_robot
+from neuracore.core.utils import backend_utils
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.join(THIS_DIR, "..", "..", "..", "examples"))
@@ -123,6 +125,8 @@ BACK_TO_BACK_CNNMLP_CONFIG = {
     "epochs": 1,
     "output_prediction_horizon": 5,
 }
+RECORDING_STOP_TIMEOUT_SECONDS = 500
+RECORDING_STOP_POLL_SECONDS = 0.5
 
 
 # ---------------------------------------------------------------------------
@@ -188,6 +192,7 @@ def _collect_demo_data(
             for _ in range(episode_length_multiplier)
         ]
         nc.start_recording(robot_name=robot_name, instance=instance_id)
+        recording_id = get_robot(robot_name, instance_id).get_current_recording_id()
         t = time.time()
         for frame_idx, action_dict in enumerate(expanded_action_traj):
             t += 1.0 / FREQUENCY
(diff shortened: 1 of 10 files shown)


tests/integration/ml/test_training_flow.py
```diff
@@ -209,13 +214,11 @@ def _collect_demo_data(
             pose = np.array([0.1 + frame_idx * 0.001, 0.2, 0.3, 0.0, 0.0, 0.0, 1.0])
             img = np.zeros((84, 84, 3), dtype=np.uint8)
             img.fill(50 + frame_idx % 200)
-            depth = np.full((64, 64), 0.75 + 0.01 * (frame_idx % 10), dtype=np.float32)
+            np.full((64, 64), 0.75 + 0.01 * (frame_idx % 10), dtype=np.float32)
```
muneebneura: We dont need to do this now as you have removed depth logging below.

tests/integration/ml/test_training_flow.py
```diff
@@ -209,13 +214,11 @@ def _collect_demo_data(
...
             point_cloud = np.linspace(0.0, 1.0, 96, dtype=np.float16).reshape(
                 32, 3
             ) + np.float16(frame_idx * 0.001)
-            rgb_points = np.full(
-                (point_cloud.shape[0], 3), 80 + frame_idx % 120, dtype=np.uint8
-            )
+            np.full((point_cloud.shape[0], 3), 80 + frame_idx % 120, dtype=np.uint8)
```
muneebneura: See my comment above, point cloud log is also removed

tests/integration/ml/test_training_flow.py
```diff
@@ -267,28 +270,51 @@ def _collect_demo_data(
...
-                instance=instance_id,
-            )
-            nc.log_point_cloud(
-                name=POINT_CLOUD_SENSOR_NAME,
-                points=point_cloud,
-                rgb_points=rgb_points,
-                timestamp=t,
-                robot_name=robot_name,
-                instance=instance_id,
-            )
```
muneebneura, line 284: What was the reason for removing these. Also worth having a chat with @aditya-neuraco about these as they were added in his pr: [#573]

tests/integration/ml/test_training_flow.py
```diff
@@ -267,28 +270,51 @@ def _collect_demo_data(
...
+    deadline = time.time() + timeout_seconds
+    saw_active_traces = False
+    last_trace_count: int | None = None
+    while time.time() < deadline:
+        data_traces = backend_utils.get_active_data_traces(recording_id)
+        last_trace_count = len(data_traces)
+        if last_trace_count > 0:
+            saw_active_traces = True
+        elif saw_active_traces:
+            return
```
muneebneura, line 306: If traces drain here before the first poll, saw_active_traces is never set and we'll spin until timeout even though recording completed successfully.

muneebneura (approved):
LGTM w/comments

tests/integration/ml/test_training_flow.py
StevenJacobs61: Good spot

tests/integration/ml/test_training_flow.py
StevenJacobs61: Yes have spoken with him, they are not necessary here and they are too much for the github shared memory to handle in CI

tests/integration/ml/test_training_flow.py
StevenJacobs61: Good spot

tests/integration/ml/test_training_flow.py
StevenJacobs61: I have changed the polling to be smaller but this is a genuine race condition that needs to be addressed for a more focused test improvement pr

PR #651, opened 2026-05-27
fix: training run name resolution

sdas-neuraco:
### Features

 - Refactored the way we handle the folder creation and checks, the bug occured due to a clash with Hydra creating the folder. 

This change uses a more Hydra Native Approach.

### DIFF
--- a/neuracore/ml/config/config.yaml
+++ b/neuracore/ml/config/config.yaml
@@ -75,8 +75,10 @@ preprocessing:
       - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
         size: [224, 224]
 
-# Output directory (generated dynamically with the resolver)
-local_output_dir: ${resolve_output_dir:${training_name},${training_name_auto_increment}}
+# Resolve the name separately from the full path so Hydra can still own
+# hydra.run.dir creation while Neuracore controls duplicate-name behavior.
+_resolved_training_name: ${resolve_training_run_name:${training_name},${training_name_auto_increment}}
+local_output_dir: ${oc.env:HOME}/.neuracore/training/runs/${_resolved_training_name}
 
 # Hydra configuration
 hydra:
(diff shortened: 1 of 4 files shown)


neuracore/ml/config/config.yaml
```diff
@@ -75,8 +75,10 @@ preprocessing:
       - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
         size: [224, 224]
 
-# Output directory (generated dynamically with the resolver)
-local_output_dir: ${resolve_output_dir:${training_name},${training_name_auto_increment}}
+# Resolve the name separately from the full path so Hydra can still own
+# hydra.run.dir creation while Neuracore controls duplicate-name behavior.
+resolved_training_name: ${resolve_training_run_name:${training_name},${training_name_auto_increment}}
```
mark-neuracore: why would we have a configuration parameter called resolved training name?  maybe add a _ before?
all parameters here are for users to configure and this sounds a bit confusing to me.

sdas-neuraco: Agree, it was ment to help hyra do solve the path resolving, adding _ will make it clearer

sdas-neuraco: These params need to be here because now Hydra handles the file creation and folder creation, we do not do that any more.

neuracore/ml/utils/training_config.py
```diff
@@ -23,8 +25,100 @@
...
+    # Find the first available suffix using the conventional ``name_N`` format.
+    suffix = 1
+    while (runs_dir / f"{run_name}_{suffix}").exists():
+        suffix += 1
+    resolved_name = f"{run_name}_{suffix}"
+    _RESOLVED_TRAINING_RUN_NAMES[cache_key] = resolved_name
+    return resolved_name
+
+
+OmegaConf.register_new_resolver(
```
mark-neuracore, line 115: question: why is this called in the script body here not in train.py or inside a method. isn't this a util file?

mark-neuracore (approved):
LGTM w/c

neuracore/ml/utils/training_config.py
sdas-neuraco: it is a util function, whcih is why its in the utils folder

PR #653, opened 2026-05-27
fix: pytorch train sample retry removal

sdas-neuraco:
### Features

 -  Removed the catch and retry for loading samples as this did not catch any expected errors, and retrying would not help any known possible failure cases.

### DIFF
--- a/neuracore/ml/datasets/pytorch_synchronized_dataset.py
+++ b/neuracore/ml/datasets/pytorch_synchronized_dataset.py
@@ -158,8 +158,6 @@ class PytorchSynchronizedDataset(PytorchNeuracoreDataset):
             self.synchronized_dataset_statistics.dataset_statistics
         )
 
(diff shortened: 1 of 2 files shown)


neuracore/ml/config/config.yaml
ypang-neuraco, line 1: Please remove changes from https://github.com/NeuracoreAI/neuracore/pull/651

ypang-neuraco (approved):
LGTM \w comments
I think the change makes sense but please double check with @stepjam as this was added a long time ago

neuracore/ml/config/config.yaml
mark-neuracore: same comment. please make sure to drop the unrelated commits when rebasing on main before pushing

sdas-neuraco: checked with stephen its all okay to be removed was a old hack

PR #654, opened 2026-05-27
fix: dataset hash conditioned on recordings

sdas-neuraco:
### Features

 - Makes the has of a dataset different if the dataset has any changes to the recordings to avoid it pulling dataset statistics if the recording has changed

this should stop the issue people where seeing where the index is out of range, when trying to train and get a tensor.


ypang-neuraco (approved):
LGTM

PR #656, opened 2026-05-29
test: update training flow integration test

ypang-neuraco:
### Features
- Use fixed batch size for training flow test (auto batchsize to be tested in a separate integration test this sprint)
- Add noise to timestep to simulate real life data recording
- Use "NYU ROT" instead of "ASU Table Top" as shared dataset to reduce test duration
- If any step fails, skip cleanup and retain training job and endpoint so logs can be viewed

### Bugfixes
- Use daemon setup and teardown for data recording
- Use target joint position for output instead of joint position
- Restore logging for target position and target gripper amount

### DIFF
--- a/tests/integration/ml/test_training_flow.py
+++ b/tests/integration/ml/test_training_flow.py
@@ -19,6 +19,7 @@ from neuracore_types import (
     EmbodimentDescription,
     JointData,
     LanguageData,
+    ParallelGripperOpenAmountData,
     RGBCameraData,
     SynchronizedPoint,
 )
@@ -26,8 +27,6 @@ from neuracore_types import (
 import neuracore as nc
 from neuracore.core.data.dataset import Dataset
 from neuracore.core.endpoint import Policy
-from neuracore.core.robot import get_robot
-from neuracore.core.utils import backend_utils
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.join(THIS_DIR, "..", "..", "..", "examples"))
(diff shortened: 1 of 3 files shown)


tests/integration/ml/test_training_flow.py
```diff
@@ -454,6 +443,10 @@ def _build_cross_embodiment_descriptions(
                 data_type in embodiment
             ), f"{data_type.value} missing from robot {robot_id} embodiment description"
         input_desc[robot_id] = {dt: embodiment[dt] for dt in input_types}
+        for data_type in output_types:
+            assert (
+                data_type in embodiment
```
aditya-neuraco: Can you rename `embodiment` to `embodiment_description` for clarity?

tests/integration/ml/test_training_flow.py
```diff
@@ -513,6 +511,7 @@ class TestTrainingFlow:
 
     @classmethod
     def setup_class(cls) -> None:
+        cls._all_steps_passed = True
```
aditya-neuraco: Can you remove the _ to conform with other variables?

tests/integration/ml/test_training_flow.py
```diff
@@ -524,6 +523,11 @@ def setup_class(cls) -> None:
 
     @classmethod
     def teardown_class(cls) -> None:
+        if not cls._all_steps_passed:
+            logger.warning(
+                "Skipping TestTrainingFlow teardown cleanup: one or more steps failed"
+            )
+            return
```
aditya-neuraco: Not sure tbh, it will keep the endpoints running unnecessarily.

aditya-neuraco (approved):
LGTM w/ comments!

tests/integration/ml/test_training_flow.py
ypang-neuraco: The endpoint will die after TTL is reached (30 mins) but it will not be deleted so we can view the logs

PR #657, opened 2026-05-29
test: document daemon online frequency boundary

favour-neuraco:
### Bugfixes
- Updated the online data-daemon performance test to reflect the validated VM frequency boundary for a fixed joint-only workload.
- Kept dataset readiness and expected recording-count checks strict, while treating per-call timing deadlines as diagnostics for performance runs.

### Items
- [NC-942](https://neuracore.atlassian.net/browse/NCP-942)


CougarTasker (approved):
LGTM 👍

PR #658, opened 2026-05-29
fix: reset daemon DB files during test cleanup

favour-neuraco:
### Bugfixes
 - Hardened behavioural correctness test cleanup by resetting daemon file-backed state directly for `STORAGE_STATE_EMPTY`.
 - The previous cleanup path opened `state.db`, deleted rows from `traces` / `recordings`, and left the DB file behind. This could fail during teardown with transient SQLite `disk I/O error` when the post-test assertion immediately reopened the DB.

### Items
 - [NCP-1020](https://neuracore.atlassian.net/browse/NCP-1020)


damonhayhurst-neuraco (approved):
LGTM

PR #610, opened 2026-05-11
fix: move from shared session singleton to create new

damonhayhurst-neuraco:
### Features

 - New session per request
 - Keep alive false

### Bugfixes

 - Fixes multithreading requests difficulties

PR #659, opened 2026-05-29
fix: share an http session per thread instead of constructing per call

CougarTasker:
### Features

 - Cold sessions were causing interation tests to fail performance requirments, 
     - previously we introduced the global session to address this but due to it being not thread safe it ended up causing subtle issues that were difficult tyo debug
     - the indivual sessions did not have the concurrency issues but were slower than the global session
     - this approach is the best of both worlds by using `threading.local()` to have a safe shared per thread session

### Bugfixes

 - None

### Items

 - [NCP-984](https://neuracore.atlassian.net/browse/NCP-984)

### Related PRs

 - https://github.com/NeuracoreAI/neuracore/pull/610
 - https://github.com/NeuracoreAI/neuracore/pull/590

### DIFF
--- a/neuracore/api/datasets.py
+++ b/neuracore/api/datasets.py
@@ -12,7 +12,7 @@ from neuracore.core.config.get_current_org import get_current_org
 from neuracore.core.const import API_URL
 from neuracore.core.data.dataset import Dataset
 from neuracore.core.exceptions import DatasetError
-from neuracore.core.utils.http_session import Session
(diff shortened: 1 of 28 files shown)


damonhayhurst-neuraco (approved):
LGTM

damonhayhurst-neuraco:
I like the context manager approach more from a readability standpoint but up to you.

PR #660, opened 2026-05-31
fix: pad input in policy inference

ypang-neuraco:
### Bugfixes
- For cross-embodiment training with 2 robots of different number of joints, the inference fails when using the robot with less joint because the input is not padded
- This fix pads the input data to the max items trained on similar to what we do in training
- Improve handling of padding in training
### Items
 - [NCRL-599](https://neuracore.atlassian.net/browse/NCRL-599)

### DIFF
--- a/neuracore/ml/utils/policy_inference.py
+++ b/neuracore/ml/utils/policy_inference.py
@@ -142,34 +142,45 @@ class PolicyInference:
         # We need to go from sync_point (single time step) to BatchedNCData
         for data_type in sync_point.data.keys():
             inputs[data_type] = []
-            max_items_for_this_data_type = len(sync_point.data[data_type])
+            embodiment_names = self.input_embodiment_description[data_type]
             trained_statistics = self.input_dataset_statistics.get(data_type)
             if trained_statistics is None:
                 raise ValueError(
                     f"Model was not trained with input statistics for {data_type}."
                 )
             max_items_trained_on = len(trained_statistics)
-            if max_items_for_this_data_type > max_items_trained_on:
+            max_items_received = len(sync_point.data[data_type])
+            if max_items_received > max_items_trained_on:
                 raise ValueError(
-                    f"Received {max_items_for_this_data_type} items for data type "
+                    f"Received {max_items_received} items for data type "
                     f"{data_type}, but model was trained on maximum of "
                     f"{max_items_trained_on} items."
                 )
-            inputs_mask[data_type] = torch.tensor(
-                [1.0] * max_items_for_this_data_type
-                + [0.0] * (max_items_trained_on - max_items_for_this_data_type),
-                dtype=torch.float32,
+
+            mask: list[float] = [0.0] * max_items_trained_on
+            batched_nc_data_class: type[BatchedNCData] = (
+                DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type]
             )
-            inputs_mask[data_type].unsqueeze_(0)  # Add batch dimension
-            for _, nc_data in sync_point.data[data_type].items():
-                tensor = DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type].from_nc_data(
-                    nc_data
-                )
+            preprocessing_methods = self.input_preprocessing_config.get(data_type, [])
+
+            for index in range(max_items_trained_on):
+                name = embodiment_names.get(index)
+                if name is None or name not in sync_point.data[data_type]:
+                    tensor = batched_nc_data_class.sample(batch_size=1, time_steps=1)
+                else:
+                    nc_data = sync_point.data[data_type][name]
+                    tensor = batched_nc_data_class.from_nc_data(nc_data)
+                    mask[index] = 1.0
+
                 tensor = apply_preprocessing_methods(
                     batched_data=tensor,
-                    methods=self.input_preprocessing_config.get(data_type, []),
+                    methods=preprocessing_methods,
                 )
                 inputs[data_type].append(tensor)
+
+            inputs_mask[data_type] = torch.tensor(mask, dtype=torch.float32).unsqueeze(
+                0
+            )
         return BatchedInferenceInputs(
             inputs=inputs,
             inputs_mask=inputs_mask,
--- a/tests/unit/ml/utils/test_policy_inference.py
+++ b/tests/unit/ml/utils/test_policy_inference.py
@@ -178,6 +178,10 @@ def test_validate_input_sync_point_raises_when_required_data_type_missing() -> N
 
 def test_preprocess_builds_inputs_and_masks_for_multiple_data_types() -> None:
     policy_inference = _make_policy_inference()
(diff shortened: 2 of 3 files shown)


neuracore/ml/utils/policy_inference.py
```diff
@@ -142,34 +142,42 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
-                + [0.0] * (max_items_trained_on - max_items_for_this_data_type),
-                dtype=torch.float32,
-            )
-            inputs_mask[data_type].unsqueeze_(0)  # Add batch dimension
-            for _, nc_data in sync_point.data[data_type].items():
-                tensor = DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type].from_nc_data(
-                    nc_data
-                )
+
+            batched_nc_data_class = DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type]
```
aditya-neuraco: Can you add type annotation here?

neuracore/ml/utils/policy_inference.py
```diff
@@ -142,34 +142,42 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
         # We need to go from sync_point (single time step) to BatchedNCData
         for data_type in sync_point.data.keys():
             inputs[data_type] = []
-            max_items_for_this_data_type = len(sync_point.data[data_type])
+            mask: list[float] = []
+            embodiment_names = self.input_embodiment_description[data_type]
```
aditya-neuraco, line 145: We should do this for the output embodiment description as well.

neuracore/ml/utils/policy_inference.py
```diff
@@ -142,34 +142,42 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
             if trained_statistics is None:
                 raise ValueError(
                     f"Model was not trained with input statistics for {data_type}."
                 )
             max_items_trained_on = len(trained_statistics)
-            if max_items_for_this_data_type > max_items_trained_on:
+            if len(sync_point.data[data_type]) > max_items_trained_on:
                 raise ValueError(
-                    f"Received {max_items_for_this_data_type} items for data type "
+                    f"Received {len(sync_point.data[data_type])} items for data type "
```
aditya-neuraco: In line 143, you can access this by doing `.items()` on `sync_point.data`. So we can avoid these two function calls. You can call it as `max_items_for_this_data_type` like before. It doesn't seem to be used anywhere else.

neuracore/ml/utils/policy_inference.py
```diff
@@ -142,34 +142,42 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
         for data_type in sync_point.data.keys():
             inputs[data_type] = []
-            max_items_for_this_data_type = len(sync_point.data[data_type])
+            mask: list[float] = []
+            embodiment_names = self.input_embodiment_description[data_type]
             trained_statistics = self.input_dataset_statistics.get(data_type)
             if trained_statistics is None:
                 raise ValueError(
                     f"Model was not trained with input statistics for {data_type}."
                 )
```
aditya-neuraco, line 150: This should include both input and output stats right?

neuracore/ml/utils/policy_inference.py
```diff
@@ -142,34 +142,42 @@ def _preprocess(self, sync_point: SynchronizedPoint) -> BatchedInferenceInputs:
...
+
                 tensor = apply_preprocessing_methods(
                     batched_data=tensor,
-                    methods=self.input_preprocessing_config.get(data_type, []),
+                    methods=preprocessing_methods,
                 )
                 inputs[data_type].append(tensor)
+
+            inputs_mask[data_type] = torch.tensor(mask, dtype=torch.float32)
+            inputs_mask[data_type].unsqueeze_(0)  # Add batch dimension
```
aditya-neuraco: Nit: IMO, this is a better way to write this function. It's about 7% faster.
- Initialize the `mask` with zeros
- flip the if condition where we append to the mask.
- Chain these two lines: 
```python
inputs_mask[data_type] = torch.tensor(mask, dtype=torch.float32)
inputs_mask[data_type].unsqueeze_(0)  # Add batch dimension
```



```python
  for data_type in sync_point.data.keys():
      inputs[data_type] = []
      embodiment_names = self.input_embodiment_description[data_type]
      trained_statistics = self.input_dataset_statistics.get(data_type)
      if trained_statistics is None:
          raise ValueError(
              f"Model was not trained with input statistics for {data_type}."
          )
      max_items_trained_on = len(trained_statistics)
      if len(sync_point.data[data_type]) > max_items_trained_on:
          raise ValueError(
              f"Received {len(sync_point.data[data_type])} items for data type "
              f"{data_type}, but model was trained on maximum of "
              f"{max_items_trained_on} items."
          )

      mask: list[float] = [0.0] * max_items_trained_on
      batched_nc_data_class = DATA_TYPE_TO_BATCHED_NC_DATA_CLASS[data_type]
      preprocessing_methods = self.input_preprocessing_config.get(data_type, [])

      for index in range(max_items_trained_on):
          name = embodiment_names.get(index)
          if name is None or name not in sync_point.data[data_type]:
              tensor = batched_nc_data_class.sample(batch_size=1, time_steps=1)
          else:
              nc_data = sync_point.data[data_type][name]
              tensor = batched_nc_data_class.from_nc_data(nc_data)
              mask[index] = 1.0

          tensor = apply_preprocessing_methods(
              batched_data=tensor,
              methods=preprocessing_methods,
          )
          inputs[data_type].append(tensor)

      inputs_mask[data_type] = torch.tensor(mask, dtype=torch.float32).unsqueeze(0)
```

tests/unit/ml/utils/test_policy_inference.py
aditya-neuraco, line 1: Add `output_embodiment_description` logic and tests for it as well.

aditya-neuraco (approved):
LGTM w/ comments.

neuracore/ml/utils/policy_inference.py
ypang-neuraco: renamed to max_items_received but we need to keep `sync_point.data[data_type]` because we need to count the items in the data_type, not the number of datatypes in the sync_point

neuracore/ml/utils/policy_inference.py
ypang-neuraco: Discussed offline, we don't need to pad the output because we can just drop the extra data

tests/unit/ml/utils/test_policy_inference.py
ypang-neuraco: Discussed offline, we don't need to pad the output because we can just drop the extra data

neuracore/ml/utils/policy_inference.py
ypang-neuraco: Todo when normalisation move to preprocessing

PR #664, opened 2026-06-03
test: skip daemon tests and add sigkill retries

damonhayhurst-neuraco:
### Features

 - tests need to retry sigkill until all candidate pids are gone

### Bugfixes

 - tests now work and those that don't are skipped


muneebneura (approved):
LGTM

PR #665, opened 2026-06-03
test: add auto batch size integration test

aditya-neuraco:
### Features

 - Added workflows and test for auto batch size tuning.

### Items

 - [NCRL-590](https://neuracore.atlassian.net/browse/NCRL-590)

### DIFF
--- a/tests/integration/ml/test_auto_batch_size.py
+++ b/tests/integration/ml/test_auto_batch_size.py
@@ -1,0 +1,308 @@
+"""Integration tests for automatic batch-size selection during training.
+
+This test submits training jobs with batch_size set to auto for representative
+algorithm, GPU type, and GPU count combinations. It verifies that each job
+reaches COMPLETED.
+The test is split into two phases:
+
+  test_start_training submits a training job and records the job ID.
+  test_wait_for_training_terminal_status waits for that job and verifies that
+  it completes successfully.
+
+Case definitions live in auto_batch_size_cases.yaml.
+"""
+
+from __future__ import annotations
+
+import atexit
+import logging
+import os
+import pprint
+import time
+from pathlib import Path
+
+import pytest
+import yaml
+from neuracore_types import DataType
+
+import neuracore as nc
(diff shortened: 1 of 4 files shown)


tests/integration/ml/test_auto_batch_size.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+            job_data = nc.get_training_job_data(training_job_id)
+
+            if final_status != "COMPLETED":
+                error_logs = None
+                try:
+                    error_logs = nc.get_training_job_logs(
+                        training_job_id,
+                        max_entries=200,
+                        severity_filter="ERROR",
```
ypang-neuraco, line 282: Maybe we should get the full log not just the error? e.g. what batch size was tested before the error

tests/integration/ml/test_auto_batch_size.py
```diff
@@ -0,0 +1,308 @@
...
+                    f"job_data={pprint.pformat(job_data)}\n"
+                    f"error_logs={pprint.pformat(error_logs)}"
+                )
+
+            logger.info(
+                f"[{test_case}] Training job {training_job_id} completed "
+                f"successfully"
+            )
+        finally:
+            cleanup_training_job(test_case, training_job_id)
```
ypang-neuraco, line 303: I think it's better to just keep the job around if it fails. That way we can look at the final resolved configs, we can download any mode artifacts, we can look at all logs/warnings not just errors etc.

ypang-neuraco (approved):
LGTM \w comments

tests/integration/ml/test_auto_batch_size.py
aditya-neuraco: It's there in the logs.

tests/integration/ml/test_auto_batch_size.py
aditya-neuraco: I think it's just better to delete it, otherwise it just adds unnecessary dead VMs.

PR #666, opened 2026-06-04
fix: improve error message when using robot from different org

aditya-neuraco:
### Bugfixes

 - Better error message when using a model trained on robots in different org but doing inference in users org.

### Items

 - [NCRL-578](https://neuracore.atlassian.net/browse/NCRL-578)


ypang-neuraco (approved):
LGTM

PR #668, opened 2026-06-04
test: run matrix jobs sequentially

muneebneura:
Running integration tests sequentially should not cause these sse overlapping issues.

All details for this pr are here: https://docs.google.com/document/d/1nZntudesqXJpK_tYUOTupUPzAhG7ZJbgBbC2bJ-TnUA/edit?tab=t.0

Also removed our previous fix to hunt and kill any alive pids in the tests which were randomly coming up because an sse would trigger a start daemon from other parallel job.

[NCP-1021](https://neuracore.atlassian.net/browse/NCP-1021)

### DIFF
--- a/.github/workflows/integration-platform-production.yaml
+++ b/.github/workflows/integration-platform-production.yaml
@@ -1,4 +1,4 @@
-name: Integration Platform
+name: Integration Platform (Production)
 
 on:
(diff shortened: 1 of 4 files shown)


stepjam (approved):
LGTM

PR #669, opened 2026-06-04
ci: fix auto batch size integration test workflow

aditya-neuraco:
- Workflow used parameters form an initial iteration of auto batch size test so tests were failing.


stepjam (approved):
Lgtm

PR #670, opened 2026-06-05
fix: only start data daemon for recordings on connected robots

muneebneura:
This issue has been revealed from the integration tests running in parallel using the same org,

### Bugfixes

 - we now only start the data daemon when the recording is for a robot this client actually connected to. Recordings from other tests or robots in the org are ignored.

### Items

 - [NCP-1022](https://neuracore.atlassian.net/browse/NCP-1022)

### DIFF
--- a/neuracore/core/streaming/recording_state_manager.py
+++ b/neuracore/core/streaming/recording_state_manager.py
@@ -81,6 +81,7 @@ class RecordingStateManager(BaseSSEConsumer):
         self.org_id = org_id or get_current_org()
         self.auth = auth if auth is not None else get_auth()
 
+        self._connected_robot_id: str | None = None
         self.recording_robot_instances: dict[RobotInstanceIdentifier, str] = dict()
         self._expired_recording_ids: set[str] = set()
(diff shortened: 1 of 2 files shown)


neuracore/core/streaming/recording_state_manager.py
```diff
@@ -270,6 +271,9 @@ def updated_recording_state(
                 details, RecordingStartPayload
             ), "recording must be started by a start event"
 
+            if robot_id not in self._connected_robot_ids:
```
stepjam: please add a comment above explaining why this line is so important :)

neuracore/core/streaming/recording_state_manager.py
```diff
@@ -81,6 +81,7 @@ def __init__(
         self.org_id = org_id or get_current_org()
         self.auth = auth if auth is not None else get_auth()
 
+        self._connected_robot_ids: set[str] = set()
```
stepjam: discussed offline. Only can connect to 1 robot at a time, so this can just be a single str. Make sure to update ln 273 to

stepjam (approved):
LGTM w/c

PR #671, opened 2026-06-05
fix: preserve RGB bytes and unskip network integrity tests

favour-neuraco:
### Bugfixes
- Updated the lossless RGB video encoding path to preserve byte-exact RGB frame data through cloud upload/sync.
- Switched the lossless RGB encoder from `libx264` with `yuv444p10le` to `libx264rgb` with `rgb24` and `crf=0`, avoiding RGB to YUV colour-space conversion that could alter pixel values.
- Unskipped the RGB network data integrity cases now that synced RGB frame-code verification passes with the strict byte-level assertion intact.

https://neuracore.atlassian.net/browse/NCP-1024

### DIFF
--- a/neuracore/core/streaming/bucket_uploaders/streaming_video_uploader.py
+++ b/neuracore/core/streaming/bucket_uploaders/streaming_video_uploader.py
@@ -56,8 +56,8 @@ class StreamingVideoUploader(BucketUploader):
         width: int,
         height: int,
         transform_frame: Callable[[np.ndarray], np.ndarray] | None = None,
-        codec: str = "libx264",
-        pixel_format: str = "yuv444p10le",
+        codec: str = "libx264rgb",
+        pixel_format: str = "rgb24",
         chunk_size: int = CHUNK_SIZE,
         video_name: str = "lossless.mp4",
         codec_context_options: dict[str, str] | None = None,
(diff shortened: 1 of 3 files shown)


stepjam (commented):
If we indeed are swapping the codec, why are these files also not getting updated?

https://github.com/search?q=repo%3ANeuracoreAI%2Fneuracore%20libx264&type=code

stepjam:
Same here:
https://github.com/search?q=repo%3ANeuracoreAI%2Fneuracore_backend%20libx264&type=code

favour-neuraco:
> Same here: https://github.com/search?q=repo%3ANeuracoreAI%2Fneuracore_backend%20libx264&type=code

synchronized_lossy_video_manager.py - left unchanged, because it intentionally generates frontend-friendly lossy videos.

Backend synchronized lossy video tests - left unchanged, because they test lossy playback generation, not byte-exact RGB preservation.

favour-neuraco:
> If we indeed are swapping the codec, why are these files also not getting updated?
> 
> https://github.com/search?q=repo%3ANeuracoreAI%2Fneuracore%20libx264&type=code

streaming_video_uploader.py - updated, because it defaults to lossless.mp4.

disk_video_encoder.py - left unchanged, because it is a generic encoder wrapper and callers pass the codec/pixel format.

test_disk_video_encoder.py - left unchanged, because it only tests FFmpeg/PyAV factory selection with dummy codec kwargs.

test_encoding_traces.py - left unchanged, existing tests patch out real encoding and cover trace behaviour

stepjam (approved):
LGTM

PR #672, opened 2026-06-06
test: update training flow start time

ypang-neuraco:
### Bugfixes
- Integration test for training flow fails because of 429 hugging face rate limit error when trying to download the BERT model
- This fix delays the start of training flow tests and stagger the staging and prod tests by 1 hour to allow the rate limit to reset


stepjam (approved):
LGTM

PR #673, opened 2026-06-08
fix: iceroxy2

StevenJacobs61:
### Features

 - Replaced shared slot transport with iceroxy2 implementation for video payloads.
 - A lot of the files changed are name changes or deletions


CougarTasker:
Closed as out of date and to avoid instabililty in the legacy path

PR #674, opened 2026-06-08
test: refactor training flow tests

ypang-neuraco:
### Features
- Refactor test_training_flow.py to smaller files
- Add tests TestResumeTrainingAfterDataAddition and TestResumeTrainingAfterDataDeletion
- Inference and resume test now uses latest successful run (i.e. if latest run was a failure, get the model from the last successful run instead)

### Items
 - [NCRL-600](https://neuracore.atlassian.net/browse/NCRL-600)
 - [NCRL-586](https://neuracore.atlassian.net/browse/NCRL-586)

### DIFF
--- a/.github/workflows/integration-ml-back-to-back-training-prod.yaml
+++ b/.github/workflows/integration-ml-back-to-back-training-prod.yaml
@@ -1,22 +1,12 @@
-name: Integration ML Training Flow - Production
+name: Integration ML Back To Back Training - Production
 
 on:
   schedule:
-    - cron: "0 2 * * *"  # 1. test_training_flow.py
+    - cron: "0 6 * * *"  # 5. test_back_to_back_training.py
   workflow_dispatch:
 
 jobs:
-  pre-commit:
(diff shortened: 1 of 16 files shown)


tests/integration/ml/shared/sim.py
aditya-neuraco, line 1: We don't need this file. Better to import these in each test in case we want to change some test later on.

tests/integration/ml/shared/constants.py
aditya-neuraco, line 1: I also don't think we should share these constants between tests. Better to keep them isolated.

aditya-neuraco (approved):
LGTM w/ comments.

I don't think we should be sharing some components between tests. It's better if tests are isolated from each other even though they have shared logic.

ypang-neuraco:
> LGTM w/ comments.
> 
> I don't think we should be sharing some components between tests. It's better if tests are isolated from each other even though they have shared logic.

Agreed, removed as much from shared as possible, only keeping helpers.

PR #678, opened 2026-06-08
feat: simplify synchronisation reporting

damonhayhurst-neuraco:
### Features

 - Synchronisation now fails early so handle contract as such


CougarTasker (approved):
LGTM 👍

PR #679, opened 2026-06-08
feat: expose data_types on client Recording wrapper

aditya-neuraco:
## Features
- Adds `data_types: set[DataType]` field to the client `Recording` wrapper class
- Populates it from `RecordingModel.data_types` in `Dataset._wrap_raw_recording`
- Enables downstream per-recording validation to inspect which datatypes each recording contains

## Items
- [NCRL-604](https://neuracore.atlassian.net/browse/NCRL-604)

## Related PRs
- #680
- #681

PR #680, opened 2026-06-08
feat: add per-recording datatype validation before training run

aditya-neuraco:
## Features
- Adds `_validate_per_recording_data_types(dataset, requested_input_types)` that iterates all recordings and collects missing datatypes grouped by type
- Wires it into `validate_training_params` so it fires before the training job API call
- Raises `ValueError` with a grouped message listing which recordings are missing which types

**Example error:**
```
Failed to start training run: some recordings are missing requested datatypes.

Missing JOINT_POSITIONS:
- Recording A
- Recording B

Missing RGB_IMAGES:
- Recording C
```

## Items
- [NCRL-604](https://neuracore.atlassian.net/browse/NCRL-604)

## Related PRs
- #679 (needs to go in first)
- #681 (needs to go in after this)

### DIFF
--- a/neuracore/core/utils/training_input_args_validation.py
+++ b/neuracore/core/utils/training_input_args_validation.py
@@ -6,11 +6,45 @@ algorithm existence, robot existence, and data spec compatibility.
 
 from __future__ import annotations
 
+from collections import defaultdict
+
 from neuracore_types import CrossEmbodimentDescription, DataType
 from omegaconf import DictConfig
 
 from neuracore.core.data.dataset import Dataset, EmbodimentDescription
 from neuracore.core.utils.embodiment_description_utils import is_robot_id
+
+
+def _validate_per_recording_data_types(
+    dataset: Dataset,
+    requested_input_types: set[DataType],
+) -> None:
+    """Validate every recording has all requested input datatypes.
+
+    Args:
+        dataset: Dataset whose recordings will be inspected.
+        requested_input_types: DataTypes that every recording must contain.
+
+    Raises:
+        ValueError: If any recording is missing a requested type, with a
+            grouped message listing affected recording names per type.
+    """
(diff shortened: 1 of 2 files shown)


neuracore/core/utils/training_input_args_validation.py
```diff
@@ -6,13 +6,51 @@
...
+            grouped message listing affected recording names per type.
+    """
+    if not requested_input_types:
+        return
+
+    missing_per_type: dict[DataType, list[str]] = defaultdict(list)
+    for recording in dataset:
+        for data_type in requested_input_types:
+            if data_type not in recording.data_types:
+                missing_per_type[data_type].append(recording.name)
```
ypang-neuraco: Since they are both sets, using set difference might be slightly faster e.g. `missing_data_types=requested_input_types-recording.data_types`

neuracore/core/utils/training_input_args_validation.py
```diff
@@ -241,6 +279,13 @@ def validate_training_params(
     Raises:
         ValueError: If any validation check fails.
     """
+    requested_input_types: set[DataType] = {
+        dt
+        for robot_data in input_cross_embodiment_description.values()
+        for dt in robot_data.keys()
+    }
```
ypang-neuraco, line 282: Should we not raise an error if requested_input_types is empty?

ypang-neuraco (commented):
LGTM \w comments

ypang-neuraco (approved):
LGTM \w comments

neuracore/core/utils/training_input_args_validation.py
aditya-neuraco: Yeah, good catch!

neuracore/core/utils/training_input_args_validation.py
aditya-neuraco: Okay! Changed to difference of sets.

PR #687, opened 2026-06-09
test: move daemon test storage state to context managers

damonhayhurst-neuraco:
### Features

 - Move test storage profile to the runners so any test using the daemon will use the same reference to test storage location (that is gitignored)


CougarTasker (approved):
LGTM 👍

PR #688, opened 2026-06-10
fix: send client-side start/end time on recording start/stop

aditya-neuraco:
### Features

 - None

### Bugfixes

 - Counterpart to https://github.com/NeuracoreAI/neuracore_backend/pull/373, RecordingStartRequest requires start_time and /recording/stop now takes a JSON body with recording_id and end_time. Without this, every start_recording call against backend main fails validation.
 - Also fixes the already-started warning heuristic: start_time is now always present in the response, so detect a pre-existing recording by comparing the returned start_time with the one we sent.

### Items

 - [NCRL-623](https://neuracore.atlassian.net/browse/NCRL-623)

### Related PRs

 - https://github.com/NeuracoreAI/neuracore_backend/pull/373

### DIFF
--- a/neuracore/core/robot.py
+++ b/neuracore/core/robot.py
@@ -11,6 +11,7 @@ import io
 import logging
 import os
 import tempfile
+import time
 import xml.etree.ElementTree as ET
 import zipfile
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


CougarTasker (approved):
LGTM 👍

PR #689, opened 2026-06-10
fix: ignore duplicate RECORDING_STOPPED in trace lifecycle coordinator

aditya-neuraco:
### Bugfixes

 - A duplicate RECORDING_STOPPED message re-entered the closing set after the first close had already cleared the unique-trace registry. The second close then reported expected_trace_count=0 to the backend, overwriting the correct count and stranding the recording as pending forever (all traces uploaded but the save task never enqueued).

cc @muneebneura 

### Items

 - [NCRL-622](https://neuracore.atlassian.net/browse/NCRL-622)


CougarTasker (approved):
LGTM 👍

PR #691, opened 2026-06-10
fix: check shared status when creating dataset

ypang-neuraco:
### Bugfixes
- Creating private dataset while having a shared dataset with the same name was not allowed
- Creating shared dataset while having a private dataset with the same name was not allowed
- Fix: check shared status of found dataset

### Items
 - [NCRL-583](https://neuracore.atlassian.net/browse/NCRL-583)

### DIFF
--- a/neuracore/core/data/dataset.py
+++ b/neuracore/core/data/dataset.py
@@ -311,11 +311,12 @@ class Dataset:
         tags: list[str] | None = None,
         shared: bool = False,
     ) -> "Dataset":
-        """Create a new dataset or return existing one with the same name.
+        """Create a new dataset or return an existing one in the same namespace.
 
         Creates a new dataset with the specified parameters. If a dataset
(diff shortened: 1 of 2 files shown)


neuracore/core/data/dataset.py
```diff
@@ -326,11 +327,11 @@ def create(
...
 
         Returns:
-            The newly created Dataset instance, or existing dataset if
-            name already exists.
+            The newly created Dataset instance, or the existing dataset if one
+            already exists in the same namespace.
         """
         ds = Dataset.get_by_name(name, non_exist_ok=True)
-        if ds is None:
+        if ds is None or shared != ds.is_shared:
```
aditya-neuraco: Can we add a logger warning here that tells user we are doing this?

aditya-neuraco (approved):
LGTM w/ comment

PR #692, opened 2026-06-10
chore: update neuracore types to v8

favour-neuraco:
Updated `neuracore_types` to `8.0.0` so the client package uses the v8 recording metadata contract that removes `RecordingMetadata.name`.


aditya-neuraco (approved):
LGTM!

PR #695, opened 2026-06-11
feat: add api for cloning dataset

sdas-neuraco:
### Features

 - New API for cloning datasets in Neuracore via CLI

### DIFF
--- a/neuracore/api/datasets.py
+++ b/neuracore/api/datasets.py
@@ -12,6 +12,7 @@ from neuracore.core.config.get_current_org import get_current_org
 from neuracore.core.const import API_URL
 from neuracore.core.data.dataset import Dataset
 from neuracore.core.exceptions import DatasetError
+from neuracore.core.utils.http_errors import extract_error_detail
 from neuracore.core.utils.http_session import thread_local_session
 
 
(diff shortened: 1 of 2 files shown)


neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
     return merged
 
 
+def dataset_clone(
```
muneebneura: Suggestion: Rename this to clone_dataset to follow naming conventions. We have other functions like get_dataset, create_dataset etc

neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
...
+    # Checks only one of dataset, dataset_name, or dataset_id is provided
+    assert (
+        sum(arg is not None for arg in [dataset, dataset_name, dataset_id]) == 1
+    ), "Exactly one of dataset, dataset_name, or dataset_id must be provided"
+
+    # If dataset_name is provided, resolve it to a dataset object
+    if dataset_name is not None:
+        dataset = Dataset.get_by_name(dataset_name)
+        if dataset is None:
+            raise DatasetError(f"Source dataset with name '{dataset_name}' not found")
```
muneebneura: Issue: This if condition is unreachable. dataset above will raise DatasetError on a 404 it never returns none.

neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
...
+        id=dataset_model.id,
+        org_id=org_id,
+        name=dataset_model.name,
+        size_bytes=dataset_model.size_bytes,
+        tags=dataset_model.tags,
+        is_shared=dataset_model.is_shared,
+        description=dataset_model.description,
+        data_types=list(dataset_model.all_data_types.keys()),
+    )
+
```
muneebneura, line 157: Question: I have checked both create_dataset and merge_dataset which set the global active dataset after creating one. Do we want to follow the same pattern here?

neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
...
+    if dataset_name is not None:
+        dataset = Dataset.get_by_name(dataset_name)
+        if dataset is None:
+            raise DatasetError(f"Source dataset with name '{dataset_name}' not found")
+
+    # No need to resolve dataset_id here because the API can handle it directly,
+    # and this avoids an unnecessary API call if the user already has the ID.
+
+    # Avoid Dataset truthiness here because it delegates to __len__ and may
+    # fetch recordings from the API.
```
muneebneura: Suggestion: These comments aren't very helpful, I would suggest removing them to keep code clean.

neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
...
+    *,
+    dataset: Dataset | None = None,
+    dataset_name: str | None = None,
+    dataset_id: str | None = None,
+) -> Dataset:
+    """Clone a dataset and all of its recordings.
+
+    Args:
+        dataset: Source dataset object, dataset ID, or dataset name. String
+            values are resolved as an ID first and then as a name.
```
muneebneura: Suggestion:  `Source dataset object to clone.`

neuracore/api/datasets.py
```diff
@@ -90,6 +91,78 @@ def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
...
+        new_dataset_name: Name for the cloned dataset.
+        dataset_name: Explicit source dataset name.
+        dataset_id: Explicit source dataset ID.
+
+    Returns:
+        Dataset: The newly cloned dataset.
+
+    Raises:
+        TypeError: If dataset is not a Dataset or string.
```
muneebneura: Question: I don't see this code raising a type error?

muneebneura (changes requested):
LGTM w/comments

neuracore/api/datasets.py
sarthakdas: removed

neuracore/api/datasets.py
sarthakdas: I would keep the recond half of the comment, as it could catch people out if they change it to [ if dataset: ] becuase the if dataset is not None is a bit of a werid setup

neuracore/api/datasets.py
sarthakdas: yeah i guess so, this seems like a recipe for problems and it could be confusing to the user which dataset is active

muneebneura: Yeah, but lets keep the behavior consistent with the other apis

muneebneura (approved):
LGTM! 👍

PR #700, opened 2026-06-11
fix: fix the training data out of index error. 

kwangneuraco:
We had this `index out of range` error during training, it happened frequently around 50% of time. The issue get fixed and improved by:

- It turns out to be the race condition of concurrent workers fetching the synchronised cached data and not protected. I added the lock to prevent this.
- Prevent the double prefetch happen by adding a syncedDataset cache
- Add test with regard to these

### DIFF
--- a/neuracore/core/data/dataset.py
+++ b/neuracore/core/data/dataset.py
@@ -2,6 +2,7 @@
 
 import logging
 import sys
+import threading
 import time
 from collections.abc import Generator, Iterator
 from typing import Optional, Union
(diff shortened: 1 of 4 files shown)


ypang-neuraco:
Discussed offline, I think if the issue is a race condition on prefetch, the fix should not be to add a lock, we should fix the splitting of recordings across workers as there should not be any overlap

sdas-neuraco:
I have checked this against another dataset for Conq Hose on Dev and seems to be working for me, got rid of index error.

kwangneuraco:
Let me dig deeper to find the real root cause

ypang-neuraco (approved):
LGTM, please clean up the comments before merging

sdas-neuraco:
I have noticed a degradation of speed when converting to a pytorch dataset with this PR

kwangneuraco:
> I have noticed a degradation of speed when converting to a pytorch dataset with this PR

How much degradation it will be? It will slowdown because it protect the page(batch of episodes) from being accessed from different threads. Previously multiple threads can access the same page and quickly fill up the cache until it is out of range. So this is expected lol.

PR #704, opened 2026-06-12
feat: direct url for banner image

CougarTasker:
### Bugfixes

 - Banner Image is missing on PyPi I belive this may because it is a relitive not absolute path 
 
 
 image


aditya-neuraco (approved):
LGTM!

PR #706, opened 2026-06-12
feat: made version validation more consistent

CougarTasker:
### Features

 - Verifiy version during login so it applies regardless of how you use neuracore

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -164,7 +164,6 @@ def connect_robot(
     Returns:
         The initialized and connected robot instance.
     """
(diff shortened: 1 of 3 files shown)


ypang-neuraco (approved):
LGTM. Should we remove `validate_version` from `core.py` since it is not used anymore?

CougarTasker:
> LGTM. Should we remove `validate_version` from `core.py` since it is not used anymore?

To keep the change backwards compatible I will keep it, that method could be use like `nc.validate_version()` by a user

PR #707, opened 2026-06-12
fix: prevent closed recordings from re-entering the closing set

aditya-neuraco:
### Features

 - None

### Bugfixes

 - `RecordingCloseRegistry.mark_closing` now ignores recordings that are already closed. Previously any code path could re-enter a closed recording into the closing set, allowing a second close to report `expected_trace_count=0` after the unique-trace history was cleared, stranding the recording as pending on the backend. The registry now enforces the closed state itself, complementing the duplicate-stop guard from #689.
 - Added `tests/unit/data_daemon/communications_management/test_recording_close_registry.py` covering the close transition, re-mark rejection, and isolation between recordings.

### Items

 - None yet

### Related PRs

 - NeuracoreAI/neuracore_backend#412

### DIFF
--- a/neuracore/data_daemon/communications_management/consumer/models.py
+++ b/neuracore/data_daemon/communications_management/consumer/models.py
@@ -835,7 +835,14 @@ class RecordingCloseRegistry:
         recording_id: str,
         closing_state: RecordingClosingState,
     ) -> None:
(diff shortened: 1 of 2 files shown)


CougarTasker (approved):
LGTM 👍

PR #709, opened 2026-06-13
fix: capture end time before stream drain

muneebneura:
### Bugfixes

 - Updated cancel recording api with the new shape
 - end_time is now captured before we call drain steams and notify daemon
 - Updated duration assertion as recording start and stop times are sent from client side.

### Items

 - [NCP-1045](https://neuracore.atlassian.net/browse/NCP-1045)

Successful test run here on thios branch: https://github.com/NeuracoreAI/neuracore/actions/runs/27464202095/job/81183523997

### Related PRs

 - None

### DIFF
--- a/neuracore/core/robot.py
+++ b/neuracore/core/robot.py
@@ -346,6 +346,7 @@ class Robot:
         if not self.id:
             raise RobotError("Robot not initialized. Call init() first.")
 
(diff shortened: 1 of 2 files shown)


aditya-neuraco (approved):
LGTM!

PR #711, opened 2026-06-15
test: increase recording duration assertion

muneebneura:
- Increased the clock tolerance for recording duration assertion in platform integration tests.
 - The tests were being flaky because our recording duration now depends on the start_recording api call as start and end time is sent from client side, so I have increased it to a max number we should allow.
 
 **Note:** We also need to look into improving the start_recording api as this latency comes from a chain of sequential firestore round trips. I have created a ticket for this here: https://neuracore.atlassian.net/browse/NCP-1046


favour-neuraco (approved):
LGTM

PR #715, opened 2026-06-16
feat: add background dataset clone warning

sdas-neuraco:
## What changed

- Adds a `wait` option to `clone_dataset`, defaulting to the existing synchronous behavior.
- Polls clone progress with a `tqdm` progress bar when waiting for recordings to become available.
- Logs a warning when cloning is left to continue in the background so callers know recordings may not be available immediately.

### DIFF
--- a/neuracore/api/datasets.py
+++ b/neuracore/api/datasets.py
@@ -4,7 +4,11 @@ This module provides functions for creating and retrieving datasets
 for robot demonstrations.
 """
 
+import logging
+import time
+
 from neuracore_types import Dataset as DatasetModel
+from tqdm import tqdm
 
 from neuracore.api.globals import GlobalSingleton
 from neuracore.core.auth import get_auth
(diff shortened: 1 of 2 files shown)


neuracore/api/datasets.py
```diff
@@ -155,6 +165,38 @@ def clone_dataset(
...
+                        pbar.update(new_cloned_recordings - cloned_recordings)
+                        cloned_recordings = new_cloned_recordings
+            finally:
+                pbar.close()
+    else:
+        logger.warning(
+            "Dataset cloning is running in the background; recordings may not be "
+            "available immediately."
+        )
+
```
muneebneura: Nesting issue: I would suggest refactoring this function to avoid nesting. You can invert the wait block as an early return which would simplify this function. 

Something like:

```
if not wait:
        logger.warning(
            "Dataset cloning is running in the background; recordings may not be "
            "available immediately."
        )
        GlobalSingleton()._active_dataset_id = cloned.id
        return cloned
```

muneebneura (approved):
LGTM w/comments

PR #716, opened 2026-06-16
fix: flakey trace unit test

CougarTasker:
### Bugfixes

 - These tests were flakey under CI due to a race condition between the thread pool executor returning and the deadline
    - when the machine was under load this was much more likely to happen basically never happened on a unloaded machine

check the runs of this to see the error in pracice 

https://github.com/NeuracoreAI/neuracore/actions/runs/27625254770/job/81689794677

### Items

 - [NCP-1047](https://neuracore.atlassian.net/browse/NCP-1047)


damonhayhurst-neuraco (approved):
LGTM

PR #721, opened 2026-06-18
feat: rewrite auto batch size tuning

kwangneuraco:
### Features

- Rewrite batch-size tuning with an affine memory model. Replace the binary-search autotuner with a memory-model estimator: probe a few batch sizes, fit peak GPU memory as an affine function of batch size, and solve for the batch that fills the VRAM budget, then apply a safety factor. 

- Add docs/batch_size_autotuning.md: https://github.com/NeuracoreAI/neuracore/blob/feat/rewrite_autobatch/docs/batch_size_autotuning.md

### Test
I have ran the autobatch size on one A100 40 GB with the `lemon pick and place` dataset. The dataset has 496 episodes, 125.19 GB big. Here is the results:

<b style="font-weight:normal;" id="docs-internal-guid-17458e01-7fff-03da-4fbc-80b48f38a08a"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">&nbsp;&nbsp;</span></p><div dir="ltr" style="margin-left:0pt;" align="left">
  | New Algo Memory Usage | New Algo Time(min) / Iterations
-- | -- | --
CNNMLP | 26.9G/40G | 8.35 / 4
ACT | 29.9G/40G | 7.31 / 4
Diffusion Policy | 26G/40G | 12 / 4
Pi0 | 30.9G/40G | 15.2 / 4
Pi05 | 30.9G/40G | 10 / 4
Groot | 31.74G/40G | 14.7 / 4

</div></b>

It is much faster than the previous binary search approach, which took around 3 hours for Diffusion Policy

### Items
https://neuracore.atlassian.net/browse/NCRL-617


tests/unit/ml/test_batch_autotuner.py
```diff
@@ -136,10 +109,55 @@ def fake_random_split(dataset, lengths, generator=None):
     assert result == 4
     mock_autotuner_cls.assert_called_once()
     kwargs = mock_autotuner_cls.call_args.kwargs
-    assert kwargs["min_batch_size"] == 2
+    assert kwargs["min_batch_size"] == 1
     assert kwargs["max_batch_size"] == 80  # clamp to train dataset length
 
 
+def test_find_optimal_batch_size_default_max_allows_above_512():
```
ypang-neuraco: Why is this test needed?

docs/batch_size_autotuning.md
aditya-neuraco, line 1: We don't really need to explain how it works and why it's needed in user facing docs. It's a nice to have.

tests/unit/ml/test_batch_autotuner.py
```diff
@@ -14,12 +14,18 @@
     PytorchSynchronizedDataset,
 )
 from neuracore.ml.trainers.batch_autotuner import (
+    BatchProbeResult,
     BatchSizeAutotuner,
     BatchSizeValidator,
+    _BatchSizeEstimationError,
+    _estimate_max_batch_size,
```
aditya-neuraco, line 21: Why are these marked as private when we are importing it in a different file?

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -500,23 +787,41 @@ def find_optimal_batch_size(
...
-    """Tune the batch size automatically via binary search."""
+    """Tune the batch size automatically via an affine memory model."""
     train_dataset, val_dataset = _split_train_val_dataset(cfg, dataset)
 
     max_batch_size = (
         cfg.max_batch_size if "max_batch_size" in cfg else len(train_dataset)
     )
     max_batch_size = min(max_batch_size, len(train_dataset))  # Clamp to train len
-    min_batch_size = cfg.min_batch_size if "min_batch_size" in cfg else 2
+    min_batch_size = cfg.min_batch_size if "min_batch_size" in cfg else 1
```
ypang-neuraco: We should also clamp min_batch_size to train len.
If either cfg.min_batch_size > len(train_dataset) or cfg.max_batch_size > len(train_dataset) we should log a warning.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -436,62 +505,280 @@ def __init__(
...
+        if not low_result.fitted:
+            raise OutOfMemoryError(
+                f"Smallest batch size {low} does not fit.",
+                device=str(self.device),
+            )
+        _log_probe(low, low_result)
+
+        # Degenerate range (e.g. max_batch_size <= min): nothing more to try.
+        if high <= low:
+            return self._apply_safety(low)
```
ypang-neuraco, line 550: This should fail early by comparing the min_batch_size and max_batch_size in the config. If this happened the user probably made a mistake in the config and should be notified, rather than allowing it and carry on.

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -375,7 +444,7 @@ def __init__(
         val_dataset: Dataset,
         train_dataloader_kwargs: dict[str, Any] | None = None,
         val_dataloader_kwargs: dict[str, Any] | None = None,
-        min_batch_size: int = 8,
+        min_batch_size: int = 1,
```
aditya-neuraco: Minimum batch size 1 breaks batch norm. BatchNorm in train mode at batch 1 raises RuntimeError: Expected more than 1 value per channel. We need to work around this or set min batch size as 2.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -436,62 +505,280 @@ def __init__(
...
         )
 
-    def find_optimal_batch_size(self) -> int:
-        """Find the optimal batch size using binary search.
+    def _downscale_to_fit(self, start: int, floor_fit: int) -> int:
+        """Scale ``start`` down by ``_DOWNSCALE_FACTOR`` until a probe fits.
+
+        ``floor_fit`` is a batch already known to fit (the search floor). Returns
+        the largest fitting batch found, reduced by ``safety_factor`` for headroom.
+        """
```
aditya-neuraco: Can you remove markdown from the docstrings?

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -436,62 +505,280 @@ def __init__(
...
-            else:
-                # This batch size failed, enter the lower half of the search range
-                high = mid - 1
-
-        # Reduce by self.safety_factor to be safe (e.g. 0.7 for 30% reduction)
-        reduced_batch_size = int(optimal_batch_size * self.safety_factor)
-        msg = (
-            f"Optimal batch size found: {optimal_batch_size}, "
-            f"Reducing by {(1 - self.safety_factor) * 100:.1f}% to {reduced_batch_size}"
+        high = min(self.min_batch_size * 8, self.max_batch_size)
```
aditya-neuraco, line 526: We should get the max bound for batch size from the available memory instead of using a heuristic number.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -27,6 +28,38 @@
...
+        total_bytes: Total VRAM of the probe device. Zero when ``fitted`` is
+            False.
+    """
+
+    fitted: bool
+    peak_reserved_bytes: int = 0
+    total_bytes: int = 0
+
+
+class _BatchSizeEstimationError(Exception):
```
aditya-neuraco, line 57: Please remove the _ from the error as this shouldn't be a private error class.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -60,38 +93,46 @@ def __init__(
...
-        logger.info(f"Testing batch size: {batch_size}")
+        logger.info("Probing batch size: %s", batch_size)
 
         # Ensure the parent GPU state is clean so the child has maximum room.
         if torch.cuda.is_available():
             torch.cuda.empty_cache()
         gc.collect()
 
-        return self._run_in_subprocess(batch_size)
+        start_time = time.perf_counter()
```
aditya-neuraco, line 112: This should be a class attribute if we want to measure it.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -60,38 +93,46 @@ def __init__(
...
+            "Probing batch size %s took %.2fs (%s)",
+            batch_size,
+            time.perf_counter() - start_time,
+            "fit" if result.fitted else "OOM",
+        )
+        return result
+
+    def test_batch_size(self, batch_size: int) -> bool:
+        """Return True if ``batch_size`` runs without an OOM-related failure."""
+        return self.probe_batch_size(batch_size).fitted
```
aditya-neuraco, line 124: This function is unnecessary.

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -542,11 +847,16 @@ def find_optimal_batch_size(
         max_batch_size=max_batch_size,
     )
 
-    # Perform binary search to find the optimal batch size
-    optimal_batch_size = autotuner.find_optimal_batch_size()
+    # Find the batch size from the affine memory model (self-sufficient: it
+    # handles OOM and overshoot internally by downscaling).
+    optimal_batch_size = autotuner.estimate_optimal_batch_size()
 
     elapsed_time = time.perf_counter() - start_time
```
aditya-neuraco, line 827: SHould also make this a class attribute.

tests/unit/ml/test_batch_autotuner.py
kwangneuraco: This is to test cases in which max batch size can exceed 512 and can be as big as dataset episode numbers. As the default value of max_batch_size is 512. I have modified the name so it is not confusing

aditya-neuraco (changes requested):
This is a great improvement over the current implementation but I would suggest these following improvements be made.

1. **Reuse one worker process.** Right now every probe spawns a new process and reloads the model and optimizer. Instead, keep one worker alive, send it batch sizes over a pipe, and rebuild only the dataloader each time. The model loads once. This should bring down the time required  significantly. To keep the current crash isolation, hold the worker across probes that fit, call `empty_cache()` and continue after a recoverable OOM, and respawn only when the process can't recover.
2. **Compute the intercept instead of probing for it.** The fixed cost (weights, gradients, optimizer state) is calculable: params · bytes_per_param · (1 + grad + optimizer_mult), roughly 4× for Adam. Once we know the intercept, one probe at a known-fitting large batch gives the slope. That cuts the probe count from about five to one or two and removes the batch-1 probe, which carries little information and crashes BatchNorm models.
3. **Skip validation during probing.** Validation runs under no_grad and stays below the training-step peak, which is what sets the OOM point. Measuring only the training step halves the work per probe without losing accuracy. 
4. **Trim per-probe overhead.** A probe reads about two batches, so run the dataloader synchronously with num_workers=0 and set cudnn.benchmark=False. The benchmark search spends time picking fast kernels we don't need, since we measure memory, not throughput.
5. **Cache the memory model.** (intercept, slope) is deterministic for a given model architecture, GPU, and input shape. Persist it, and a later run on the same hardware skips probing entirely and solves the line directly. Repeated training of the same architectures then needs no probes.

neuracore/ml/trainers/batch_autotuner.py
kwangneuraco: Hi this is not exactly the min_batch_size and max_batch_size anymore. This is the probe size

neuracore/ml/trainers/batch_autotuner.py
kwangneuraco: Hi this is not the maximum batch size it want to try, it is the size that is used as a probe to estimate the memory usage

neuracore/ml/trainers/batch_autotuner.py
kwangneuraco: Correct i have changed to 2

neuracore/ml/trainers/batch_autotuner.py
kwangneuraco: Which one?

tests/unit/ml/test_batch_autotuner.py
kwangneuraco: both symbols are used only inside batch_autotuner.py itself and the test file — no other production module imports them. That's exactly the case where the underscore is correct.

neuracore/ml/trainers/batch_autotuner.py
kwangneuraco: This is fine because it is only used inside the Autobatch class

kwangneuraco:
> This is a great improvement over the current implementation but I would suggest these following improvements be made.
> 
> 1. **Reuse one worker process.** Right now every probe spawns a new process and reloads the model and optimizer. Instead, keep one worker alive, send it batch sizes over a pipe, and rebuild only the dataloader each time. The model loads once. This should bring down the time required  significantly. To keep the current crash isolation, hold the worker across probes that fit, call `empty_cache()` and continue after a recoverable OOM, and respawn only when the process can't recover.
> 2. **Compute the intercept instead of probing for it.** The fixed cost (weights, gradients, optimizer state) is calculable: params · bytes_per_param · (1 + grad + optimizer_mult), roughly 4× for Adam. Once we know the intercept, one probe at a known-fitting large batch gives the slope. That cuts the probe count from about five to one or two and removes the batch-1 probe, which carries little information and crashes BatchNorm models.
> 3. **Skip validation during probing.** Validation runs under no_grad and stays below the training-step peak, which is what sets the OOM point. Measuring only the training step halves the work per probe without losing accuracy.
> 4. **Trim per-probe overhead.** A probe reads about two batches, so run the dataloader synchronously with num_workers=0 and set cudnn.benchmark=False. The benchmark search spends time picking fast kernels we don't need, since we measure memory, not throughput.
> 5. **Cache the memory model.** (intercept, slope) is deterministic for a given model architecture, GPU, and input shape. Persist it, and a later run on the same hardware skips probing entirely and solves the line directly. Repeated training of the same architectures then needs no probes.

Hi I agree that it can be further improved. But given the time constraint and also the current one is already much better than the old one, let's get it in. For all 5 suggestions I implemented #3 because it makes sense and easy to do

neuracore/ml/trainers/batch_autotuner.py
aditya-neuraco: `elapsed_time` and `start_time`

aditya-neuraco (approved):
LGTM

PR #722, opened 2026-06-19
fix: make the video frame cache crash- and concurrency-safe

kwangneuraco:
The video fetch is easy to crash when the fetch is interrupted (because backend is too busy and memory is full) and resume fetch. This is because the video is partially fetched and is regarded as complete when resume again, but cause errors when decoding.

### Features

- rewrite batch-size tuning with an affine memory model: replace the binary-search autotuner with a memory-model estimator: probe a few batch sizes, fit peak GPU memory as an affine function of batch size, and solve
for the batch that fills the VRAM budget, then apply a safety factor. Details see the doc: https://github.com/NeuracoreAI/neuracore/blob/feat/rewrite_autobatch/docs/batch_size_autotuning.md

### Items
 - https://neuracore.atlassian.net/browse/NCRL-670

### DIFF
--- a/neuracore/core/data/dataset.py
+++ b/neuracore/core/data/dataset.py
@@ -30,6 +30,9 @@ from ..exceptions import AuthenticationError, DatasetError
 from ..utils.http_errors import extract_error_detail
 from ..utils.http_session import thread_local_session
 
+# Cursor pagination is inherently sequential (each page needs the previous
+# page's cursor), so a larger page size would mean fewer serial round-trips
+# The backend clamps `limit` to MAX_PAGE_SIZE (30) in the backend.
 PAGE_SIZE = 30
 SYNC_PROGRESS_POLL_INTERVAL_S = 5.0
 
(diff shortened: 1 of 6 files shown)


neuracore/core/data/synced_dataset.py
```diff
@@ -92,6 +92,12 @@ def _perform_synced_data_prefetch(self, max_prefetch_workers: int) -> None:
         Args:
             max_prefetch_workers: Number of threads to use for prefetching synced data.
         """
+        # Indexing the last recording pages in all metadata up front, so the
+        # threaded prefetch below just reads cache instead of paging concurrently.
+        num_recordings = len(self.dataset)
+        if num_recordings > 0:
+            self.dataset[num_recordings - 1]
```
aditya-neuraco, line 99: I'm confused. What is this line for?

neuracore/core/data/synced_recording.py
```diff
@@ -217,27 +218,51 @@ def _download_video_and_cache_frames_to_disk(
...
 
+    @staticmethod
+    def _decoding_lock_path(frames_dir: Path) -> Path:
+        """Return the decoding lock path for a frames directory.
+
+        The lock is a sibling of ``frames_dir`` rather than a file inside it, so
+        the frames directory can be published atomically (moved into place) once
+        decoding completes.
+        """
+        return frames_dir.parent / f"{frames_dir.name}.recording.lock"
```
aditya-neuraco: Better to inline this logic. I'm not a fan of small helper functions, creates unnecessary indirection.

neuracore/core/data/dataset.py
```diff
@@ -30,7 +30,9 @@
 from ..utils.http_errors import extract_error_detail
 from ..utils.http_session import thread_local_session
 
-PAGE_SIZE = 30
+# Cursor pagination is inherently sequential (each page needs the previous
+# page's cursor), so a larger page size means fewer serial round-trips.
+PAGE_SIZE = 100
```
aditya-neuraco: We should check if out data fetching APIs allow this to be greater than 30.

neuracore/core/data/dataset.py
```diff
@@ -159,14 +161,25 @@ def _initialize_num_recordings(self) -> None:
...
+        Args:
+            required_len: If given, the caller only needs the cache to reach
+                this length. A caller that waited on the lock and finds the
+                cache already long enough returns immediately instead of
+                firing a redundant page fetch (avoids a thundering-herd of
+                over-fetches when many threads block on the lock at once).
         """
         with self._page_lock:
+            # Another thread may have loaded what we need while we waited.
+            if required_len is not None and len(self._recordings_cache) >= required_len:
```
aditya-neuraco, line 181: What is `len(self._recordings_cache) >= required_len:` for? This will never happen right?

aditya-neuraco (approved):
LGTM w/ comments

PR #725, opened 2026-06-22
fix: surface backend error detail on dataset merge/clone

aditya-neuraco:
### Features

 - None

### Bugfixes

 - Show the backend's specific failure reason when merging or cloning a dataset fails, instead of a raw dump of the server response.
 - Fall back to the HTTP status and reason when the server sends no readable error detail, so the user always gets a meaningful message.
 - Add tests covering merge and clone for the error, fallback, and success paths.

### Items

 - [NC-647](https://neuracore.atlassian.net/browse/NC-647)
 - [NC-673](https://neuracore.atlassian.net/browse/NC-673)

### Related PRs

 - None


ypang-neuraco (approved):
LGTM

PR #726, opened 2026-06-24
feat: improve pointcloud logging

ypang-neuraco:
### Features
- Update pointcloud logging from base64 json encoding to storing in binary (metadata still in json)
- Update synced_recording to cache the binary files

### Benchmark against old encoding
| Case | Points | RGB | Legacy encode | New encode | Encode speedup | Legacy decode | New decode | Decode speedup | Legacy frame size | New frame size | Frame size ratio | Legacy disk (20 frames) | New disk (20 frames) | Disk size ratio |
|------|--------|-----|---------------|------------|----------------|---------------|------------|----------------|-------------------|----------------|------------------|-------------------------|----------------------|-----------------|
| small_xyz | 4,096 | No | 0.100 ms | 0.011 ms | 9.2x | 0.076 ms | 0.016 ms | 4.6x | 33.0 KB | 24.9 KB | 1.3x | 660.0 KB | 501.8 KB | 1.3x |
| medium_xyz_rgb | 65,536 | Yes | 2.130 ms | 0.028 ms | 77.0x | 1.549 ms | 0.057 ms | 27.2x | 786.7 KB | 590.1 KB | 1.3x | 15.73 MB | 11.81 MB | 1.3x |
| large_xyz_rgb | 307,200 | Yes | 11.897 ms | 0.292 ms | 40.7x | 7.915 ms | 0.575 ms | 13.8x | 3.69 MB | 2.77 MB | 1.3x | 73.73 MB | 55.31 MB | 1.3x |

### Items
 - [NCRL-639](https://neuracore.atlassian.net/browse/NCRL-639)

### Related PRs
- https://github.com/NeuracoreAI/neuracore_types/pull/95

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -1,5 +1,6 @@
 """Synchronized recording iterator."""
 
+import json
 import logging
 import os
 import shutil
@@ -9,7 +10,7 @@ import tempfile
 import time
 from collections.abc import Callable
 from pathlib import Path
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import numpy as np
 import requests
@@ -18,10 +19,12 @@ from neuracore_types import (
     CameraData,
     CrossEmbodimentUnion,
     DataType,
+    PointCloudData,
     SynchronizationDetails,
 )
 from neuracore_types import SynchronizedEpisode as SynchronizedEpisodeModel
 from neuracore_types import SynchronizedPoint, SynchronizeRecordingRequest
+from neuracore_types.nc_data.point_cloud_data import decode_point_cloud_frame
 from PIL import Image
 
 from neuracore.core.data.cache_manager import CacheManager
@@ -32,6 +35,9 @@ from ..auth import get_auth
 from ..const import API_URL
 
 logger = logging.getLogger(__name__)
+
+POINT_CLOUD_TRACE_BIN_FILE = "trace.bin"
+POINT_CLOUD_TRACE_INDEX_FILE = "trace.json"
 
 if TYPE_CHECKING:
     from neuracore.core.data.dataset import Dataset
@@ -126,6 +132,26 @@ class SynchronizedRecording:
         response.raise_for_status()
         return SynchronizedEpisodeModel.model_validate(response.json())
 
+    def _get_recording_file_url(self, filepath: str) -> str:
+        """Get a signed download URL for a file in this recording.
+
+        Args:
+            filepath: Recording-root-relative path
+                (e.g. ``rgbs/cam1/lossless.mp4``).
+
+        Returns:
+            URL string for downloading the file.
+        """
+        auth = get_auth()
+        session = thread_local_session()
+        response = session.get(
+            f"{API_URL}/org/{self.dataset.org_id}/recording/{self.id}/download_url",
+            params={"filepath": filepath},
+            headers=auth.get_headers(),
+        )
+        response.raise_for_status()
+        return response.json()["url"]
+
     def _get_video_url(self, camera_type: DataType, camera_id: str) -> str:
         """Get streaming URL for a specific camera's video data.
 
@@ -140,9 +166,6 @@ class SynchronizedRecording:
             requests.HTTPError: If every candidate is absent (404), or for any
                 non-404 HTTP error.
         """
-        auth = get_auth()
-        session = thread_local_session()
-
         filename_preference = (
             _DEPTH_VIDEO_FILENAME_PREFERENCE
             if camera_type == DataType.DEPTH_IMAGES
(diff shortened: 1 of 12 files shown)


ypang-neuraco:
@aditya-neuraco I did a benchmark on with and without zstd, the file size only shrunk by ~10% but the encode/decode time increased a lot. I don't think it is worth it at this stage. Maybe we can find a better compression method down the line.

RealSense results on a single frame (307K pts + RGB)
Disk: 2.77 MB → 2.43–2.56 MB (~1.1x at L3, ~1.14x at L6)
Encode: 0.14 ms (v1) → 5.2 ms (zstd L3)
Decode: 0.41 ms (v1) → 2.2 ms (zstd)

aditya-neuraco:
> @aditya-neuraco I did a benchmark on with and without zstd, the file size only shrunk by ~10% but the encode/decode time increased a lot. I don't think it is worth it at this stage. Maybe we can find a better compression method down the line.
> 
> 
> 
> RealSense results on a single frame (307K pts + RGB)
> 
> Disk: 2.77 MB → 2.43–2.56 MB (~1.1x at L3, ~1.14x at L6)
> 
> Encode: 0.14 ms (v1) → 5.2 ms (zstd L3)
> 
> Decode: 0.41 ms (v1) → 2.2 ms (zstd)

Agreed, doesn't make sense to add it.

neuracore/data_daemon/registration_management/registration_manager.py
aditya-neuraco, line 42: We don't import these so let's not import the filename from `neuracore_types.nc_data.point_cloud_data` for consistency in line 18. Move the actual constant here.

neuracore/core/data/synced_recording.py
```diff
@@ -122,24 +128,21 @@ def _get_synced_data(self) -> SynchronizedEpisodeModel:
...
         """
         auth = get_auth()
         session = thread_local_session()
         response = session.get(
             f"{API_URL}/org/{self.dataset.org_id}/recording/{self.id}/download_url",
-            params={"filepath": f"{camera_type.value}/{camera_id}/lossless.mp4"},
+            params={"filepath": filepath},
             headers=auth.get_headers(),
         )
         response.raise_for_status()
```
aditya-neuraco, line 152: - Extract this into a helper `_get_recording_file_url`. Keep the `_get_video_url` with camera_type and camera_id as parameters
- Add a `_get_point_cloud_url` function and use the `_get_recording_file_url` inside it and the `_get_video_url` function.

neuracore/core/streaming/data_stream.py
```diff
@@ -254,6 +255,40 @@ def log(self, data: NCData, *, send_to_daemon: bool = True) -> None:
...
+
+    def __init__(self, data_type_name: str):
+        """Initialize the point cloud data stream.
+
+        Args:
+            data_type_name: Name of the point cloud stream
+        """
+        super().__init__(data_type=DataType.POINT_CLOUDS, stream_name=data_type_name)
+
+    def log(self, data: PointCloudData) -> None:
```
aditya-neuraco, line 347: We should add a metadata parameter here like oher data streams.

neuracore/data_daemon/registration_management/registration_manager.py
```diff
@@ -15,6 +15,7 @@
 
 import aiohttp
 from neuracore_types import DataType
+from neuracore_types.nc_data.point_cloud_data import POINT_CLOUD_TRACE_BIN_FILENAME
```
aditya-neuraco: Other data types don't use this pattern. For now, just define POINT_CLOUD_TRACE_BIN_FILENAME in thsi file around lines 39 to 41, and in a follow up PR, add these constants in all data types.

neuracore/core/streaming/data_stream.py
```diff
@@ -254,6 +255,40 @@ def log(self, data: NCData, *, send_to_daemon: bool = True) -> None:
...
+            data
+        )
+        total_bytes = len(header) + len(metadata_json) + len(points_view)
+        parts: tuple[bytes | memoryview, ...]
+        if rgb_view is not None:
+            total_bytes += len(rgb_view)
+            parts = (header, metadata_json, points_view, rgb_view)
+        else:
+            parts = (header, metadata_json, points_view)
+        self._send_to_daemon_parts(parts, total_bytes=total_bytes)
```
aditya-neuraco, line 364: You can collapse the branching by filtering out None and summing generically:

```suggestion
        self._latest_data = data
        if not self.is_recording():
            return

        header, metadata_json, points_view, rgb_view = encode_point_cloud_frame_parts(data)
        parts = tuple(p for p in (header, metadata_json, points_view, rgb_view) if p is not None)
        total_bytes = sum(len(p) for p in parts)
        self._send_to_daemon_parts(parts, total_bytes=total_bytes)
```

aditya-neuraco (approved):

aditya-neuraco (approved):
LGTM w / comments

neuracore/core/streaming/data_stream.py
ypang-neuraco: This makes it more confusing, the camera data stream should be updated instead. It is passing in the metadata and the frame separately, then combining them in the log function, then dropping it when serialising the metadata.

This is more in line with other json data streams.

PR #727, opened 2026-06-24
chore: add cspell dictionary sort pre-commit hook

aditya-neuraco:
- Add exact dedup and case-insensitive sort for cSpell dictionaries.


CougarTasker (approved):
LGTM

PR #728, opened 2026-06-24
fix: fix gpu not available error in the autobatch integration test

kwangneuraco:
- If has `gpu not available` error, retry up to 5 times, with 30mins interval. If still not working, fail.


ypang-neuraco (approved):
LGTM

PR #732, opened 2026-06-28
fix: slow output syncpoint loading

ypang-neuraco:
### Bugfixes
- Per-output-sensor calls re-sliced and re-projected the same recording window on every load_sample, which duplicated expensive sync-point loads after adding per-type output alignment.
- Load syncpoints once per sample, then slice it per data type so target outputs align to the input step and non-target outputs start at the next step.


stepjam (approved):
LGTM

PR #735, opened 2026-06-30
chore: reduced video fps for long running cases

CougarTasker:
### Bugfixes

 - When I reintroduced this test on our 14 core machines these tests pass with ease, but when test on a vm like github action on 2 and 4 efficency cores it was not possible to encode fast enough.
 - These long running test cases have two parrelel streams of 60fps 1080p fottage that needs to be encoded at real time or the spool will grow uncontrollably . 
 - tested with ffmpeg directly no deamon or contention on all lossy codecs on raw and compressed frames under the best possible conditions and 15fps is the higest safe value.


damonhayhurst-neuraco (approved):
LGTM

PR #738, opened 2026-06-30
chore: sync algorithm config yamls to __init__ defaults and guard drift

aditya-neuraco:
### Bugfixes
- The training configuration files for each algorithm had drifted from the model code. The values a model uses when run through Hydra come from these config files, and several no longer matched the defaults written in the model classes. This change makes every config file agree with its model, so the settings are consistent in both places.
- Corrected values that disagreed: Diffusion Policy (`weight_decay`, `optimizer_betas`), GR00T (`finetune_vision_encoder`, `load_backbone_bf16`, `use_relative_deltas`), Pi0 (`freeze_language_model_only`), Pi05 (`finetune_vision_encoder_and_action_expert`). These adjust how the affected models train.
- Added settings missing from the config files (they previously fell back to code defaults): ACT and Diffusion Policy (`use_resnet_stats`), Diffusion Policy (`lr_scheduler_type`, `lr_scheduler_num_warmup_steps`), GR00T (`use_pretrained_weights`, `use_tiny_vlm`, `max_action_pos_embeddings`).
- Added a test that checks every config file matches its model's defaults, so this drift cannot recur unnoticed.

### DIFF
--- a/neuracore/ml/config/algorithm/act.yaml
+++ b/neuracore/ml/config/algorithm/act.yaml
@@ -21,3 +21,4 @@ algorithm:
 
   # Backbone
   freeze_backbone: false   # Whether to freeze image encoder backbone
+  use_resnet_stats: true   # Use ResNet normalization statistics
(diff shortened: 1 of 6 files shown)


ypang-neuraco (approved):
LGTM

PR #740, opened 2026-07-01
fix: gate hard tests from the legacy daemon

CougarTasker:
### Bugfixes

 - Gates the intergration tests for the python deamon so it runs the same cases as before and only the rust deamon runs the new cases 

#### Runs 

 - A fresh run on this branch with the reduced set: https://github.com/NeuracoreAI/neuracore/actions/runs/28506434303
 

### Items

 - [NCP-1071](https://neuracore.atlassian.net/browse/NCP-1071)


damonhayhurst-neuraco (approved):
LGTM

PR #744, opened 2026-07-01
test: skip connect robot in algorithm eval

ypang-neuraco:
### Bugfixes
- Skip connecting to robot in algorithm eval since we pass in sync points directly


aditya-neuraco (approved):
LGTM

PR #719, opened 2026-06-17
feat: integrate rust daemon behind a flag

CougarTasker:
### Features

 - Dispatches to the Rust daemon binary and native producer only when `NCD_RUST_DAEMON` is truthy; the Python daemon stays the default, so behaviour is unchanged out of the box.
 - Bundles the compiled binary and producer cdylib into the wheel (`setup.py`, `MANIFEST.in`, `build_wheel_artefacts.sh`) and adds the `build-wheels` CI workflow.
 - Adds unit tests for the gating layer.

### Bugfixes

 - None

### Items

 - [NCP-984](https://neuracore.atlassian.net/browse/NCP-984)

### Related PRs

 - Stacked Rust daemon release — review/merge in order:
   1. #718 — workspace + CI + docs
   2. #719 — Python integration + wheel packaging (dark, default-off) **(this PR)**
   3. #720 — integration tests + CI matrix

PR #748, opened 2026-07-02
fix: evict stale joint bindings on stream removal

aditya-neuraco:
### Bugfixes
 - Fixed recordings after the first one in a session being saved as empty. When a recording ended, the robot removed its data streams, but the joint logging cache still pointed at the removed stream objects. The next recording reused those untracked streams, so the stop message told the daemon there was no data to wait for, and the daemon closed the recording before the data arrived. Removing a stream now also clears its cache entries, so the next recording creates and tracks fresh streams.
 - Added a regression test that prunes the streams and checks the next log call registers live ones.

### Items
 - [NCRL-708](https://neuracore.atlassian.net/browse/NCRL-708)

### Related PRs
 - https://github.com/NeuracoreAI/neuracore/pull/719

### DIFF
--- a/neuracore/core/robot.py
+++ b/neuracore/core/robot.py
@@ -243,13 +243,20 @@ class Robot:
         self._data_streams[stream_id] = stream
 
     def _remove_data_stream(self, stream_id: str, stream: DataStream) -> None:
-        """Remove a stream from the robot registry if it still matches."""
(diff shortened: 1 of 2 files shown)


CougarTasker (approved):
LGTM 👍

PR #751, opened 2026-07-03
feat: add live-reloadable video_codec daemon setting

CougarTasker:
### Features

 - Added a `video_codec` setting and a `set_video_encoding_options` method so users can select the RGB video codec. Also settable via the `--video-codec` CLI option, the `NCD_VIDEO_CODEC` environment variable, and the daemon profile. The daemon picks up a change live (config watcher + on-demand refresh) so it applies to the next recording without a restart. This PR only wires up and propagates the setting — encoding output is unchanged; the behaviour that acts on the codec follows in the stacked PR.

### Items

 - [NCP-945](https://neuracore.atlassian.net/browse/NCP-945)

### Related PRs

 - #752 — stacked on top of this PR (the lossy encoding behaviour)

PR #752, opened 2026-07-03
feat: encode lossy-only RGB video for the selected codec

CougarTasker:
### Features

 - Encodes RGB cameras as a single full-resolution lossy video (libx264 CRF 23) and drops the lossless archive when `h264_medium` is selected, saving storage and bandwidth. Depth cameras keep their lossless storage. Adds a playback fallback so recordings captured lossy-only read the lossy video when no lossless archive exists.

### Items

 - [NCP-945](https://neuracore.atlassian.net/browse/NCP-945)

### Related PRs

 - #751 — this PR is stacked on top of it (targets `feat/live-video-codec-config`, not `main`)

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, cast
 
 import numpy as np
+import requests
 import wget
 from neuracore_types import (
     CameraData,
@@ -37,6 +38,9 @@ if TYPE_CHECKING:
 
 MAX_DECODING_ATTEMPTS = 3
 _FFMPEG_AVAILABLE: bool | None = None
+
+_RGB_VIDEO_FILENAME_PREFERENCE = ("lossless.mp4", "lossy.mp4")
+_DEPTH_VIDEO_FILENAME_PREFERENCE = ("lossless.mp4",)
 
 
 class SynchronizedRecording:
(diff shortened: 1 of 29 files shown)


neuracore/core/data/synced_recording.py
```diff
@@ -133,17 +137,38 @@ def _get_video_url(self, camera_type: DataType, camera_id: str) -> str:
...
+                response.raise_for_status()
+            except requests.HTTPError as error:
+                if response.status_code != 404:
+                    raise
+                last_not_found = error
+                continue
+            return response.json()["url"]
+
+        assert last_not_found is not None
+        raise last_not_found
```
aditya-neuraco, line 169: This function is a bit hard to read. It would be better to write it this way.

```suggestion

            URL string for downloading the video file.
        Raises:
            requests.HTTPError: If the API request fails.
            requests.HTTPError: If every candidate is absent (404), or for any
                non-404 HTTP error.
        """
        auth = get_auth()
        session = thread_local_session()
        response = session.get(
            f"{API_URL}/org/{self.dataset.org_id}/recording/{self.id}/download_url",
            params={"filepath": f"{camera_type.value}/{camera_id}/lossless.mp4"},
            headers=auth.get_headers(),
        filename_preference = (
            _DEPTH_VIDEO_FILENAME_PREFERENCE
            if camera_type == DataType.DEPTH_IMAGES
            else _RGB_VIDEO_FILENAME_PREFERENCE
        )
        for video_filename in filename_preference:
            response = session.get(
                f"{API_URL}/org/{self.dataset.org_id}/recording/{self.id}/download_url",
                params={
                    "filepath": f"{camera_type.value}/{camera_id}/{video_filename}"
                },
                headers=auth.get_headers(),
            )
            if response.status_code == 404:
                continue
            response.raise_for_status()
            return response.json()["url"]

        raise requests.HTTPError(
            f"No candidate filename found for recording {self.id} "
            f"(camera {camera_type.value}/{camera_id}); tried: {filename_preference}"
        )
```

neuracore/data_daemon/recording_encoding_disk_manager/encoding/video_trace.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         chunk_size: int = CHUNK_SIZE,
         lossy_name: str = LOSSY_VIDEO_NAME,
         lossless_name: str = LOSSLESS_VIDEO_NAME,
+        codec_option_overrides: dict[str, str] | None = None,
```
aditya-neuraco: Would be nice to make this a dataclass instead of a dict. dict[str, str] can be anything. or even a [TypedDict](https://typing.python.org/en/latest/spec/typeddict.html) would be ice to have as we use strings as keys.

rust/data_daemon/src/cli/coordinators.rs
```diff
@@ -70,6 +70,7 @@ pub(crate) fn spawn_cloud_coordinators(
     recordings_root: Arc<PathBuf>,
     config_path: PathBuf,
     fallback_org_id: Option<String>,
+    config_rx: ConfigRx,
```
aditya-neuraco: Nit: This adds the parameter in-between the existing parameters. It's recommended to add new parameters after the existing parameters. 

But existing callers call at the right position so this doesn't break anything to it's fine as it is.

aditya-neuraco (commented):
Will review other part later.

rust/data_daemon/src/pipeline/trace_actor.rs
```diff
@@ -623,12 +657,15 @@ impl ActorState {
...
-            pending_encodes, ..
+            pending_encodes,
+            codec,
+            ..
         } = &mut self.writer
         else {
             // Should be unreachable — we just allocated the writer above.
             return;
         };
+        let codec = *codec;
```
aditya-neuraco, line 668: Why do we need to dereference it here?

aditya-neuraco (approved):
LGTM w/ comment

rust/data_daemon/src/pipeline/trace_actor.rs
CougarTasker: tldr; technically we don't but for correctness we do; the rust borrow checker requires it. 

So we get the codec by value but we will need a mutable refrence for the pending_encodes. the codec derives copy so when we derefrence it here we are making a clone which will be safe to use in the async encoding task.

aditya-neuraco: Ah, makes sense.

PR #754, opened 2026-07-03
fix: use NEURACORE_ORG_ID as the single org id source for the daemon

damonhayhurst-neuraco:
### Bugfixes

 - Rust daemon was not picking up NEURACORE_ORG_ID


CougarTasker (approved):
LGTM 👍

PR #755, opened 2026-07-03
test: sort data type params for deterministic collection and parallelize ML CI

aditya-neuraco:


tests/unit/ml/algorithms/test_act.py
```diff
@@ -31,14 +31,16 @@
         OrderedSet([data_type]),
         id="".join(w.capitalize() for w in data_type.value.split("_")),
     )
-    for data_type in ACT.get_supported_input_data_types()
+    for data_type in sorted(ACT.get_supported_input_data_types(), key=lambda d: d.value)
```
ypang-neuraco, line 34: Is the order causing the test to fail? If so we should look into the underlying issue

ypang-neuraco (approved):
LGTM \w comment

tests/unit/ml/algorithms/test_act.py
aditya-neuraco: No, since it's `get_supported_input_data_types` is a set, the iteration order will always be random. The problem is parallel workers for parallel tests where detecting different order so pytest was not able to collect the tests.

It's a problem when we iterate on this set across multiple workers. We can use something like [SortedSet](https://grantjenks.com/docs/sortedcontainers/sortedset.html) for the `get_supported_input_data_types` and `get_supported_output_data_types` function later down the line.

PR #757, opened 2026-07-05
ci: install ffmpeg and fix integration tests to use the installed wheel

aditya-neuraco:
### Bugfixes
 - Staging ML and importer integration tests now install ffmpeg (required by the Rust data daemon) in a single consolidated system-packages step, and their test setup imports the installed neuracore package instead of the source checkout, which previously hid the bundled Rust extension and broke daemon startup.

### Related PRs
 - #756


stepjam (approved):
LGTM

PR #759, opened 2026-07-05
ci: launch importer from the installed package in integration tests, run suites in parallel

aditya-neuraco:
### Bugfixes
- Launch the importer from the installed `neuracore` command instead of `python -m`. The `python -m` form put the source folder on the path, which lacks the bundled rust binary, so every recording failed and the staging tests timed out.
- Clone the robot description repository once per session and reuse it, instead of per test.
- Run each dataset suite (lerobot, mcap, rlds) as its own parallel job in the staging and production workflows.

### DIFF
--- a/.github/workflows/integration-importer-production.yaml
+++ b/.github/workflows/integration-importer-production.yaml
@@ -36,6 +36,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.11"]
+        suite: [lerobot, mcap, rlds]
 
     steps:
       - name: Show disk space
@@ -146,7 +147,7 @@ jobs:
           echo "Test harness environment: production"
           echo "Test harness ref: ${{ steps.harness-ref.outputs.harness_ref }}"
           echo "Test harness commit: $(git rev-parse HEAD)"
-          echo "Test harness path: tests/integration/importer/"
+          echo "Test harness path: tests/integration/importer/test_${{ matrix.suite }}_dataset_import.py"
           python - << 'EOF'
           import neuracore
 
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-importer-production.yaml
```diff
@@ -153,7 +154,7 @@ jobs:
           echo "Test harness environment: production"
           echo "Test harness ref: ${{ steps.harness-ref.outputs.harness_ref }}"
           echo "Test harness commit: $(git rev-parse HEAD)"
-          echo "Test harness path: tests/integration/importer/"
+          echo "Test harness path: tests/integration/importer/test_${{ matrix.suite }}_dataset_import.py"
```
stepjam, line 157: I'm not a fan of this. If someone adds a new test, the expectation will be that it automatically runs, but with this change it will not. I also think running the in parallel may overload the machine.

stepjam (approved):
LGTM w/c

.github/workflows/integration-importer-staging.yaml
```diff
@@ -159,7 +160,7 @@ jobs:
           echo "Test harness environment: staging"
           echo "Test harness ref: ${{ steps.harness-ref.outputs.harness_ref }}"
           echo "Test harness commit: $(git rev-parse HEAD)"
-          echo "Test harness path: tests/integration/importer/"
+          echo "Test harness path: tests/integration/importer/test_${{ matrix.suite }}_dataset_import.py"
```
stepjam, line 163: Same here

.github/workflows/integration-importer-production.yaml
aditya-neuraco: Yes, I share the same concern about the test but everytime we add an integration test, we have to add a workflow anyway. So whoever writes a new test, it's their job to update the workflow and verify their test will run.

.github/workflows/integration-importer-staging.yaml
aditya-neuraco: Answered above.

.github/workflows/integration-importer-production.yaml
stepjam: Given the risk, I suggest at the very least having this in a separate PR to discuss with others

PR #761, opened 2026-07-06
fix: remove unused dependency groups in test

damonhayhurst-neuraco:
### Bugfixes

 - Remove mjcf dependency from tests


aditya-neuraco (approved):
LGTM 

Commit message is wrong. It should be fix: remove unused dependency groups in test

PR #767, opened 2026-07-06
fix: process objects not pickled correctly

damonhayhurst-neuraco:
### Bugfixes

 - MacOs uses spawn which pickles process targets by reference rather than ubuntu that does no pickling at all. Hoist nested functions to be straight references so proper pickling can occur and assertions can be completed.


CougarTasker (approved):
LGTM 👍

PR #768, opened 2026-07-06
fix: version error

StevenJacobs61:
### Features

 - Handle updated error message from backend displaying the mismatch between the client and required version

### DIFF
--- a/neuracore/core/cli/select_current_org.py
+++ b/neuracore/core/cli/select_current_org.py
@@ -119,8 +119,13 @@ def run(
         config_manager.config.current_org_id = organization.id
         config_manager.save_config()
 
-    except AuthenticationError:
-        typer.echo("Failed to Authenticate, please try again", err=True)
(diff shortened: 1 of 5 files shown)


neuracore/core/cli/select_current_org.py
```diff
@@ -119,8 +119,13 @@ def run(
         config_manager.config.current_org_id = organization.id
         config_manager.save_config()
 
-    except AuthenticationError:
-        typer.echo("Failed to Authenticate, please try again", err=True)
+    except AuthenticationError as exc:
+        error_message = str(exc)
+        if not error_message.startswith("Neuracore client version mismatch"):
```
CougarTasker: Blocker: This is a fragile contract, the backend would not expect the wording to effect anything. instead it would be more robust to assert that the `required_version` is in the body payload.

CougarTasker (approved):
LGTM \w comment

PR #770, opened 2026-07-07
perf: parallelise PNG compression and scale daemon encode threads

CougarTasker:
### Features

 - None

### Bugfixes

Two independent video-encoding throughput improvements, so the pipeline keeps up with high-detail / multi-camera capture instead of back-pressuring the producer:

 - **Producer — parallel PNG compression.** The single background writer thread used to PNG-compress every camera's frames inline, so high-detail multi-camera capture bottlenecked on one thread (a detailed 720p frame is ~24 ms of miniz work; one thread tops out below two 30 fps cameras). Compression is now offloaded to a small worker pool feeding the writer in submission order — the writer only muxes the NUT chunk. `nc.log_rgb` stays ~1 ms (it still just enqueues the raw frame) and per-stream FIFO ordering is preserved. (`data_daemon_bridge/src/writer.rs`, `nut_writer.rs`)
 - **Daemon — adaptive encode threads.** Each ffmpeg transcode used a fixed 2 libx264 threads per output, tuned for the many-camera 8-context case; with few cameras on a many-core host most cores sat idle. Each encode now gets `cores / active_encodes` threads, so total encoder threads stay ≈ `2 × cores` at any concurrency — it collapses to the original 2 at full load (8-context unchanged) and fills the idle cores otherwise (~2.4× faster per-chunk lossless encode for a few-camera 1080p load). Lossless output is bit-exact regardless of thread count. (`data_daemon/src/pipeline/trace_actor.rs`, `encoding/video_encoder.rs`)

### Items

 - [NCP-1099](https://neuracore.atlassian.net/browse/NCP-1099)

### Related PRs

 - None

### DIFF
--- a/neuracore-dictionary.txt
+++ b/neuracore-dictionary.txt
@@ -298,6 +298,8 @@ Multihead
 multinode
 multirun
 muxer
+muxes
+muxing
 nans
 nbconvert
 nbformat
--- a/rust/data_daemon/src/encoding/video_encoder.rs
+++ b/rust/data_daemon/src/encoding/video_encoder.rs
@@ -48,7 +48,8 @@ pub const DEFAULT_FFMPEG_BINARY: &str = "ffmpeg";
 /// foreground logging threads while ffmpeg still consumes otherwise-idle CPU.
 const ENCODER_NICENESS: libc::c_int = 10;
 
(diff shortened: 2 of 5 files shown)


damonhayhurst-neuraco (approved):
LGTM 👍

PR #771, opened 2026-07-08
fix: endpoint name

StevenJacobs61:
### Features

 - User can no longer create an endpoint with the same name as an existing endpoint.
 - The error message is displayed to the users' terminal upon receiving the relevant 409 error from the backend.


muneebneura (approved):
LGTM 👍

PR #772, opened 2026-07-08
fix: drop stale materialize_artefacts call from integration script

CougarTasker:
### Bugfixes

 - `rust/scripts/run_integration_tests.sh` called `materialize_artefacts` in `main()`, but no such function exists

### Related PRs

 - #754


damonhayhurst-neuraco (approved):
LGTM 👍

PR #773, opened 2026-07-08
fix: retry transient backend errors across storage handlers and metric sync

aditya-neuraco:


damonhayhurst-neuraco:
I find this confusing. 

1. The consts you've added in core imply that we retry on these status calls for all backend api calls.
2. The redirection away from thread_local_session() leads to a misleading read.
3. You've localised some retry mechanisms to a locally named mixin that should really be generic.

My suggestion would be to extend thread_local_session() with retry mechanisms that can be overriden when necessary.

neuracore/core/utils/http_session.py
```diff
@@ -39,25 +39,45 @@
...
+    status_forcelist=_RETRYABLE_STATUS_CODES,
+    raise_on_status=False,  # return the final response rather than raising
+)
+
 _thread_local = threading.local()
 
 
-def thread_local_session() -> requests.Session:
-    """Return a retry-enabled Session cached per thread and process."""
+def thread_local_session(retry_transient: bool = False) -> requests.Session:
```
CougarTasker, line 53: Issue: there is a lot of examples of this being used should this not be the default?

CougarTasker (approved):
LGTM \w comment

neuracore/core/utils/http_session.py
aditya-neuraco: We don't want to retry all client API calls, only some of them in the ML side of things.

PR #778, opened 2026-07-10
fix: surface backend error detail when resuming a training run

aditya-neuraco:


neuracore/core/utils/http_errors.py
```diff
@@ -21,4 +21,4 @@ def extract_error_detail(response: requests.Response) -> str | None:
     if not isinstance(detail_payload, dict):
         return str(detail_payload)
 
-    return detail_payload.get("exception") or detail_payload.get("error")
+    return detail_payload.get("error")
```
ypang-neuraco, line 24: Why did you remove exception? Maybe if you want to prioritise error just change the order. Or consider concatenating them.

neuracore/api/training.py
```diff
@@ -222,10 +223,13 @@ def resume_training_run(job_id: str, additional_epochs: int) -> dict:
             f"{API_URL}/org/{org_id}/training/jobs/{job_id}/resume/{additional_epochs}",
             headers=auth.get_headers(),
         )
-        response.raise_for_status()
```
ypang-neuraco, line 225: Consider making a helper `raise_for_status_with_detail` so it can be reused

ypang-neuraco (changes requested):
LGTM \w comments

ypang-neuraco (approved):
LGTM \w comments

neuracore/core/utils/http_errors.py
aditya-neuraco: I talked to the platform team and reading the exception can leak the backend code, so that's why I removed it.

neuracore/api/training.py
aditya-neuraco: Okay!

PR #780, opened 2026-07-13
fix: surface version mismatch errors through the CLI

aditya-neuraco:
### Features
 - None

### Bugfixes
 - When the installed client is too old for the backend, CLI commands now show a clear error instead of failing with a generic authentication message. The error includes the short summary from the backend, the version you have installed, and the exact command to fix it (pip install --upgrade neuracore).
  ```
 Neuracore client version mismatch. Installed version: 13.3.0. To resolve this, upgrade to the latest client: pip install --upgrade neuracore
 ```
 - Added a dedicated VersionMismatchError exception so the login and select-org commands can tell a version problem apart from a real authentication failure.
 - The login command now checks the client version before asking for credentials, so users find out about an outdated client right away.
 - Removed a duplicate print of the version error to stderr; the message is now shown once through the CLI.

### Items
 - [NCRL-710](https://neuracore.atlassian.net/jira/software/projects/NCRL/boards/67?selectedIssue=NCRL-710)

### Related PRs
 - None

### DIFF
--- a/neuracore/core/auth.py
+++ b/neuracore/core/auth.py
@@ -7,7 +7,6 @@ the application.
 """
 
 import os
-import sys
 from typing import Any
 
 import requests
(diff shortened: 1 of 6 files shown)


stepjam (approved):
LGTM

PR #782, opened 2026-07-13
test: macos fix for pooled workers, surface errors and logging per context

damonhayhurst-neuraco:
### Features

 - Logging per context with namespace scoped logging per context
 - Per test logout for better test scoping

### Bugfixes

 - MacOS uses spawn rather than fork, so init auth per worker
 - Append the repo root to the path so each worker with it's own interpreter can see the code
 - Performance network tests should assert deadlines

### DIFF
--- a/.github/workflows/integration-platform-staging.yaml
+++ b/.github/workflows/integration-platform-staging.yaml
@@ -170,8 +170,14 @@ jobs:
         NEURACORE_API_URL: https://staging.api.neuracore.com/api
         NEURACORE_API_KEY: ${{ secrets.STAGING_SERVICE_API_KEY }}
         NEURACORE_ORG_ID: ${{ vars.STAGING_SERVICE_ORG_ID }}
-      # --import-mode=importlib keeps the repo root off sys.path
-      run: pytest -o log_cli=true --log-cli-level=INFO --import-mode=importlib tests/integration/platform/data_daemon/
+      # --import-mode=importlib keeps the repo root off sys.path. log_cli_format
+      # namespaces live log lines by worker process so interleaved context output
+      # stays attributable.
+      run: >-
+        pytest -o log_cli=true --log-cli-level=INFO
+        -o "log_cli_format=%(levelname)-8s %(processName)-11s %(filename)s:%(lineno)s %(message)s"
+        --import-mode=importlib
+        tests/integration/platform/data_daemon/
 
     - name: Run Consume Stream Integration Test
       env:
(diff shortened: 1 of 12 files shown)


.github/workflows/integration-platform-staging.yaml
```diff
@@ -171,7 +171,7 @@ jobs:
         NEURACORE_API_KEY: ${{ secrets.STAGING_SERVICE_API_KEY }}
         NEURACORE_ORG_ID: ${{ vars.STAGING_SERVICE_ORG_ID }}
       # --import-mode=importlib keeps the repo root off sys.path
-      run: pytest -o log_cli=true --log-cli-level=INFO --import-mode=importlib tests/integration/platform/data_daemon/
+      run: pytest -o log_cli=true --log-cli-level=INFO -o "log_cli_format=%(levelname)-8s %(processName)-11s %(filename)s:%(lineno)s %(message)s" --import-mode=importlib tests/integration/platform/data_daemon/
```
CougarTasker: Question: did you mean to leave this in, its quite long?

tests/integration/platform/data_daemon/conftest.py
```diff
@@ -30,12 +34,32 @@
...
-# cspell:ignore nodeid getfixturevalue
+# cspell:ignore nodeid getfixturevalue modifyitems callspec
+
+# Add the repo root to the path so sub workers on macos can unpickle pool tasks
+# correctly. pytest --import-mode=importlib imports tests files without touching the
+# sys.path. So this is to compensate. Repo root at the front would shadow the installed
+# neuracore wheel so the path is appended rather than prepended.
+_REPO_ROOT = str(Path(__file__).resolve().parents[4])
+if _REPO_ROOT not in sys.path:
+    sys.path.append(_REPO_ROOT)
```
CougarTasker, line 45: Nit: is there not a cleaner way to acheive this from the workflow.yaml?

CougarTasker (approved):
LGTM \w comments

.github/workflows/integration-platform-staging.yaml
damonhayhurst-neuraco: Yes this is our custom logging to show which context is doing what during the test otherwise they are just log lines in simple queue with no namespace scoping.

tests/integration/platform/data_daemon/conftest.py
damonhayhurst-neuraco: With the way the PYTHONPATH is resolved at runtime, the repo root would always shadow the wheel.

PR #783, opened 2026-07-13
feat: enforce monotonic log timestamps

CougarTasker:
### Features

 - The `log_*` functions now enforce that timestamps are strictly increasing per sensor stream within a recording. 

### Items

 - [NCP-1121](https://neuracore.atlassian.net/browse/NCP-1121)

### DIFF
--- a/neuracore/core/streaming/data_stream.py
+++ b/neuracore/core/streaming/data_stream.py
@@ -73,6 +73,7 @@ class DataStream(ABC):
         self._stream_name = stream_name
         self._producer_channel: ProducerChannel | None = None
         self._use_data_bridge = is_rust_daemon_enabled()
+        self._last_logged_timestamp: float | None = None
 
     @property
     def data_type(self) -> DataType:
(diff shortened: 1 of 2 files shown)


stepjam (changes requested):
Unfortunately this is not the right fix. If we are going to enforce this, it needs to be on a per-stream basis. It is valid to have e.g. joint positions be logged at timestamps 0.1 and then a message come through for a RGB frame at timestamp 0.01.

CougarTasker:
> Unfortunately this is not the right fix. If we are going to enforce this, it needs to be on a per-stream basis. It is valid to have e.g. joint positions be logged at timestamps 0.1 and then a message come through for a RGB frame at timestamp 0.01.

This pr is on a per-stream basis. `_last_logged_timestamp` has been added on a per `DataStream` . There is also a test `test_monotonic_check_is_per_stream` that enforces this 


https://github.com/NeuracoreAI/neuracore/blob/f49f50fbaa38acf418d484c3b324c3fb3fc529db/neuracore/core/streaming/data_stream.py#L76
https://github.com/NeuracoreAI/neuracore/blob/f49f50fbaa38acf418d484c3b324c3fb3fc529db/tests/unit/core/test_data_stream.py#L313

stepjam (approved):
Ah I must have not looked at the file name when I first reviewed this. 

LGTM!

aditya-neuraco (approved):
LGTM

PR #787, opened 2026-07-14
feat: toggle recording reaper for data daemon

damonhayhurst-neuraco:
### Features

 - Recording Reaper can be toggled using a flag in the profile


CougarTasker (approved):
LGTM 👍

PR #788, opened 2026-07-15
fix: update google colab to v13.4.0

aditya-neuraco:
- Update google colab
- Add version the model is trained on in the pretrained weights filename.


sdas-neuraco (approved):
LGTM

PR #789, opened 2026-07-15
feat: add check if dataset size is larger then VM size

sdas-neuraco:
### Features

 - Adds a check to see if the dataset size is larger then the VM size. If it is larger then it throws an error.

### DIFF
--- a/neuracore/api/training.py
+++ b/neuracore/api/training.py
@@ -155,6 +155,16 @@ def start_training_run(
             algorithm_jsons=algorithm_jsons,
         )
     )
+
+    # Validate that the machine size is large enough for the dataset.
+    dataset_size_gb = dataset.size_bytes / 1_000_000_000
+    if disk_size_gb <= dataset_size_gb:
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


neuracore/api/training.py
```diff
@@ -155,6 +155,16 @@ def start_training_run(
...
         )
     )
+
+    # Validate that the machine size is large enough for the dataset.
+    dataset_size_gb = dataset.size_bytes / 1_000_000_000
+    if disk_size_gb <= dataset_size_gb:
+        raise ValueError(
+            f"Dataset {dataset_name} is {dataset_size_gb:.2f} GB, "
+            f"but selected VM disk is {disk_size_gb} GB. "
+            "Please increase disk_size_gb."
```
aditya-neuraco, line 165: This is a user facing error, so we should say "Please run the training with a larger disk space" or something along these lines.

aditya-neuraco (approved):
LGTM with comment.

neuracore/api/training.py
sdas-neuraco: the parameter in the function is disk_size_gb. For the front end we will have to have another check in the backend.

PR #790, opened 2026-07-15
ci: test harness referenced from shared workflow for ml

damonhayhurst-neuraco:
### Features

 - test harness referenced from shared workflow
 - prod workflows use shared workflow setup from pypi
- remove p2p from daemon tests


CougarTasker (approved):
LGTM

PR #791, opened 2026-07-15
fix: install bigym via pip instead of git clone in Colab notebook

aditya-neuraco:
- Fix import errors on colab.


sdas-neuraco (approved):
LGMT

PR #793, opened 2026-07-17
fix: guard empty train/val split

ypang-neuraco:
### Bugfixes
- train.py computes train_size and val_size from len(dataset) and validation_split, then builds loaders unconditionally. Small datasets can produce train_size == 0 or val_size == 0, causing confusing late failures or bogus metrics. Add an explicit validation error before random_split

### Items
 - [NCRL-705](https://neuracore.atlassian.net/browse/NCRL-705)

### DIFF
--- a/neuracore/ml/train.py
+++ b/neuracore/ml/train.py
@@ -397,6 +397,24 @@ def run_training(
         train_split = 1 - cfg.validation_split
         train_size = int(train_split * dataset_size)
         val_size = dataset_size - train_size
+        if train_size == 0 and val_size == 0:
+            raise ValueError(
+                "The training and validation sets are both empty. "
+                "Try adding more recordings to the dataset, "
+                "or changing the validation split."
+            )
+        if train_size == 0:
+            raise ValueError(
+                "The training set is empty. "
+                "Try adding more recordings to the dataset, "
+                "or changing the validation split."
+            )
+        if val_size == 0:
+            raise ValueError(
+                "The validation set is empty. "
+                "Try adding more recordings to the dataset, "
+                "or changing the validation split."
+            )
 
         # Use random split with fixed seed for deterministic behavior
         generator = torch.Generator().manual_seed(cfg.seed)
--- a/tests/unit/ml/test_train.py
+++ b/tests/unit/ml/test_train.py
@@ -1958,6 +1958,44 @@ class TestRunTraining:
             20,
         ]  # train_size=80, val_size=20 for 100 samples with 0.2 split
 
+    @pytest.mark.parametrize(
+        ("dataset_size", "validation_split", "expected_message"),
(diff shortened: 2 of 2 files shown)


tests/unit/ml/test_train.py
```diff
@@ -1958,6 +1958,44 @@ def mock_random_split_side_effect(dataset, lengths, generator=None):
...
         ]  # train_size=80, val_size=20 for 100 samples with 0.2 split
 
+    @pytest.mark.parametrize(
+        ("dataset_size", "validation_split", "expected_message"),
+        [
+            (0, 0.2, "training and validation sets are both empty"),
+            (1, 0.2, "training set is empty"),
+            (3, 0.0, "validation set is empty"),
+        ],
+    )
```
aditya-neuraco, line 1968: The test its can be a bit better. Second one doesn't tell me the training set is emty explicitly.

neuracore/ml/train.py
```diff
@@ -397,6 +397,18 @@ def run_training(
         train_split = 1 - cfg.validation_split
         train_size = int(train_split * dataset_size)
         val_size = dataset_size - train_size
+        empty_sets = ""
+        if train_size == 0 and val_size == 0:
+            empty_sets = "The training and validation sets are both empty."
```
aditya-neuraco: The variable name is misleading. It's a error string. It can be a tuple: `("training", "validation")`

neuracore/ml/train.py
```diff
@@ -397,6 +397,18 @@ def run_training(
...
+        elif train_size == 0:
+            empty_sets = "The training set is empty."
+        elif val_size == 0:
+            empty_sets = "The validation set is empty."
+        if empty_sets:
+            raise ValueError(
+                f"{empty_sets} Try adding more recordings to the dataset, "
+                "or changing the validation split."
+            )
```
aditya-neuraco: More readable to just inline the erros like this.

```suggestion
if train_size == 0 and val_size == 0:
    raise ValueError("The training and validation sets are both empty. Try adding more recordings to the dataset, or changing the validation split.")
if train_size == 0:
    raise ValueError("The training set is empty. Try adding more recordings to the dataset, or changing the validation split.")
if val_size == 0:
    raise ValueError("The validation set is empty. Try adding more recordings to the dataset, or changing the validation split.")
```

aditya-neuraco (approved):
LGTM w/ comments

neuracore/ml/train.py
sdas-neuraco: agree

tests/unit/ml/test_train.py
ypang-neuraco: I don't want to mock too much of the logic

PR #798, opened 2026-07-21
feat: add large dataset recording deletion test

sdas-neuraco:
## What changed

This PR adds an integration test for deleting recordings from a large cloned dataset.

The test:

1. Loads the shared Freiburg dataset.
2. Saves its recording IDs and metadata for later checks.
3. Checks that the dataset is large enough to use more than one API page.
4. Clones the dataset with a unique name.
5. Checks that the clone matches the source.
6. Deletes recordings at the start, end, and around a page boundary.
7. Fetches the clone again until the deletes appear in the API.
8. Checks that only the chosen recordings were deleted.
9. Deletes all remaining recordings from the clone.
10. Checks that the clone is empty.
11. Merges the empty clone with the source and checks the result.
12. Fetches the source again to make sure it was not changed.
13. Deletes the temporary clone and merged dataset, even if the test fails.

### DIFF
--- a/tests/integration/platform/test_large_dataset_recording_deletion.py
+++ b/tests/integration/platform/test_large_dataset_recording_deletion.py
@@ -1,0 +1,233 @@
+"""End-to-end coverage for deleting recordings from a large cloned dataset.
+
+This test deliberately exercises the real Neuracore API. It requires valid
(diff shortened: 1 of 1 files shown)


sdas-neuraco:
Replaced by #799 so the branch follows the required `feat/...` naming rule.

PR #799, opened 2026-07-21
feat: add large dataset recording deletion test

sdas-neuraco:
## What changed

This PR adds an end-to-end integration test for deleting every recording from a large cloned dataset.

The test:

1. Loads the shared Freiburg dataset and snapshots its recording metadata.
2. Verifies the source spans more than one API page.
3. Clones the dataset with a unique name and checks that the clone matches the source.
4. Deletes every recording from the clone.
5. Polls fresh `Dataset` objects until the backend reports that the clone is empty, avoiding stale client caches and eventual-consistency races.
6. Merges the empty clone with the source and checks that the result matches the source.
7. Refetches the source to verify that cloning, deletion, and merging did not mutate it.
8. Cleans up temporary datasets by object or UUID-based name, including when creation succeeds but the client call fails before assignment.

### DIFF
--- a/tests/integration/platform/test_large_dataset_recording_deletion.py
+++ b/tests/integration/platform/test_large_dataset_recording_deletion.py
@@ -1,0 +1,233 @@
+"""End-to-end coverage for deleting recordings from a large cloned dataset.
+
+This test deliberately exercises the real Neuracore API. It requires valid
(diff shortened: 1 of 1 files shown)


damonhayhurst-neuraco:
Merge/Clone fixes for test per #908

PR #800, opened 2026-07-22
fix: recording loss from stop/start inversion

CougarTasker:
### Bugfixes

 - Fixed a recording-loss race on the ipc lifecycle path: the bridge's `stop_recording` published `StopRecording` only **after** blocking on the video-writer flush barrier, so under a slow flush (~19 ms observed in CI traces) the stop reached the wire **after** the next recording's `StartRecording`.
   - Bridge: `stop_recording` now publishes `StopRecording` **before** the flush barrier,
   - Dispatcher `handle_start` persists the retired window's stop when a start finds a live window, and `handle_stop` no longer lets a stale stop
   - New dispatcher test `inverted_stop_after_next_start_preserves_both_recordings` replays the exact inverted sequence from the CI trace capture

### Items

 - [NCP-1160](https://neuracore.atlassian.net/browse/NCP-1160)

### Validation (trace-level CI runs stacked on this commit)

Retry-loop validation on `ci/cougar-tracing-run` (runs 29916355272 / 29916353564, cancelled mid-loop after conclusive streaks): **macOS 6/6, ubuntu 4/4, importer-rlds 18/18 iterations passed — zero orphan drops (`dropped datum outside any recording window`) on every leg.** Pre-fix, the same macOS harness failed 8 of 9 iterations.

 - **Root fix proven:** macOS stop publish→receive latency median fell **3.31 ms → 0.95 ms** (now start-like) — the direct signature of publishing `StopRecording` before the flush barrier; 80/80 same-source boundaries were received stop-before-next-start.
 - **Defense layer proven necessary:** ubuntu logged zero inversion warns (the reorder alone sufficed), but macOS logged 2 residual dispatcher-side inversions under the 8-context per-thread burst case (exactly its two highest-latency stops, 38.9 ms / 20.0 ms) — both absorbed losslessly by the new `handle_start`/`handle_stop` guards (`start arrived with prior recording still live` + `stop arrived after a later recording started; refining`). With per-thread producer channels, stop and next-start can ride different ports, so the dispatcher guards are load-bearing, not just belt-and-braces.
 - **DB integrity:** every completed iteration shows full trace counts (6×32 reduced case, 16×26 burst case), zero unstopped/cancelled/zero-trace recordings.
 - Importer leg: 540 expected-trace-count reports, all full (no short counts) — 18 clean shared imports where the (separate, backend-side) save-promotion race previously struck ~1-in-5–7 iterations; suggestive but not proof that earlier stop publication narrows that race window.

### DIFF
--- a/rust/data_daemon/src/pipeline/dispatcher.rs
+++ b/rust/data_daemon/src/pipeline/dispatcher.rs
@@ -544,10 +544,17 @@ impl Dispatcher {
         // A well-behaved producer stops before starting; if a live window is
         // somehow still open, retire it to `closing` bounded at the new start's
         // publish time so it stops catching data published after this point.
+        // This also happens under an inverted start/stop pair (a slow stop
+        // reaching the daemon after the next start), so persist the retired
+        // recording's stop here — using the new start's publish time as the
+        // exclusive upper bound of its membership range — rather than leaving
+        // its row open. A later stolen stop refines the exact boundary.
+        let mut retired: Option<i64> = None;
         if let Some(mut previous) = entry.live.take() {
             if previous.stopped_at_ns.is_none() {
                 previous.stopped_at_ns = Some(publish_timestamp_ns);
                 previous.stop_recv_at = Some(recv_at);
+                retired = Some(previous.recording_index);
             }
             entry.closing.push(previous);
         }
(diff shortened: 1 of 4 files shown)


StevenJacobs61 (approved):
LGTM, am I right in thinking this is a patch which accommodates the Damon receiving start/stop out of order, as opposed to ensuring they arrive in the correct order? If is it might be useful to know why they arrive out of order in the first place

PR #801, opened 2026-07-22
feat: add dataset recording deletion

sdas-neuraco:
## Summary

- add `Dataset.delete_recording()` with separate `recording_name` and `recording_id` arguments
- support direct deletion by ID and name lookup with duplicate-name checks
- clear recording and robot caches after deletion and refresh dataset metadata
- add unit coverage for successful deletion and failure cases

## Why

This section of code has been hacked together in multiple places, including integration tests, separate scripts, and tools. Since deleting recordings is a common operation outside the SDK, this PR formally adds the function to `Dataset`.

Please Check in detail for the Page Lock as im not 100% sure on this.

### DIFF
--- a/neuracore/core/data/dataset.py
+++ b/neuracore/core/data/dataset.py
@@ -461,6 +461,82 @@ class Dataset:
         self.__dict__.clear()
         self.__class__ = _DeletedDataset
 
+    def delete_recording(
+        self,
+        recording_name: str | None = None,
+        recording_id: str | None = None,
+    ) -> None:
+        """Delete a recording from this dataset by its name or ID.
+
+        Args:
+            recording_name: The recording's human-readable name.
+            recording_id: The recording's unique ID.
+
+        Raises:
+            ValueError: If neither or both identifiers are supplied.
+            DatasetError: If no recording has the supplied identifier, or if the name
+                matches more than one recording.
+            requests.HTTPError: If the API request fails.
+        """
+        # The user must provide either a name or an ID, but not both.
+        if (recording_name is None) == (recording_id is None):
+            raise ValueError(
+                "Exactly one of recording_name or recording_id must be provided."
+            )
+
+        if recording_id is not None:
+            # An ID can be sent to the API without loading the dataset first.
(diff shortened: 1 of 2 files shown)


neuracore/core/data/dataset.py
```diff
@@ -461,6 +461,82 @@ def delete(self) -> None:
...
+            if not matches:
+                raise DatasetError(
+                    f"Recording {recording_name!r} not found in dataset {self.name!r}."
+                )
+            # Names can be the same, so use an ID when more than one name matches.
+            if len(matches) > 1:
+                raise DatasetError(
+                    f"Multiple recordings named {recording_name!r} exist in dataset "
+                    f"{self.name!r}; delete the recording by ID instead."
+                )
```
CougarTasker: Issue: for large datasets this search will be slow, I reccomend doing this on the backend instead if it is required

tests/unit/core/data/test_dataset.py
```diff
@@ -806,6 +806,120 @@ def test_len(self, dataset_dict, recordings_list):
...
+        )
+
+        with caplog.at_level(logging.WARNING):
+            dataset.delete_recording(recording_id="rec1")
+
+        assert "failed to refresh dataset" in caplog.text
+        assert dataset._recordings_cache == []
+        assert dataset._num_recordings is None
+
+    def test_delete_recording_rejects_ambiguous_name(
```
CougarTasker, line 923: Issue: missing test cases, deleting recordings and itterating over the dataset object

CougarTasker (approved):
LGTM \w comments

sdas-neuraco:
Backend support for server-side recording-name lookup is in NeuracoreAI/neuracore_backend#586. This PR now requests at most two exact matches from that endpoint, avoiding a full dataset download while retaining duplicate-name detection.

neuracore/core/data/dataset.py
sdas-neuraco: adressed created backend PR: https://github.com/NeuracoreAI/neuracore_backend/pull/586

PR #802, opened 2026-07-22
fix: missing recording index test failure

StevenJacobs61:
### Bugfixes

 - Do not manually delete the .wal and .shm files, these are owned by sqlite3 and we are letting it manage the cleanup and tear down of them which it does well, because un-linking them whilst any connection still knows about them, such as the pytest process in this case, can produce the cross test state problems we see here. 
 - Disabled rust for shared transport layer tests as they are not applicable to the rust daemon.
 - Explicitly closed the database connection as opposed to just removing files in between tests, as leaving the connection open means that sqlite can still hold the file descriptors, locks, references to wal & shm files etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,9 @@ examples/test_streaming/
 
 .data_daemon_test_state/
 
+# Local Neuracore login helper (contains credentials)
+login_neuracore.sh
+
 # Bundled rust artefacts built into the package tree by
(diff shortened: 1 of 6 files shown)


CougarTasker (approved):
LGTM 👍

CougarTasker:
Questsion: Does closing the database eventually remove it from disk?

PR #804, opened 2026-07-22
fix: importer robot instance collisions

CougarTasker:
### Bugfixes

 - Two importer runs recording to the same robot at the same time no longer merge their recordings. Importer workers always used robot instances `0..max_workers-1`, and the platform joins concurrent `/recording/start` calls for the same (robot, instance) into a single recording. Each importer invocation now draws a random per-run instance base (1,000–999,999) and workers use `base + worker_id`, so any two runs use disjoint instances.
   - A merged recording could never finalize: each run PUTs its own expected-trace-count while the backend holds the sum of both runs' trace registrations (observed `409 "34 traces already registered"` = 17+17), so the recording stays unlistable in every dataset and each run's dataset comes up short.

### Items

 - [NCP-1163](https://neuracore.atlassian.net/browse/NCP-1163)

### DIFF
--- a/neuracore/importer/core/base.py
+++ b/neuracore/importer/core/base.py
@@ -110,6 +110,7 @@ class NeuracoreDatasetImporter(ABC):
     """Importer workflow that manages workers and Neuracore session setup."""
 
     DISK_CHECK_INTERVAL_SECS: float = 10.0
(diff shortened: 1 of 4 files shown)


StevenJacobs61 (approved):
LGTM

PR #805, opened 2026-07-22
fix: recording reaper completion race

CougarTasker:
### Bugfixes

 - The recording reaper can no longer destroy a recording whose `UPLOAD_COMPLETE` status is still queued in the status coordinator. 
 - A reconcile pass in the status coordinator re-drives completions whose acknowledgement never landed, so the recording is not stuck.

### Evidence
Without the guard, reclaim raced the status coordinator's 200 ms completion batch: in the observed incident the reaper deleted the recording row ~2 ms before the batch flush read it, `flush_batch` treated the missing row as not-created-yet and deferred forever, and the backend never received the final trace completion — the recording stayed 16/17 complete server-side and CI hung draining it (recording `bbe50b98`, staging, 2026-07-21 23:00Z; request logs confirmed the PUT never arrived).

Tests (fail-before verified for the gate): `sweep_spares_a_recording_with_an_unacknowledged_completion` (the incident shape), the extended reclaim-filter store test, `flush_batch_stamps_acked_completions` (stamp exactness), `claim_unreported_completions_bounds_attempts_and_unblocks_reclaim` (grace window, attempt bounding, gate opens on exhaustion), and `reconcile_re_sends_dropped_completion_until_reclaimable` (dropped batch → re-send → stamp → reclaim).

### Items

 - [NCP-1162](https://neuracore.atlassian.net/browse/NCP-1162)

### Related PRs

 - Stacked on #804

### DIFF
--- a/rust/data_daemon/src/cloud/watchers/recording_reaper.rs
+++ b/rust/data_daemon/src/cloud/watchers/recording_reaper.rs
@@ -169,6 +169,17 @@ mod tests {
     /// Drive a stopped recording with one fully-uploaded trace through every
     /// notify + progress gate so the server-side reclaim filter reports it.
     async fn seed_reclaimable_stopped(store: &SqliteStateStore, instance: i64) -> i64 {
+        let index = seed_stopped_unreported(store, instance).await;
+        store
+            .mark_traces_completion_reported(&[format!("t-{instance}")])
+            .await
+            .unwrap();
+        index
+    }
+
+    /// Same as [`seed_reclaimable_stopped`] but the trace's `UPLOAD_COMPLETE`
+    /// has not yet been acknowledged by the backend.
+    async fn seed_stopped_unreported(store: &SqliteStateStore, instance: i64) -> i64 {
         let index = store
             .create_recording(new_recording(instance))
             .await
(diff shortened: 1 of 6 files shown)


StevenJacobs61 (approved):
LGTM

rust/data_daemon/src/cloud/watchers/recording_reaper.rs
```diff
@@ -208,6 +219,36 @@ mod tests {
         index
     }
 
+    /// The `bbe50b98` incident shape: everything settled locally, but the
```
damonhayhurst-neuraco, line 222: what is bbe50b98?

CougarTasker: This is the recording that was caught in the logs of a CI job that had this problem where the recording was removed just before the last batch of traces were completed from the backend.

This is the regression test that recreates the problem and fails before the fix and passes after it.

damonhayhurst-neuraco: Ok cool. I think I prefer that as a comment rather than whats there.

favour-neuraco (approved):
LGTM

PR #806, opened 2026-07-23
fix: lossy video pts collapse

CougarTasker:
### Bugfixes

Fixed the lossy video issue behind the dashboard "Video missing logged frames". This happened because ffmpeg segment's encoder time base and mp4 track timescale from a per-chunk *guessed* frame rate; the guess is unstable across the chunks of one recording whole chunks get crammed onto consecutive single ticks with backwards decoded PTS.
 
  - `encode_chunk` now pins every output to the NUT's microsecond clock so segment PTS equal the capture timestamps and the concat is uniform: no cramming, no backwards PTS, capture jitter preserved. Verified A/B: the previously-corrupting mixed-chunk concat now decodes monotonic.
  - Hardened the producer's PTS derivation as defence in depth:  Clean input is byte-identical (regression-tested).

> No data was lost due to this bug so i am migrating the existing recordings that have been uploaded

### Items

 - [NCP-1164](https://neuracore.atlassian.net/browse/NCP-1164)

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -427,6 +427,7 @@ dependencies = [
  "clap",
  "crc32c",
  "dashmap",
+ "data_daemon_bridge",
  "data_daemon_shared",
  "dirs",
  "iceoryx2",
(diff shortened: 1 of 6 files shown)


damonhayhurst-neuraco (approved):
LGTM

favour-neuraco (approved):
lgtm

PR #807, opened 2026-07-23
fix: add ceiling for variable duration

damonhayhurst-neuraco:
### Bugfixes

 - Duration mode variable would surpass the 300s limit if the recording was 300s long


CougarTasker (approved):
LGTM 👍

PR #809, opened 2026-07-23
fix: fix resume delete dataset assert

ypang-neuraco:
### Bugfixes
- Allow identifying deleted dataset by ID instead of just name


aditya-neuraco (approved):
LGTM

PR #810, opened 2026-07-23
fix: frame buffer for improved memory footprint of data daemon tests

damonhayhurst-neuraco:
### Bugfixes

 - Tests were encoding frames on the fly causing an 6.2MB overhead in memory on each call. Encode frames once and provide a buffer.


CougarTasker (approved):
LGTM 👍

PR #813, opened 2026-07-24
fix: disable timer coalescing for mac

damonhayhurst-neuraco:
### Bugfixes

 - Mac scheduler batches software timers that are milliseconds apart. Disable this so that time.sleep values are properly respected for stochastic test cases.


stepjam (approved):
LGTM

PR #815, opened 2026-07-27
fix: set thread policy for macos performance

damonhayhurst-neuraco:
### Bugfixes

 - With MacOS we need to tune the thread policy as the scheduler sometime neglects real time responsiveness


favour-neuraco (approved):
LGTM! great work

PR #817, opened 2026-07-28
fix: retry transient recording completion checks

favour-neuraco:
### Bugfixes

- Retry transient connection errors and read timeouts while polling recording upload completion, preventing `stop_recording(wait=True)` from failing during temporary backend capacity or connection disruption.
- Validated with the previously affected large staging integration case: transient failures were retried successfully and the test passed.

Successful staging run: https://github.com/NeuracoreAI/neuracore/actions/runs/30354571394/job/90260030430

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -366,7 +366,11 @@ def stop_recording(
         if backend_utils.is_recording_upload_complete(recording_id):
             return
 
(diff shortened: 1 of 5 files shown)


damonhayhurst-neuraco:
We already have a way to retry transient errors. Take a look at the definition `thread_local_session()`.

damonhayhurst-neuraco (approved):
LGTM

PR #820, opened 2026-07-28
fix: retry transient recording completion checks (#817)

StevenJacobs61:
### Features

 - Updated relevant workflow files to use the rust daemon by default.


StevenJacobs61:
Closed with merge mistake

PR #821, opened 2026-07-28
fix: workflow use rust daemon

StevenJacobs61:
### Features

 - Updated the CI workflows that use the data daemon to use the rust data daemon by default.


aditya-neuraco (approved):
LGTM

PR #823, opened 2026-07-29
feat: rust daemon on by default

damonhayhurst-neuraco:
### Features

 - Turn the rust daemon on by default - fallback to the python daemon if the value is false or the binary is not available
 - Separate os control tests specific to legacy python daemon

### Bug Fixes
 - Prod CI was using incorrect pytest collection methodology for running tests, `import-mode=importlib`. Bring it up to parity with staging.


StevenJacobs61 (approved):
LGTM

PR #824, opened 2026-07-29
fix: multi producer processes

StevenJacobs61:
### Bugfixes

 - Bind robot Id and instance from the robot receiving an SSE recording started event. This enables multiple producers in different process to be able to log to the same daemon independently.


favour-neuraco (approved):
LGTM

PR #825, opened 2026-07-29
fix: preserve recording duration during dataset imports

favour-neuraco:
### Bugfixes
 - Preserve the correct duration of imported MCAP, LeRobot, and RLDS/TFDS recordings by passing explicit capture timestamps to start_recording() and stop_recording(). Fixed-frequency datasets use their existing step timeline, while MCAP timestamps are rebased onto a single Unix-time anchor. This ensures duration reflects the source episode data rather than how quickly the importer executes.

### Items
 - [NC-982](https://neuracore.atlassian.net/browse/NCP-982)


sdas-neuraco (approved):
LGTM

PR #826, opened 2026-07-30
ci: drop mujoco from daemon prod test

damonhayhurst-neuraco:
### Bugfixes

 - Prod CI job was incorrectly using Mujoco. Python 3.14 is incompatible with Mujoco anyway.


favour-neuraco (approved):
LGTM

PR #827, opened 2026-07-30
ci: add axis filtering for daemon tests so we can run subset of environments

damonhayhurst-neuraco:
### Features

 - Filter workflows so we can run a subset of environments on manual dispatch of data daemon tests


StevenJacobs61 (approved):
LGTM

PR #828, opened 2026-07-30
fix: reduce request timeout

StevenJacobs61:
### Bugfixes

 - Reduced the timeout to 3s as a 10s timeout was excessive and caused the test to fail a less than 10s stop recording assertion.


favour-neuraco (approved):
LGTM

PR #829, opened 2026-07-30
fix: auth test error

StevenJacobs61:
### Features

 - None

### Bugfixes

- Race condition in the tests between logout and teardown exposed hanging peer to peer connections which could send a heartbeat and cause an auth error. Logout who now owns stream Orchastrater tear down.

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -104,7 +104,15 @@ def logout() -> None:
     Logs out from the Neuracore server and resets all global state including
     active robots, recording IDs, dataset IDs, and version validation status.
     """
-    get_auth().logout()
+    # Streaming tasks use the shared Auth instance at execution time. Stop and
+    # drain them before clearing its token so an in-flight heartbeat or track
+    # submission cannot become an unauthenticated background exception.
+    try:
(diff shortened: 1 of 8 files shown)


damonhayhurst-neuraco:
What tests is this fixing? We don't use P2P in our data daemon tests on the CI.

StevenJacobs61:
> What tests is this fixing? We don't use P2P in our data daemon tests on the CI.

In connect_robot a StreamManagerOrchastrator is created and its loop started. In logout, previously the shared access token was cleared whilst the StreamManagerOrchastrator loop is still alive, leaving a race condition in teardown where logout is called and the SMO still sends a heartbeat or reconnect calling Auth.get_headers triggering an auth error. So now logout also tears down the SMO.

damonhayhurst-neuraco:
> In connect_robot a StreamManagerOrchastrator is created and its loop started. In logout, previously the shared access token was cleared whilst the StreamManagerOrchastrator loop is still alive, leaving a race condition in teardown where logout is called and the SMO still sends a heartbeat or reconnect calling Auth.get_headers triggering an auth error. So now logout also tears down the SMO.

What tests is this fixing though?

PR #830, opened 2026-07-30
fix: multi process retries

StevenJacobs61:
### Features

 - None

### Bugfixes

- Added retires to robot init requests as 500 error from one failed node can cause the recording to fail under high congestion. 
- Added lock around connect robot in multi process tests.

### DIFF
--- a/tests/integration/platform/data_daemon/behavioural_correctness/test_multi_process_producers.py
+++ b/tests/integration/platform/data_daemon/behavioural_correctness/test_multi_process_producers.py
@@ -65,6 +65,7 @@ def _same_source_producer(
     producer_index: int,
     robot_name: str,
     dataset_name: str,
+    robot_connection_lock: Any,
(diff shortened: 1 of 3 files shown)


tests/integration/platform/data_daemon/behavioural_correctness/test_multi_process_producers.py
```diff
@@ -79,7 +80,9 @@ def _same_source_producer(
     try:
         ensure_login()
         nc.get_dataset(dataset_name)
-        robot = nc.connect_robot(robot_name, instance=0, overwrite=False)
+        # Robot registration is setup rather than the multiprocess behavior under test.
```
damonhayhurst-neuraco, line 83: Suggestion: This comment could be improved to explain why we're using the lock

damonhayhurst-neuraco (approved):
LGTM

PR #831, opened 2026-07-30
fix: reduce daemon test overhead

StevenJacobs61:
### Features

 - Reduced overhead of data daemon test cases and removed unused network integrity tests


stepjam:
Closed following discussion on Friday. Reducing number of recordings is not how we should be fixing the tests

PR #832, opened 2026-07-30
ci: increase max parallelisation of per wheel testing for staging daemon tests

damonhayhurst-neuraco:
### Features

 - Increase max parallelisation of per wheel testing for data daemon staging from 2 to 4. 

### Why
 - Data Daemon tests are taking too long
- We need ensure our staging can handle suitable workload

### How
- Increase the rate of tests hitting our staging backend to diagnose performance


stepjam (approved):
LGTM

PR #833, opened 2026-07-31
fix: multi process retries (#830)

StevenJacobs61:
### Features

 - Introduced allure metrics logging in tests


StevenJacobs61:
Wrong name

PR #836, opened 2026-07-31
fix: manual test timestamps

StevenJacobs61:
### Features

 - Implemented pre-commuted manual timestamps for test logging.

### DIFF
--- a/tests/integration/platform/data_daemon/behavioural_correctness/test_cancel_recording.py
+++ b/tests/integration/platform/data_daemon/behavioural_correctness/test_cancel_recording.py
@@ -37,7 +37,7 @@ from tests.integration.platform.data_daemon.shared.test_case.build_test_case_con
 from tests.integration.platform.data_daemon.shared.test_case.constants import (
     MAX_TIME_TO_START_S,
     STOP_RECORDING_OVERHEAD_PER_SEC,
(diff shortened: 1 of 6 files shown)


tests/integration/platform/data_daemon/behavioural_correctness/test_cancel_recording.py
```diff
@@ -61,7 +61,7 @@
             video_count=1,
             image_width=64,
             image_height=64,
-            timestamp_mode=TIMESTAMP_MODE_REAL,
+            timestamp_mode=TIMESTAMP_MODE_MANUAL,
```
favour-neuraco, line 64: Is this second case intended to remain TIMESTAMP_MODE_REAL? Manual is already the default, so these two cases are now identical and we lose real-timestamp coverage.

favour-neuraco (approved):
LGTM with comment

tests/integration/platform/data_daemon/behavioural_correctness/test_cancel_recording.py
StevenJacobs61: I have removed this dublicate and kept all stochastic cases intact also in order not to weaken the assertions.

PR #847, opened 2026-08-03
chore: limit perf testing to only mac and linux py314

stepjam:
As discussed in meeting, limit the perf tests to py3.14 on mac and linux.


StevenJacobs61 (approved):
LGTM

PR #849, opened 2026-08-04
fix: bound untimed HTTP requests and keep pooled connections alive

stepjam:
When the client goes several minutes without calling the API, the network path between it and the server silently discards the idle TCP connection. Because no request set a time limit, the client's next call was written into that dead connection and then waited ~16 minutes for the kernel to give up instead of reconnecting.

The fix lives in `thread_local_session()` so all ~60 previously-untimed call sites are covered without touching any of them:

1. **TCP keep-alive** (60s idle / 20s interval / 3 probes) keeps the flow observed as live so intermediaries do not drop it. If the peer has gone, the probes fail and the socket is discarded in the background rather than being written into.
2. **`TCP_USER_TIMEOUT=30s`** (Linux) caps retransmission of unacknowledged data, replacing the ~924s `tcp_retries2` default.
3. **`DEFAULT_TIMEOUT=(5s, 30s)`** supplied by `_DefaultTimeoutAdapter.send()` only when a caller passes none, so explicit timeouts still win.

### Features

 - `Timer` apportions every measurement between HTTP and local time making future debugging easier.
 
 - `Timer` blocks with a limit of at least 5s dump every thread's stack at the moment they breach, while still stuck. Sub-second per-frame logging timers are excluded, as they run tens of thousands of times per test.
 
 - `http_time_snapshot()` exposes cumulative request count and duration for the calling thread.

### Bugfixes

 - Requests issued through `thread_local_session()` are now bounded in time and their pooled connections kept alive, so a silently-dropped connection fails in seconds or is transparently retried rather than blocking for ~16 minutes. This affects the shipped client as well as CI: any user whose script pauses between API calls behind NAT hits the same hang.

Note: With this, coupled with some backend fixes unrelated to pooled connections, integration tests for NDD should go green.


.github/workflows/integration-pf-data-daemon-staging.yaml
```diff
@@ -174,7 +174,8 @@ jobs:
       # the expensive half of this workflow, so it runs one Python on each OS.
       matrix:
         python-version: ["3.14"]
-        os: ["ubuntu-24.04", "macos-26"]
+        # Need large mac for more RAM to run the performance suite.
```
StevenJacobs61, line 177: Does this mean we are supporting smaller devices for functionality but admit performance will suffer?

StevenJacobs61 (approved):
LGTM

.github/workflows/integration-pf-data-daemon-staging.yaml
stepjam: Exactly. If the user uses a potato, then they will get an error message that the spool overflowed. There needs to be a small PR by someone to have that error give concrete next steps. E.g. log slower or get a better machine.

PR #850, opened 2026-08-04
feat: poll for dataset statistics while they calculate

stepjam:
Dataset statistics are the per-item mean/std/min/max/quantiles that training uses to normalize data. `PytorchSynchronizedDataset` fetches them from the platform before training, and the backend used to compute them inside the request — up to six minutes on a large dataset. They are now calculated in the background, so the client starts a job and waits on it.

### Features
 - `SynchronizedDataset.calculate_statistics` now starts the calculation, polls it with a progress bar, and fetches the result. Its signature and return type are unchanged, so `PytorchSynchronizedDataset` and its on-disk statistics cache needed no edits.
 - Repeat calls for the same recordings and cross-embodiment descriptions join the running calculation or reuse its result, so an interrupted call resumes instead of starting over. The timeout message says so.
 - The bar counts recordings, then switches to "Aggregating dataset statistics" for the final merge, so a long tail does not look like a hang. An already-complete calculation returns without printing anything.
 - New `examples/example_dataset_statistics.py` prints a dataset's statistics without plotting anything. 

### Related PRs

 - https://github.com/NeuracoreAI/neuracore_types/pull/109
 - https://github.com/NeuracoreAI/neuracore_backend/pull/488

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -137,6 +137,23 @@ python example_retrieve_dataset.py
 If you want to just view your data, then the best way is via the [web interface](https://www.neuracore.com/dashboard/datasets).
 
 
+### Print Dataset statistics
+
+This example shows you how to:
+- Describe which data items a policy reads and which it predicts, by index
+- Ask Neuracore for a synchronized dataset's **statistics** (count, mean, std, min, max, q01, q99)
+- Print them per data type and item, without plotting anything
+
+These are the same statistics training uses to normalize your data. They are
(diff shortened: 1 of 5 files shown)


stepjam:
NOTE: These unit tests will fail until https://github.com/NeuracoreAI/neuracore_types/pull/109 goes in

CougarTasker (approved):
LGTM

PR #861, opened 2026-08-06
fix: chunk checkpoint uploads and run them off the training thread

stepjam:
### Features
 - None

### Bugfixes

 - Checkpoint/artifact/algorithm/endpoint uploads were a single PUT of the whole file to a GCS resumable session, so a transient stall on large checkpoints aborted the entire upload with no resume, forcing a full restart. Uploads now go in 16MiB `Content-Range` chunks with per-chunk retry/backoff, `308`-continue resume, and session-expiry (`404`/`410`) recovery.
 - `save_checkpoint()`/`save_model_artifacts()` now upload on a background worker instead of blocking the training loop, with a wait-before-overwrite guard for reused local paths (`checkpoint_latest.pt`, `model.nc.zip`) and a flush before `DistributedTrainer.train()` returns so the process can't exit mid-upload.

### Items
 - None

### Related PRs
 - None
 
 
 Note: there will be another PR to remove checkpoint_latest.

### DIFF
--- a/neuracore/ml/utils/training_storage_handler.py
+++ b/neuracore/ml/utils/training_storage_handler.py
@@ -1,8 +1,10 @@
 """TrainingStorageHandler for managing model training artifacts and checkpoints."""
 
 import logging
+import threading
+from concurrent.futures import Future, ThreadPoolExecutor
 from pathlib import Path
-from typing import IO, Any
+from typing import Any
 
 import requests
 import torch
@@ -69,6 +71,16 @@ class TrainingStorageHandler(UploadStorageMixin):
                     f"Training job {self.training_job_id} not found or access denied."
                 )
 
+        # Checkpoint/artifact uploads run on this single background worker so
+        # save_checkpoint()/save_model_artifacts() don't block the training
+        # loop for the duration of a large upload. One worker means every
+        # submitted job runs in strict submission order, which is what makes
+        # delete_checkpoint() safe to call right after a save without extra
+        # coordination. Created lazily since it's only needed when uploading.
+        self._upload_executor: ThreadPoolExecutor | None = None
+        self._pending_uploads_lock = threading.Lock()
+        self._pending_uploads: dict[Path, Future] = {}
+
     def _get_upload_url(self, filepath: str, content_type: str) -> str:
         """Get a signed upload URL for a file in cloud storage.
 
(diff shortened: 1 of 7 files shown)


neuracore/ml/utils/training_storage_handler.py
```diff
@@ -236,12 +330,31 @@ def delete_checkpoint(self, relative_checkpoint_path: Path) -> None:
...
+
+        if self.log_to_cloud:
+            with self._pending_uploads_lock:
+                pending = [
+                    future
+                    for path, future in self._pending_uploads.items()
+                    if path.is_relative_to(artifacts_dir)
+                ]
+            for future in pending:
+                future.result()
```
ypang-neuraco: We can reuse `_wait_for_pending_upload` here

neuracore/ml/utils/upload_storage_mixin.py
```diff
@@ -3,11 +3,44 @@
...
+"""Size of each resumable-upload chunk, in bytes.
+
+Must be a multiple of 256 KiB (the GCS resumable-upload requirement for every
+non-final chunk); 16 MiB = 64 x 256 KiB. Matches the chunk size used by the
+Rust data daemon's resumable uploader
+(``rust/data_daemon/src/cloud/coordinators/upload_transfer.rs``), so that a
+transient connection stall costs at most one chunk retry instead of
+restarting the whole file — the flakiness a single giant PUT is prone to on a
+210 MB checkpoint.
+"""
```
ypang-neuraco: Maybe cleanup the docs by removing the mention of the old implementation

neuracore/ml/utils/upload_storage_mixin.py
```diff
@@ -32,10 +122,139 @@ def _get_upload_url(self, filepath: str, content_type: str) -> str:
...
+                    )
+                last_error = f"HTTP {response.status_code}: {response.text}"
+
+            if attempt < _MAX_CHUNK_RETRIES - 1:
+                time.sleep(_BACKOFF_BASE_SECONDS * 2**attempt)
+
+        raise _UploadFailedError(
+            f"chunk at offset {chunk_start} failed after "
+            f"{_MAX_CHUNK_RETRIES} attempts: {last_error}"
+        )
```
ypang-neuraco, line 207: Does the broken file get cleaned up if a chunk fails?

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/utils/upload_storage_mixin.py
stepjam: On the GCS-side resumable session, which may already hold several successfully-uploaded chunks, is just abandoned. So yes it gets "cleaned up" on the server.

neuracore/ml/utils/upload_storage_mixin.py
stepjam: done

neuracore/ml/utils/training_storage_handler.py
stepjam: done

PR #862, opened 2026-08-06
ci: manual specification of test suites for data daemon

damonhayhurst-neuraco:
### Features

 - Able to specify whether you want functional and/or performance on manual dispatch


favour-neuraco (approved):
LGTM

PR #864, opened 2026-08-06
ci: build intel wheel for mac

damonhayhurst-neuraco:
### Features

 - Build intel wheel for mac as `macos-26-large` runner runs on intel mac and needed for perf tests
 - Wheel builds on `macos-26-intel`, smaller intel box but assumption is that wheel doesn't need the resources of large to be built


stepjam (approved):
LGTM

PR #866, opened 2026-08-06
fix: add timeout for ffmpeg encode with disclosed error

damonhayhurst-neuraco:
### Features

 - ffmpeg encoder processes maybe be hanging and we should be informed about that

### DIFF
--- a/rust/data_daemon/src/encoding/video_encoder.rs
+++ b/rust/data_daemon/src/encoding/video_encoder.rs
@@ -31,10 +31,34 @@ use std::ffi::OsString;
 use std::io::Write;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::time::Duration;
 
 use data_daemon_shared::service_name::VIDEO_SPOOL_TICKS_PER_SECOND;
 
 use tokio::process::Command;
+
+/// Floor for a single ffmpeg invocation's deadline.
+const FFMPEG_TIMEOUT_FLOOR: Duration = Duration::from_secs(120);
+
+/// Extra deadline granted per MiB of input, on top of the floor.
+const FFMPEG_TIMEOUT_PER_MIB_S: u64 = 1;
+
+/// Deadline for an ffmpeg invocation reading `input_bytes` of input.
+fn ffmpeg_timeout(input_bytes: u64) -> Duration {
+    FFMPEG_TIMEOUT_FLOOR
+        + Duration::from_secs(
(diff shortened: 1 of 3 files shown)


rust/data_daemon/src/encoding/video_encoder.rs
```diff
@@ -709,17 +749,26 @@ impl VideoEncoder {
...
-            binary: self.binary.clone(),
-            source,
-        })?;
+        let output = match result {
+            Ok(result) => result.map_err(|source| VideoEncodeError::Spawn {
+                binary: self.binary.clone(),
+                source,
+            })?,
+            Err(_) => {
+                return Err(VideoEncodeError::Timeout {
```
CougarTasker, line 766: Question: does this not convert all types of error into a timeout error?

rust/data_daemon/src/encoding/video_encoder.rs
```diff
@@ -31,11 +31,35 @@ use std::ffi::OsString;
...
 
 use data_daemon_shared::service_name::VIDEO_SPOOL_TICKS_PER_SECOND;
 
 use tokio::process::Command;
 
+/// Floor for a single ffmpeg invocation's deadline.
+const FFMPEG_TIMEOUT_FLOOR: Duration = Duration::from_secs(120);
+
+/// Extra deadline granted per MiB of input, on top of the floor.
+const FFMPEG_TIMEOUT_PER_MIB_S: u64 = 1;
```
CougarTasker, line 44: Suggestion: These number seem arbitrary or machine dependant, I think a clean cap is fairer

CougarTasker (approved):
LGTM \w comment

rust/data_daemon/src/encoding/video_encoder.rs
damonhayhurst-neuraco: The match statement is for the timeout function itself so Error will always be timeout but, agree, it looks confusing and is unnecessary indirection.

rust/data_daemon/src/encoding/video_encoder.rs
damonhayhurst-neuraco: I would push back on this. Providing a hard cap means we might be providing the wrong failure reason while the encode is still doing work. Having this value informed by process rate mitigates against this.

CougarTasker: We discussed this offline, what we are looking for is a more granular update rather than simply how big the file is. DH suggested we look at getting progress updates from standard out instead

CougarTasker (changes requested):

PR #868, opened 2026-08-07
fix: validate client version should retry on transient errors

ypang-neuraco:
### Bugfixes
- validate version can fail on transient backend service unavailable errors and asks users to upgrade neuracore (https://github.com/NeuracoreAI/neuracore/actions/runs/31136836144/job/92738329547)
- enable transient retries on `verify-version`
- only raise `VersionMismatchError` on 4xx version responses to avoid confusion


stepjam (approved):
LGTM

PR #871, opened 2026-08-07
chore: ml tests are now self contained

stepjam:
This PR makes ml integration runs independent. Before this, some integration tests required previous jobs to have run, which caused issues if you needed to re run a job, or if the job that came before failed for some reason.


ypang-neuraco (commented):
I'm not sure we should do this, I think it is better that the inference and resume test does not depend on data recording. If the inference/resume tests failed they should not have cleaned up the latest training run and a re-run should be possible. If the latest training flow test failed then the inference/resume test should look for the previous successful run.

stepjam:
> I'm not sure we should do this, I think it is better that the inference and resume test does not depend on data recording. If the inference/resume tests failed they should not have cleaned up the latest training run and a re-run should be possible. If the latest training flow test failed then the inference/resume test should look for the previous successful run.

We do need to chanage something, because our last failure was because the dataset (for what ever reason) did not exist. What about running on a shared dataset instead in that case?

ypang-neuraco:
Discussed offline, we should go back to the earlier approach where training/inference/resume is all in one test, but a failure in inference should not block the resume test and vice versa

stepjam:
@ypang-neuraco PR updated as discussed

PR #874, opened 2026-08-10
feat: route streaming endpoints to the stream service

stepjam:
The backend now serves long-lived connections from a Cloud Run service of its own, reached under an `/api/stream/` path prefix. This points the client's streaming calls at it.

### Features
 - Added `STREAM_API_URL` and switched the WebRTC signalling calls and the recording notification stream over to it. Ordinary API calls still use `API_URL`. **Breaking:** this version needs a backend running the matching release.

### Related PRs
 - NeuracoreAI/neuracore_backend#511
 - NeuracoreAI/neuracore_frontend#423

> [!IMPORTANT]
> Must merge together with the backend PR. The backend serves these endpoints only at the new paths, so an older client version cannot open streams against the new backend.

### DIFF
--- a/changelogs/pending-changelog.md
+++ b/changelogs/pending-changelog.md
@@ -10,3 +10,5 @@ Example: "This release adds support for multi-GPU training and improves streamin
 ## Summary
 
 <!-- Append your summary here -->
+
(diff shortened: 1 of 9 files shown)


CougarTasker (approved):
LGTM 👍

PR #875, opened 2026-08-11
feat: add daemon version check

CougarTasker:
### Features
 - The data daemon exposes a new iceoryx2 request/response service (`neuracore/data_daemon/version`) that reports the neuracore version it was built with. The service copies the existing `health` service pattern, so old producers are unaffected.
 - `ensure_daemon_running()` now checks the running daemon's version against the installed package before the daemon is used, with the same time budget as the health probe. On mismatch it raises `DaemonVersionMismatchError` with steps tailored to the situation: an adopted stale daemon gets stop and restart guidance, a just-launched stale daemon is stopped again and gets rebuild guidance, and an install without a binary is told to reinstall first. A daemon that answers the health probe but stays silent on the version query for the whole budget is reported as too old to report a version.
 - Failure causes other than a real mismatch do not blame the daemon: missing package metadata (bare source tree) and transport failures skip the check with a warning, and a native extension that predates the version query raises an error naming the extension, with reinstall and rebuild commands that work (`pip install --force-reinstall`).
 - The rust workspace version is now synced to the python package version (13.7.2). All three crates inherit it, the daemon reports plain `CARGO_PKG_VERSION`, and `bump2version` now rewrites `rust/Cargo.toml` and `rust/Cargo.lock` together with `pyproject.toml` on every release. A new unit test pins the two versions together so a drift fails CI instead of the release job.

### Items
 - [NCP-1213](https://neuracore.atlassian.net/browse/NCP-1213)

> Note:
> An emergency deploy from a branch cut before this change will fail the version bump (`rust/Cargo.toml` still at 0.1.0 there); the fix is to cherry-pick the Cargo version sync onto the branch.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -6,3 +6,11 @@ tag_name = v{new_version}
 message = Bump version to {new_version} [skip ci]
 
 [bumpversion:file:pyproject.toml]
+
+[bumpversion:file:rust/Cargo.toml]
+search = version = "{current_version}"
+replace = version = "{new_version}"
+
+[bumpversion:file:rust/Cargo.lock]
+search = version = "{current_version}"
+replace = version = "{new_version}"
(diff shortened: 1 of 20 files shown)


damonhayhurst-neuraco (approved):
LGTM

PR #876, opened 2026-08-12
fix: explicit boundary and flush marker mitigates against orphaned video chunks

damonhayhurst-neuraco:
### Bugfixes

 - Setting explicit boundary informs the writer to open the new chunk at the start of the recording window 
 - Setting a flush marker instead of using inherent holdback duration to ensure chunks that are intended for the recording survive. Timeout still in play in case flush marker doesn't arrive.


CougarTasker (approved):
LGTM 👍

PR #877, opened 2026-08-12
fix: hold a recording window open for every producer that owns its video

damonhayhurst-neuraco:
### Bugfixes

 - Ensure that all process that own video for the recording have their chunks intended for the recording survive the boundary


CougarTasker (approved):
LGTM, these changes probably could have been merged with 876 touching the same bits of the same files

PR #879, opened 2026-08-12
fix: cut a video chunk at the recording boundary it straddles

damonhayhurst-neuraco:
### Features

 - Chunks have offset applied making sure any chunk that overspills the recording window is split at ffmpeg encode.


CougarTasker (approved):
LGTM 🎉  we have been needing this

PR #880, opened 2026-08-12
fix: confine a non-owning producer to the recording it was told about

damonhayhurst-neuraco:
### Bugfixes

 - Function for non owning video producers to mark the boundary window for their source. **Caveat** Process owning start recording call must call stop.

### DIFF
--- a/neuracore/core/robot.py
+++ b/neuracore/core/robot.py
@@ -10,6 +10,7 @@ import io
 import logging
 import os
 import tempfile
+import threading
 import time
 import uuid
 import xml.etree.ElementTree as ET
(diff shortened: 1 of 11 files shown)


neuracore/core/robot.py
```diff
@@ -358,20 +367,24 @@ def _drain_streams_and_notify_daemon(
...
+        their local gate and run their writer barrier anyway; these are per-process
+        and their tail chunks are sealed nowhere else.
         """
         try:
             self._stop_all_streams()
             context = self._get_daemon_recording_context()
             try:
-                context.stop_recording(timestamp=timestamp)
+                if self._owns_daemon_recording:
+                    context.stop_recording(timestamp=timestamp)
```
CougarTasker, line 370: blocker: I do not like silently dropping the user's stop recording here I think we need another approach

CougarTasker (changes requested):

neuracore/core/robot.py
damonhayhurst-neuraco: This is addressed in the caveat I outlined in the description. I agree we shouldn't be dropping a user's stop recording function ever. However, this will never happen in 100% of cases we know about as start/stop recording are always called from the same process. 
I don't think this should be approach long term as we shouldn't assume that the user is going to call stop recording from the same process they started the recording on. It should be process agnostic. 
I'd argue that calling stop recording from a different process is an edge case and should be revisited down the line and our integration suite should account for this. 
Happy to continue pursuing this or I can add a strongly worded comment?

CougarTasker: I disagree that it is an edge case, what was your motivation to confine the start and stop to the same proccess?

damonhayhurst-neuraco: Simplicity, real world use cases

PR #883, opened 2026-08-13
feat: image augmentation

ypang-neuraco:
### Features
- Add ColorJitter and GaussianNoise for RGB (defaults in config.yaml)
- Separate preprocessing configs for training and inference (validation uses inference config)
- Only inference configs are stored in the artifact

### DIFF
--- a/neuracore/ml/config/config.yaml
+++ b/neuracore/ml/config/config.yaml
@@ -59,7 +59,32 @@ trim_start_end: true
 
 # Dict[str, Dict[DataType, List[MethodConfig]]], e.g.:
 # for images, we must have a resize step to reshape all images from different sources to the same size.
-preprocessing:
+# train_preprocessing is used on the training split (may include augmentations).
+# inference_preprocessing is used for validation and packaged into the model archive.
+train_preprocessing:
+  input:
+    RGB_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+      - _target_: neuracore.ml.preprocessing.methods.color_jitter.ColorJitter
+        brightness: 0.2
+        contrast: 0.2
+        saturation: 0.2
+        hue: 0.05
+      - _target_: neuracore.ml.preprocessing.methods.gaussian_noise.GaussianNoise
+        std: 5.0
+    DEPTH_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+  output:
+    RGB_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+    DEPTH_IMAGES:
+      - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
+        size: [224, 224]
+
+inference_preprocessing:
   input:
     RGB_IMAGES:
       - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
--- a/neuracore/ml/preprocessing/methods/color_jitter.py
+++ b/neuracore/ml/preprocessing/methods/color_jitter.py
@@ -1,0 +1,66 @@
+"""Color jitter preprocessing method (training augmentation)."""
+
+from __future__ import annotations
+
+import torchvision.transforms as T
+from neuracore_types import BatchedNCData, BatchedRGBData, DataType
+
+from ..base import PreprocessingMethod
+
+
+class ColorJitter(PreprocessingMethod):
+    """Randomly jitter brightness, contrast, saturation, and hue of RGB frames.
+
+    Expects float frames in ``[0, 255]`` (as produced by ``BatchedRGBData``),
+    applies torchvision color jitter in ``[0, 1]``, then scales back.
+    """
+
+    def __init__(
+        self,
+        brightness: float = 0.3,
+        contrast: float = 0.3,
+        saturation: float = 0.3,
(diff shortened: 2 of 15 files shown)


neuracore/ml/preprocessing/methods/color_jitter.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(
+        self,
+        brightness: float = 0.3,
+        contrast: float = 0.3,
+        saturation: float = 0.3,
+        hue: float = 0.05,
+        train_only: bool = True,
+    ) -> None:
+        """Initialize color jitter parameters (torchvision-style)."""
```
stepjam: Please complete docstring

neuracore/ml/preprocessing/methods/color_jitter.py
```diff
@@ -0,0 +1,60 @@
...
+    def __call__(self, data: BatchedNCData) -> BatchedNCData:
+        """Apply color jitter to RGB frames."""
+        if not isinstance(data, BatchedRGBData):
+            raise TypeError(
+                f"Unsupported batched data type for color_jitter: {type(data)!r}"
+            )
+
+        frame = data.frame
+        batch_size, time_steps, channels, height, width = frame.shape
+        flat = frame.reshape(batch_size * time_steps, channels, height, width)
```
stepjam: Flattening should be necessary; it expects [..., 1 or 3, H, W] shape, where ... means an arbitrary number of leading dimensions

neuracore/ml/preprocessing/methods/gaussian_noise.py
```diff
@@ -0,0 +1,55 @@
...
+        self.clip = clip
+        self.train_only = train_only
+
+    @staticmethod
+    def allowed_data_types() -> frozenset[DataType]:
+        """Return data types supported by this method."""
+        return frozenset({DataType.RGB_IMAGES})
+
+    def __call__(self, data: BatchedNCData) -> BatchedNCData:
+        """Add Gaussian noise to RGB frames."""
```
stepjam: complete docstring args (also the same in a few other places across the PR, so please check)

neuracore/ml/config/config.yaml
```diff
@@ -64,6 +64,15 @@ preprocessing:
     RGB_IMAGES:
       - _target_: neuracore.ml.preprocessing.methods.resize_pad.ResizePad
         size: [224, 224]
+      - _target_: neuracore.ml.preprocessing.methods.color_jitter.ColorJitter
+        brightness: 0.2
+        contrast: 0.2
+        saturation: 0.2
+        hue: 0.05
+        train_only: true
```
stepjam: I suggest we go with a different architecture here. Rather than having a train_only flag, have a train_preprocessing, and a validation_preprocessing Dict[str, Dict[DataType, List[MethodConfig]]].

This does 2 things:
- Makes it more flexible for the future; e.g. having different types of preprocessing based on training / inference. 
- Will reduce the number of conditionals we need in the code

stepjam (changes requested):

neuracore/ml/utils/dataset_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    train_indices = indices[:train_size]
+    val_indices = indices[train_size:]
+
+    train_dataset = copy(dataset)
+    train_dataset._input_preprocessing_config = train_input_preprocessing_config
+    train_dataset._output_preprocessing_config = train_output_preprocessing_config
+
+    val_dataset = copy(dataset)
+    val_dataset._input_preprocessing_config = inference_input_preprocessing_config
+    val_dataset._output_preprocessing_config = inference_output_preprocessing_config
```
stepjam: Accessing private variables; set these as public

neuracore/ml/train.py
```diff
@@ -746,8 +739,8 @@ def _main(cfg: DictConfig) -> None:
             input_cross_embodiment_description=input_cross_embodiment_description,
             output_cross_embodiment_description=output_cross_embodiment_description,
             output_prediction_horizon=cfg.output_prediction_horizon,
-            input_preprocessing_config=input_preprocessing_config,
-            output_preprocessing_config=output_preprocessing_config,
+            input_preprocessing_config=PreprocessingConfiguration(),
+            output_preprocessing_config=PreprocessingConfiguration(),
```
stepjam: Passing in empty `PreprocessingConfiguration` seems a little unintuitive here. Perhaps pass in the train_input_preprocessing_config, which we will overwrite for validation later

neuracore/ml/train.py
```diff
@@ -416,10 +419,16 @@ def run_training(
                 "or changing the validation split."
             )
 
-        # Use random split with fixed seed for deterministic behavior
-        generator = torch.Generator().manual_seed(cfg.seed)
-        train_dataset, val_dataset = random_split(
-            dataset, [train_size, val_size], generator=generator
```
stepjam, line 422: Nit: rather than have to redefine random_split ourselves in `split_train_val_datasets`, what about:

```python
generator = torch.Generator().manual_seed(cfg.seed)
train_dataset, val_dataset = random_split(
    dataset, [train_size, val_size], generator=generator

# Note that both train_dataset and val_dataset will have train preprocessing config, 
# so update val_dataset to use inference preprocessing config
val_dataset = copy(val_dataset)
val_dataset.input_preprocessing_config = inference_input_preprocessing_config
val_dataset. output_preprocessing_config = inference_ouput_preprocessing_config
```

stepjam (approved):
LGTM w/c

PR #884, opened 2026-08-13
fix: batch size validation improvements

dennisthevara:
### Bugfixes
 - Currently, batch size auto validation applies a safety factor fraction to the batch size that barely fits in memory. Sometimes auto batch size validation still results in subsequent OOM issues since this safety factor is eyeballed and not deterministic. This PR makes the safety factor more conservative until a more robust validation flow can be implemented.
 - Manually set batch sizes would not apply the safety factor as above, and would validate any batch size that barely fit in memory, without considering the additional memory overhead of other utilities that would be spun up during actual training, resulting in false positives. This PR 

### Items
 - [NCRL-711](https://neuracore.atlassian.net/browse/NCRL-711)

### DIFF
--- a/neuracore/ml/trainers/batch_autotuner.py
+++ b/neuracore/ml/trainers/batch_autotuner.py
@@ -38,6 +38,12 @@ _NUM_WARMUP_ITERATIONS = 1
 # this factor and re-probed until it fits -- a cheap alternative to bisection.
 _DOWNSCALE_FACTOR = 0.9
 
+# Validation reuses the autotuner's own headroom (budget * safety) so a manually
+# set batch size is accepted only if it leaves the same margin an auto-selected
+# one would.
+_VALIDATION_MAX_GPU_UTILIZATION = 0.9
+_VALIDATION_SAFETY_FACTOR = 0.6
+
 
 @dataclass
 class BatchProbeResult:
@@ -417,7 +423,7 @@ class BatchSizeAutotuner:
         min_batch_size: int = 2,
         max_batch_size: int = 512,
         num_iterations: int = 2,
-        safety_factor: float = 0.7,
+        safety_factor: float = 0.6,
     ):
         """Initialize the batch size auto-tuner.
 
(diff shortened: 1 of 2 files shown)


neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -834,14 +840,48 @@ def find_optimal_batch_size(
     return optimal_batch_size
```
sarthakdas, line 844: I would put this inside the Batchsize Validator Class as this is validating the batch size

neuracore/ml/trainers/batch_autotuner.py
```diff
@@ -415,7 +421,7 @@ def __init__(
         min_batch_size: int = 2,
         max_batch_size: int = 512,
         num_iterations: int = 2,
-        safety_factor: float = 0.7,
+        safety_factor: float = 0.6,
```
sarthakdas, line 426: Use the Global Variable here

sarthakdas (commented):
I have a slight problem with this because a batchsize generated by Auto would give a value and then if I were to manually try this batchsize this would fail, as i would hit the Vram limit. We should impliment this same celing as Auto Size. Two options here we let this go though and we create a seperate ticket for this on Jira.

neuracore/ml/trainers/batch_autotuner.py
dennisthevara: the `_fits_within_budget` method was added to the validator class

neuracore/ml/trainers/batch_autotuner.py
dennisthevara: global variable applied

dennisthevara:
> I have a slight problem with this because a batchsize generated by Auto would give a value and then if I were to manually try this batchsize this would fail, as i would hit the Vram limit. We should impliment this same celing as Auto Size. Two options here we let this go though and we create a seperate ticket for this on Jira.

Agreed, since unifying the divergent methodologies would require a more significant re-write, I recommend pushing this through as a hotfix, and creating a re-write ticket to resolve the divergence.

sdas-neuraco (approved):
LGTM

PR #885, opened 2026-08-14
fix: strip data type prefix from remote sync point keys

stepjam:
### Features
 - New integration test covering multi-node policy inference. Nightly workflows running it against staging and production.

### Bugfixes
 - Remote sync point keys no longer carry a data type prefix. Providers label JSON tracks `"{DATA_TYPE}:{sensor_name}"`, but `get_latest_sync_point` keys local streams by the bare sensor name, so data logged on a remote node arrived as `JOINT_POSITIONS:joint_0` and could never match an embodiment description. This affected every JSON data type (joints, grippers, poses, language, custom 1d, point clouds); cameras were unaffected.


sdas-neuraco (approved):
LGTM

PR #888, opened 2026-08-16
perf: run image augmentation on the training device, not the dataloader

stepjam:
### Features

 - `PreprocessingMethod` gains an `on_cpu` flag. `True` (the default) keeps a method in the dataset, per sample, inside a DataLoader worker; `False` runs it in the trainer, per batch, once the batch is on the training device.
 - `ColorJitter` and `GaussianNoise` opt out and now run batched on the device. `ResizePad` stays on the worker: it must precede collation, since recordings differ in resolution and a batch cannot be stacked until frames match.

### Trade-off
 - `T.ColorJitter` draws one factor set per call, so per batch on the device every sample in a batch shares a colour transform where per sample it got its own. Accepted deliberately: augmentation still varies batch to batch, and per-sample factors would mean reimplementing all four adjustments with broadcasting, including a batched RGB/HSV conversion for hue.

### DIFF
--- a/neuracore/ml/config/config.yaml
+++ b/neuracore/ml/config/config.yaml
@@ -65,6 +65,12 @@ trim_no_movement_at_start_threshold: null
 # for images, we must have a resize step to reshape all images from different sources to the same size.
 # train_preprocessing is used on the training split (may include augmentations).
 # inference_preprocessing is used for validation and packaged into the model archive.
+#
+# ORDERING: methods do not necessarily run in the order listed here. Each method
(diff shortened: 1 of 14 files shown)


ypang-neuraco (approved):
LGTM, please leave a note (maybe in config.yaml) that the preprocessing methods may not apply in the order specified in the config due to this change

stepjam:
> please leave a note (maybe in config.yaml) that the preprocessing methods may not apply in the order specified in the config due to this change

Agreed. Have done.

PR #889, opened 2026-08-16
perf: precompute static dataset config instead of rebuilding per sample

stepjam:
### Features
 - `load_sample` no longer rebuilds config that cannot change once the dataset exists. Padding widths, the indexed embodiment mapping and its sort order are derived once in `__init__`.
 - `__getitem__` recovers a timestep from a precomputed per-episode start offset rather than scanning `episode_indices` for the episode's first occurrence, which was O(N) per access and so O(N^2) per epoch.
 - `collate_fn` caches which attributes hold tensors per type, instead of re-deriving it by reflection for every item of every data type of every batch.


ypang-neuraco (approved):
LGTM

PR #890, opened 2026-08-16
perf: cache built training samples on disk across epochs and runs

stepjam:
### Features

 - Fully built training samples are cached under `~/.neuracore/training/sample_cache/<hash>/<recording_id>/<timestep>.pt`, skipping frame decode and tensor construction on every epoch after the  first and on any later run of the same configuration. Controlled by
   `dataset_sample_cache`.
 - Reads happen before the recording is touched, keyed on a recording id resolved at construction, so a hit is one file read.
 
### Measurement
 - 24.44 ms/sample cold (build and store) against 0.67 ms/sample warm: 36.6x.
 - The first epoch is slower by the cost of writing entries; it pays back within the next epoch.

### DIFF
--- a/neuracore/ml/datasets/pytorch_synchronized_dataset.py
+++ b/neuracore/ml/datasets/pytorch_synchronized_dataset.py
@@ -29,6 +29,7 @@ from neuracore.core.utils.training_input_args_validation import (
     _validate_cross_embodiment_description_against_dataset,
 )
 from neuracore.ml import BatchedTrainingSamples
+from neuracore.ml.datasets.batch_sample_cache import BatchSampleCache
 from neuracore.ml.datasets.pytorch_neuracore_dataset import PytorchNeuracoreDataset
 from neuracore.ml.preprocessing.base import PreprocessingConfiguration
 from neuracore.ml.utils.json_serialization import JsonValue, to_json_serializable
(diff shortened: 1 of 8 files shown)


neuracore/ml/datasets/pytorch_synchronized_dataset.py
```diff
@@ -287,21 +328,26 @@ def _validate_cross_embodiment_specs(
             description_kind="Output",
         )
 
-    def _get_episode_indices(self) -> tuple[list[int], list[int]]:
+    def _get_episode_indices(self) -> tuple[list[int], list[int], list[str]]:
         """Map each sample to its episode, and each episode to its first sample.
```
ypang-neuraco: please add short description of new output

ypang-neuraco (approved):
LGTM \w comments
Will need to update the minimum disk size estimation logic (https://github.com/NeuracoreAI/neuracore_types/pull/104)

stepjam:
> minimum disk size estimation logic

Noted on the  minimum disk size estimation logic

neuracore/ml/datasets/pytorch_synchronized_dataset.py
stepjam: Done. And also renamed it too to make it more clear

PR #891, opened 2026-08-16
feat: log host and accelerator utilisation, raise dataloader workers

stepjam:
### Features

 - Host and accelerator utilisation is logged under a `System/` prefix, so it groups separately from `train/` and `val/`: CPU %, RAM %, disk % for the cache volume, GPU utilisation %, and GPU memory reserved (GB and %).
 - Doubled num_train_workers and num_val_workers


ypang-neuraco (approved):
LGTM

PR #892, opened 2026-08-16
perf: remove per-step syncs, load payloads lazily, cache autotune results

stepjam:
### Features

 - Autotune results are cached on disk against the model, its input shapes and the GPU, so a repeat launch of the same configuration skips probing entirely. 
 - Batch size resolution moved out of `train.py` into  `neuracore/ml/utils/batch_size.py` alongside its cache.

### Bugfixes

 - Per-step device synchronisations removed from the training loop:
   - `MemoryMonitor` no longer calls `torch.cuda.synchronize()`. `memory_reserved()` reads host-side allocator bookkeeping, so the barrier bought nothing while serialising every step. 
   - Losses and metrics accumulate on-device and are read back once per epoch, replacing a blocking `.item()` per loss and per metric per step.
   - The progress-bar `loss.item()` ran unconditionally, including on cloud runs where the bar is disabled. It now runs only when the bar is visible.
 - Histograms are rank-guarded and skipped when the backend discards them. `CloudTrainingLogger` records nothing, so cloud runs were walking every parameter twice per log step to hand it to a no-op.
 - The training-progress PUT no longer blocks the training thread; updates coalesce onto a background worker and are flushed at shutdown.
 - Sync point payloads load lazily. The output window is projected onto the output description alone and reuses the already-loaded input timestep, so a  joints-only output no longer decodes a camera frame per horizon step and discards it.
 - Autotune probes reuse one batch instead of loading three serially; the two defensive `deepcopy` calls on the dataset are gone (probes run in spawned  subprocesses, so the parent is already isolated by pickling); packaged YAML  loads are memoised; only rank 0 verifies the training job exists.

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -519,12 +519,19 @@ class SynchronizedRecording:
                 self._delete_decoding_lock(lock_file)
 
     def _load_sync_point_payloads(
-        self, sync_point: SynchronizedPoint
+        self,
+        sync_point: SynchronizedPoint,
+        data_types: frozenset[DataType] | None = None,
     ) -> SynchronizedPoint:
         """Load lazy sensor payloads from disk cache for a sync point.
 
         Args:
             sync_point: Sync point with metadata-only camera and point cloud entries.
+            data_types: If given, only these data types are materialised and the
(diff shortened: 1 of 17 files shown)


stepjam:
Closing for now as some of these changes are not urgent. Will pull out the urgent ones.

PR #893, opened 2026-08-16
perf: remove per-step device syncs from the training loop

stepjam:
### Bugfixes / Features

 - Dropped a full device barrier from every training step. `MemoryMonitor.check_memory` called `torch.cuda.synchronize()` before reading `memory_reserved()`, which is host-side allocator bookkeeping and
   accurate without one. 
 - `MemoryMonitor` defaults to `gpu_id=0`, so on multi-GPU every rank was watching GPU 0's memory instead of its own. It now takes the rank's device.
 - Removed the per-step device-to-host syncs. Losses and metrics accumulate on device via `MetricAccumulator` and are read back once per epoch, replacing an `.item()` per loss and per metric per step. This also drops two unbounded per-step lists that were held for a whole epoch.
 - `pbar.set_postfix` called `loss.item()` unconditionally, including on cloud runs where the bar is disabled and nobody sees the result. It is now gated on the bar being visible and on it being a logging step.
 - Weight and gradient histograms are skipped when the backend discards them (`CloudTrainingLogger.log_histogram` is a no-op, so cloud runs walked every parameter twice per logging step for nothing) and on non-zero ranks, which were racing into one TensorBoard directory. `_log_scalars` was already  rank-guarded; these two were not. 
 - Training progress updates no longer block the training loop on an HTTPS PUT. They run on their own worker with coalescing, so a slow endpoint costs staleness rather than a backlog, and are flushed before the process exits.

### DIFF
--- a/neuracore/ml/trainers/distributed_trainer.py
+++ b/neuracore/ml/trainers/distributed_trainer.py
@@ -21,6 +21,7 @@ from neuracore.ml.logging.system_metrics import (
 )
 from neuracore.ml.logging.training_logger import TrainingLogger
 from neuracore.ml.preprocessing.base import PreprocessingConfiguration
+from neuracore.ml.trainers.metric_accumulator import MetricAccumulator
 from neuracore.ml.utils.device_utils import get_default_device
 from neuracore.ml.utils.memory_monitor import MemoryMonitor, OutOfMemoryError
 from neuracore.ml.utils.preprocessing import apply_device_preprocessing
@@ -30,6 +31,11 @@ logger = logging.getLogger(__name__)
 
 # Only update the training metadata every N steps to avoid excessive API calls
 UPDATE_TRAINING_METADATA_EVERY = 20
+
+# Checking RAM and VRAM headroom reads /proc/meminfo and the CUDA allocator,
+# which is far more often than a slow-moving quantity needs. Matches the
+# interval the dataset uses for the same check.
+CHECK_MEMORY_INTERVAL = 100
 
 
 class NestedModule(nn.Module):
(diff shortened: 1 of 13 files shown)


neuracore/ml/trainers/distributed_trainer.py
```diff
@@ -134,6 +140,8 @@ def __init__(
         self.num_epochs = num_epochs
         self.log_freq = log_freq
         self.system_log_freq = system_log_freq
+        self._pbar_enabled = rank == 0 and not storage_handler.log_to_cloud
+        self._histograms_enabled = rank == 0 and training_logger.supports_histograms
```
ypang-neuraco, line 144: question: why not get rid of pbar and histograms? maybe not so useful for our users
However, if it is useful for our users we should add it to the frontend

neuracore/ml/trainers/distributed_trainer.py
```diff
@@ -173,25 +181,29 @@ def train_epoch(self, epoch: int) -> dict[str, float]:
...
-        epoch_losses: list[dict[str, float]] = []
-        epoch_metrics: list[dict[str, float]] = []
+        accumulator = MetricAccumulator()
 
         memory_monitor = MemoryMonitor(
-            max_ram_utilization=0.8, max_gpu_utilization=0.95
+            max_ram_utilization=0.8,
+            max_gpu_utilization=0.95,
+            # Defaults to GPU 0, which would have every rank watching one
+            # device's memory rather than its own.
```
ypang-neuraco: please remove LLM comment

neuracore/ml/trainers/distributed_trainer.py
```diff
@@ -243,12 +258,13 @@ def train_epoch(self, epoch: int) -> dict[str, float]:
...
                     prefix=SYSTEM_METRIC_PREFIX,
                 )
-            pbar.set_postfix(
-                {"loss": f"{loss.item():.4f}", "step": self.global_train_step}
-            )
-            epoch_losses, epoch_metrics = self._accumulate_epoch_metrics(
-                batch_output, epoch_losses, epoch_metrics
-            )
+            # loss.item() forces a device sync. Only pay for it when a human
+            # is actually watching the bar, and then only as often as we log.
```
ypang-neuraco: please remove LLM comment

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/trainers/distributed_trainer.py
stepjam: Histograms are used for tensorboard (local training), which can be handy for us when training locally. Also, it leaves it open for us to add support later down the line. 
PBar is used for local training, not cloud training.

PR #894, opened 2026-08-17
fix: let any process publish the recording stop, named by window marker

damonhayhurst-neuraco:
### Features

 - Pass window markers on IPC so daemon can resolve recording window across all participating processes

---


damonhayhurst-neuraco:
Superseded. This approach put recording identity on the wire; the truncation bug it targeted is now fixed daemon-side in `fix/name-the-window-a-stop-closes` (`fa2850c7`), which keeps window identity inside the daemon.

PR #895, opened 2026-08-17
test: add producers that log across the recording lifecycle

damonhayhurst-neuraco:
### Features

 - Threaded implementation where log_* is called outside the recording window at frame rate
 - Old threaded implementation is maintained for backwards compatibility


muneebneura (approved):
LGTM

PR #896, opened 2026-08-17
test: verify frame codes against what the producer painted

damonhayhurst-neuraco:
### Features

 - Frame codes we previously contiguous but now could come from any selection of the collection because of continuous logging. Update the assertions to reflect this.

### DIFF
--- a/tests/integration/platform/data_daemon/shared/disk_helpers.py
+++ b/tests/integration/platform/data_daemon/shared/disk_helpers.py
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
     )
     from tests.integration.platform.data_daemon.shared.test_case.build_test_case_context import (  # noqa: E501
         ContextResult,
+        ObservedFrameCodes,
     )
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/platform/data_daemon/shared/disk_helpers.py
```diff
@@ -676,6 +678,47 @@ def _frame_code_failure(
...
+    owed = observed.inside.get(camera_name, [])
+    tolerated = set(owed) | observed.unknowable.get(camera_name, set())
+    prefix = f"{video_path} (camera {camera_name!r}):"
+
+    unexpected = [code for code in codes if code not in tolerated]
+    assert not unexpected, (
+        f"{prefix} {len(unexpected)} frame(s) the producer never logged into"
+        f" this recording — first few: {unexpected[:3]}"
+    )
+    missing = [code for code in owed if code not in set(codes)]
```
muneebneura: Issue: This set is rebuilt on every iteration (O(n2)). I would suggest hoisting the _codes out of this loop. A module level helper is clearer and testable

muneebneura (approved):
LGTM w/c

tests/integration/platform/data_daemon/shared/disk_helpers.py
damonhayhurst-neuraco: Good catch — hoisted to a `present` local above the comprehension, matching the `tolerated` set two lines up.

Left it inline rather than extracting a helper: `_assert_observed_frame_codes` is already the module-level helper, and a separate function for one `set()` call would just add a hop.

PR #897, opened 2026-08-17
test: behavioural correctness test that uses explicit timestamps in start/stop recording

damonhayhurst-neuraco:
### Features

 - Add test for explicit timestamps on recording start/stop are propagated correctly.


favour-neuraco (approved):
LGTM

PR #898, opened 2026-08-17
test: bound the trailing RGB gap at the recording stop boundary

damonhayhurst-neuraco:
### Features

 - Assert no trailing gap for RGB streams


favour-neuraco (approved):
LGTM

PR #899, opened 2026-08-17
test: split the context module by concern

damonhayhurst-neuraco:
### Features

 - Split build test case context into relevant modules for creating producers, defining streams and what they plan to stream, and defining boundaries of relevant timestamps for assertions


muneebneura (approved):
LGTM

PR #900, opened 2026-08-17
test: make the producer channel a case subclass

damonhayhurst-neuraco:
### Features

 - Make producer channels a primary citizen of test case so implementations can be extended into the future


favour-neuraco (approved):
LGTM

PR #903, opened 2026-08-17
test: reclaim leftover daemons only when the daemon-stop check fails

damonhayhurst-neuraco:
### Bugfixes

 - Reclaim left over daemons when test fails isolation assertion so that the failure doesn't leak into other tests.
 - Perform similarly on session end so mid suite termination cleans up


favour-neuraco (approved):
LGTM

PR #905, opened 2026-08-17
ci: change daemon tests to run at 18 minutes past the hour

damonhayhurst-neuraco:
## Summary
- Change the production and staging streaming integration test cron schedules from midnight to 18 minutes past midnight.


CougarTasker (approved):
LGTM 👍

PR #907, opened 2026-08-19
perf: speed up video chunk encoding

CougarTasker:
### Bugfixes

 - The lossy preview downscale now uses the `fast_bilinear` scaler in place of the default swscale bicubic, which raises chunk encode throughput by about 32% on the slow half of the CI runner pool (worst sampled machine 2.75 to 3.51 chunks/s against a 2.5 chunks/s arrival rate).

### Items


damonhayhurst-neuraco (approved):
LGTM

PR #908, opened 2026-08-19
fix: dataset merge blocking till complete and dataset clone description now provided

damonhayhurst-neuraco:
### Bugfixes

 - Dataset merge was non blocking so dataset object returned would be incomplete. Now blocks and provides similar status update as cloning
 - Dataset description on clone was not provided on request so did not propagate

### DIFF
--- a/neuracore/api/datasets.py
+++ b/neuracore/api/datasets.py
@@ -52,12 +52,15 @@ def get_dataset(name: str | None = None, id: str | None = None) -> Dataset:
     return _active_dataset
 
 
-def merge_datasets(name: str, dataset_names: list[str]) -> Dataset:
+def merge_datasets(
+    name: str, dataset_names: list[str], *, wait: bool = True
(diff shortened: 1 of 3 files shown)


neuracore/api/datasets.py
```diff
@@ -141,13 +183,21 @@ def clone_dataset(
...
         dataset_id = source_dataset.id
 
+    description = getattr(source_dataset, "description", None)
+    payload = {
+        "name": new_dataset_name,
+        "sourceDatasetId": dataset_id,
+    }
+    if description is not None:
+        payload["description"] = description
+
```
muneebneura, line 195: Issue: The description wont get carried over if we clone only with dataset id. It needs to be resolved as well in this function like we do for source dataset and dataset name.

Please add a unit test for this case as well

muneebneura (approved):
LGTM w/c

PR #909, opened 2026-08-19
fix: teardown managers logout

StevenJacobs61:
Streaming managers and authenticated background tasks remained active after logout. This allowed SSE/P2P work to race with authentication being cleared, causing authentication errors, retaining stale recording state, and potentially reusing stale resources after a later login.

### Bugfixes

- Shuts down and resets the global P2P orchestrator and recording-state manager during logout.
- Cancels background tasks and recording timers, closes HTTP sessions, clears cached state, and allows clean re-   login.
- Safely closes active P2P connections without dictionary-mutation races.
- Prevents queued work and delayed SSE notifications from restoring recording state after shutdown or a local stop.
- Prevents late Robot.close() calls from recreating the recording manager after logout.
- Removes unused session-injection, fork-handling, and test-only defensive paths.

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -18,7 +18,10 @@ from neuracore.core.streaming.p2p.provider.global_live_data_enabled import (
 from neuracore.core.streaming.p2p.stream_manager_orchestrator import (
     StreamManagerOrchestrator,
 )
-from neuracore.core.streaming.recording_state_manager import get_recording_state_manager
+from neuracore.core.streaming.recording_state_manager import (
+    get_recording_state_manager,
+    shutdown_recording_state_manager,
+)
 from neuracore.core.utils import backend_utils
 
 from ..core.auth import get_auth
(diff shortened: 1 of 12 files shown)


neuracore/api/core.py
```diff
@@ -103,7 +106,16 @@ def logout() -> None:
...
-    get_auth().logout()
+    # Stop and drain every task that uses the shared Auth instance before its
+    # token is cleared. Each cleanup remains best-effort with respect to the
+    # next one, and credentials are always cleared even if cleanup raises.
+    try:
+        try:
+            StreamManagerOrchestrator.shutdown_global()
+        finally:
+            shutdown_recording_state_manager()
+    finally:
```
CougarTasker, line 117: Blocker: This is slightly concerning snipit, I don't think this is the right approach to solving this problem, this looks like it would have a lot of edge cases. the shutdown would have to be done just right

Suggestion: instead i think making the async work robust to these authentication errors is the right pattern. A inside out approach where the inner loop checks auth and it propergates out rather than this outside in flow.

StevenJacobs61:
Closed and created different solution aimed at the same fix in this new branch/PR ->  https://github.com/NeuracoreAI/neuracore/pull/929

PR #910, opened 2026-08-19
fix: support ffmpeg 9 frame timing option

CougarTasker:
### Bugfixes

 - The daemon selects the passthrough frame timing option that the installed ffmpeg accepts, instead of always sending `-vsync`. ffmpeg 8 removed `-vsync` and ffmpeg 5.1 added `-fps_mode`, so no single spelling covers the supported range. This unblocks the daemon staging runs that fail at preflight since CI moved to ffmpeg 9.
 - The unit workflow now runs the suite a second time against ffmpeg 9, so both ends of the supported range are tested.

### Items

 - [NCP-1275](https://neuracore.atlassian.net/browse/NCP-1275)

### DIFF
--- a/.github/workflows/rust-data-daemon.yaml
+++ b/.github/workflows/rust-data-daemon.yaml
@@ -55,19 +55,44 @@ jobs:
           PYO3_PYTHON: ${{ steps.setup-python.outputs.python-path }}
         run: cargo clippy --workspace --all-targets -- -D warnings
 
-      # The video-encoder tests early-return when ffmpeg is absent, so without
-      # this the encode/preflight path (and the ffmpeg-4.4.2 `-vsync` vs
-      # `-fps_mode` compatibility the daemon defends against) is never exercised
-      # in CI. ubuntu-22.04 ships ffmpeg 4.4.x, matching the target host.
-      - name: Install ffmpeg
+      # The daemon resolves the passthrough option from the installed ffmpeg
+      # (`-vsync` before 5.1, `-fps_mode` from 8.0), so both ends need a run.
+      # apt gives 4.4.x; setup-ffmpeg cannot pin this end because it serves only
+      # the in-support versions, all past the `-vsync` removal.
+      - name: Install oldest tested ffmpeg
         run: |
           sudo apt-get update
           sudo apt-get install -y ffmpeg
+          ffmpeg -version | head -n 1
 
-      - name: Test (workspace)
+      # A runner-image bump would otherwise leave `-vsync` untested, silently.
+      - name: Check the oldest ffmpeg predates -fps_mode
+        run: |
+          if ffmpeg -hide_banner -loglevel error -f lavfi \
+              -i testsrc=duration=1:size=16x16:rate=1 \
+              -fps_mode passthrough -f null - 2>/dev/null; then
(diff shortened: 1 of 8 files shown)


.github/workflows/rust-data-daemon.yaml
```diff
@@ -55,19 +55,45 @@ jobs:
...
           PYO3_PYTHON: ${{ steps.setup-python.outputs.python-path }}
-        run: cargo test --workspace --verbose
+        run: |
+          ffmpeg -version | head -n 1
+          cargo test --workspace --verbose
+
+      # Takes precedence over the apt build on PATH. Keep the pin in step with
+      # the setup-neuracore actions in NeuracoreAI/shared-workflows.
+      - name: Install newest tested ffmpeg
+        uses: AnimMouse/setup-ffmpeg@v1.2.5
```
aditya-neuraco: Please pin this action to a sha hash. Since it’s a third party action, if an attacker gets a hold of the auth creds, they can push malicious code on the same tag and it can affect us.

.github/workflows/rust-data-daemon.yaml
```diff
@@ -55,19 +55,45 @@ jobs:
...
-      # The video-encoder tests early-return when ffmpeg is absent, so without
-      # this the encode/preflight path (and the ffmpeg-4.4.2 `-vsync` vs
-      # `-fps_mode` compatibility the daemon defends against) is never exercised
-      # in CI. ubuntu-22.04 ships ffmpeg 4.4.x, matching the target host.
-      - name: Install ffmpeg
+      # The daemon resolves the passthrough option from the installed ffmpeg
+      # (`-vsync` before 5.1, `-fps_mode` from 8.0), so both ends need a run.
+      # apt gives 4.4.x; setup-ffmpeg cannot pin this end because it serves only
+      # the in-support versions, all past the `-vsync` removal.
+      - name: Install oldest tested ffmpeg
```
aditya-neuraco, line 62: Nit: Would be nice to print the actual tested version in the run step that follows.

aditya-neuraco (approved):
LGTM w/ comments

.github/workflows/rust-data-daemon.yaml
CougarTasker: Pinned to `178e0b4a408f06ce40d896c3cd79f50fa6f8b0c3` (the commit v1.2.5 points at), with the tag kept as a trailing comment.

Worth noting the same action is used unpinned in `NeuracoreAI/shared-workflows` (both setup-neuracore actions, added in shared-workflows#6), so the same exposure exists there. I have not touched that repo in this PR.

.github/workflows/rust-data-daemon.yaml
CougarTasker: Added `ffmpeg -version | head -n 1` to this step, so the "apt gives 4.4.x" claim in the comment above is checked against the log rather than trusted.

It was already printed at the top of each test step; I have dropped that from the oldest-ffmpeg test step now it prints here, and left it on the newest one, where the install is a `uses:` step with no run block of its own. Each leg still prints its version exactly once. Latest run shows `4.4.2-0ubuntu0.22.04.1` and `n9.0.1-6-g9d4ca21220`.

PR #911, opened 2026-08-20
perf: batch video chunk encodes

CougarTasker:
### Bugfixes

 - The daemon now batches up to 8 video chunk encodes into one ffmpeg invocation when a trace falls behind, via the concat demuxer with per-entry `duration` directives computed from capture timestamps. Measured ~12% encode throughput on its own and +52% combined with the `fast_bilinear` scaler from PR #907. When the encoder keeps pace, batches hold one chunk and the invocation is byte-identical to today's single-chunk path.

### Items

### Related PRs

 - Builds on the scaler change from #907
 - #912 make PTS exact


damonhayhurst-neuraco:
Discussed offline: Would be good to update documentation based on decoupling for EncodeWorker from TraceWriter.

damonhayhurst-neuraco (approved):
LGTM

PR #912, opened 2026-08-20
fix: preserve capture timestamps in final videos

CougarTasker:
### Bugfixes

 - Final videos now carry exact trace-relative capture timestamps. Previously every chunk boundary drifted and real capture gaps collapsed to one frame interval; the finalise concat list now declares each segment's true capture span. Verified exact on gapped, jittered, contiguous and composed fixtures, both codec branches.
 - Each segment carries the real mp4 placement extent its batch dictated to ffmpeg, so a finalise span can never declare the next segment inside a segment's content. Deriving that extent from the announced stamps instead was proven to undershoot for writer-synthesized PTS, producing backwards PTS in the final video.

### Items

### Related PRs

 - Stacked on [#911](https://github.com/NeuracoreAI/neuracore/pull/911); retarget after it merges


damonhayhurst-neuraco (approved):
LGTM

PR #915, opened 2026-08-20
ci: pin ffmpeg to 8.1 in importer staging workflow

CougarTasker:
### Bugfixes
 - Pin FFmpeg to 8.1 in the `Integration PF - Importer (Staging)` workflow. The shared `setup-neuracore-from-source` action installs FFmpeg 9.0, but `torchcodec` (used by the lerobot importer path) only ships builds for FFmpeg up to 8. (runs [32316292952](https://github.com/NeuracoreAI/neuracore/actions/runs/32316292952), [32350186238](https://github.com/NeuracoreAI/neuracore/actions/runs/32350186238)). `mcap`/`rlds` were unaffected since they don't decode video through `torchcodec`.

### Related PRs
 - Follows on from #910, which fixed a different FFmpeg-9 regression

PR #916, opened 2026-08-20
fix: support ffmpeg 9 video decoding

CougarTasker:
### Bugfixes
 - `SynchronizedRecording._decode_video` hard-coded `-vsync 0` for the ffmpeg decode pass. FFmpeg 8 removed `-vsync`, so on ffmpeg 8+ every decode falls back to PyAV
 - Adds `_resolve_frame_sync_arg()`, a Python port of the same probe used on the daemon side. it tries `-fps_mode` (ffmpeg >= 5.1, the only spelling from 8.0 on) and falls back to `-vsync` if ffmpeg rejects it as unrecognized, memoized once per process. `_decode_video` now uses the resolved flag instead of the hard-coded one.

### Related PRs
 - Stacked on #915.
 - Companion fix to #910 (daemon side); this is the SDK-side decode path #910 left open.

### DIFF
--- a/neuracore/core/data/synced_recording.py
+++ b/neuracore/core/data/synced_recording.py
@@ -44,8 +44,62 @@ if TYPE_CHECKING:
 MAX_DECODING_ATTEMPTS = 3
 _FFMPEG_AVAILABLE: bool | None = None
 
+# `-fps_mode` is accepted from ffmpeg 5.1 onwards and is the only spelling
+# accepted from 8.0, where the legacy `-vsync` name was removed. Resolved
+# once per process and cached, like _FFMPEG_AVAILABLE above.
+_FPS_MODE_ARG = "-fps_mode"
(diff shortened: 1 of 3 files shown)


neuracore/core/data/synced_recording.py
```diff
@@ -238,8 +292,8 @@ def _decode_video(self, video_location: Path, video_frame_cache_path: Path) -> N
                         "ffmpeg",
                         "-i",
                         str(video_location),
-                        "-vsync",
-                        "0",
+                        _resolve_frame_sync_arg(),
```
damonhayhurst-neuraco: Suggestion: the indirection is confusing here? What happens when we throw on this function call? Resolving this function upfront would be a better read.

damonhayhurst-neuraco (approved):
LGTM w/ comment

PR #919, opened 2026-08-20
ci: install ffmpeg shared libs for torchcodec

CougarTasker:
### Bugfixes
 - torchcodec dlopens `libavutil`/`libavcodec`/... directly, never the `ffmpeg` CLI, so #915's CLI-only pin to 8.1 didn't fix it (still failed on the exact same missing-`.so` error). Supersedes #915: removes its CLI downgrade and instead fetches just the shared libs torchcodec needs at 8.1, leaving the CLI at the shared action's 9.0.

https://github.com/NeuracoreAI/neuracore/actions/runs/32362701672

### Related PRs
 - Supersedes #915's CLI downgrade.

### DIFF
--- a/.github/workflows/integration-importer-staging.yaml
+++ b/.github/workflows/integration-importer-staging.yaml
@@ -101,11 +101,16 @@ jobs:
           extras: dev,import
           pytorch-cpu: "true"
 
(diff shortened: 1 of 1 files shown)


aditya-neuraco (approved):
LGTM

PR #920, opened 2026-08-20
fix: surface a stalled spool backlog as its own error

damonhayhurst-neuraco:
### Features

 - Add specific error code for spool backlog error so producers can catch explicity

---


CougarTasker (approved):
LGTM

PR #921, opened 2026-08-20
fix: gate a producer's data on the window the daemon announces

damonhayhurst-neuraco:
### Features

 - Add a gate to the bridge so data meant for the window can be provided to the window

### Bugfixes

 - Multi process recordings could only be informed by an online only sse gate

### Items

### Related PRs

 - None

---

### DIFF
--- a/neuracore-dictionary.txt
+++ b/neuracore-dictionary.txt
@@ -285,6 +285,7 @@ mcap
 MCAP
 meanpooling
 memfd
+memoryview
 meshgrid
 metadatas
 metafunc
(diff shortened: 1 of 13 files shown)


CougarTasker:
Discussed offline: The potetial savings of a few mem-copys in some cases is not worth the added complexity and failure modes of dropping data in the producer. We prefer the simpler model of allways providing data to the deamon

damonhayhurst-neuraco:
Superseded by #933, which replaces this PR and #922 with a single commit. Rather than announcing the window out to producers and gating locally, the daemon routes each published sample against its own window state — so `gate.rs` and `announcer.rs` are removed rather than refined.

PR #923, opened 2026-08-20
test: keep the timer stats of a context worker that never returns

damonhayhurst-neuraco:
### Features

 - Retain timer stats for logging functions on a test case that fails

---


CougarTasker (approved):
LGTM 👍

PR #927, opened 2026-08-21
test: batch delete cloud robots

damonhayhurst-neuraco:
### Bugfixes

 Batch delete robots was calling robot/list which returns robots in batches of 30 until all had arrived. With my account, I have 2964 so each robot delete resulted in 94 /robot/list calls. For heavier daemon tests with 8 robots the test cleanup was costing 752 /robot/list calls.

Fixed now such that robots can be deleted in batches such that in my case only 94 /robot/list calls will be made to clean up 8 robots. 

This should be a non issue on staging test accont with proper cleanup being applied regularly.

### Items

### Related PRs

 - None

### DIFF
--- a/tests/integration/platform/data_daemon/shared/test_infrastructure.py
+++ b/tests/integration/platform/data_daemon/shared/test_infrastructure.py
@@ -19,7 +19,8 @@ from typing import TYPE_CHECKING
 import pytest
 
 import neuracore as nc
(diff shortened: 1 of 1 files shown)


CougarTasker (approved):
LGTM

aditya-neuraco (approved):
LGTM

PR #928, opened 2026-08-21
perf: prefetch recording videos concurrently from a single thread

stepjam:
### Features

 - New `prefetch_inflight_requests` training config option: how many network requests the post-synchronisation prefetch keeps outstanding at once, all from a single thread. This is now the knob for prefetch speed; `max_prefetch_workers` sizes ffmpeg decoding only.

### Bugfixes

 - Post-synchronisation video prefetch is about x4 faster. It was latency-bound rather than bandwidth-bound: per-object time-to-first-byte dominates and a single stream is slow, while aggregate throughput scales with the number of requests in flight. Giving each recording a thread did not exploit that, because a thread alternated between waiting on the network and running ffmpeg. Network work now runs on one asyncio event loop with several requests outstanding; ffmpeg keeps its thread pool, fed by a bounded queue so staged videos cannot pile up on disk. Peak thread count is unchanged.
 - Each recording's downloads now start as soon as its own metadata arrives, instead of after every recording's, so bytes begin moving almost immediately. Metadata and transfers get separate concurrency budgets, and each recording's signed URLs are minted while it still holds its metadata slot — sharing one budget let long transfers starve the short calls, and a mint asking for its own slot queued behind every remaining metadata request.
 - Cached frames are written at a cheaper zlib level. Decoding is CPU-bound on that compression rather than on the video, and ffmpeg already saturates the machine at a few concurrent processes, which is why `max_prefetch_workers` made little difference. Frames stay lossless and pixel-identical, so existing caches remain valid, at a few percent more disk.
 - Dropped the `wget` package in favour of the shared pooled session on the remaining synchronous downloads. It wrapped `urllib.request.urlretrieve`, which has no timeout and no retry, so a stalled connection could block a worker indefinitely.
 - Download-target collection no longer blocks on another worker's lock, which could previously hold up every remaining camera until that lock went stale. Stale locks are cleared, live ones skipped, and waiting is left to the lazy path.
 - Synchronized metadata is fetched once by the prefetch and passed to each `SynchronizedRecording`, rather than every recording requesting its own.


neuracore/ml/config/config.yaml
```diff
@@ -62,9 +62,15 @@ dataset_sample_cache: true
 num_train_workers: 4
 num_val_workers: 2
 
-# Maximum number of workers to use for prefetching recordings
+# Threads the prefetch uses to decode videos with ffmpeg. Downloading does not
+# use these threads.
 max_prefetch_workers: 4
```
ypang-neuraco: maybe rename to max_prefetch_decode_workers

neuracore/ml/config/config.yaml
```diff
@@ -62,9 +62,15 @@ dataset_sample_cache: true
...
 
+# Network requests the prefetch keeps outstanding at once, all from a single
+# thread. Raise this, rather than max_prefetch_workers, to make the prefetch
+# faster; throughput flattens out beyond ~16.
+prefetch_inflight_requests: 16
```
ypang-neuraco: something like num_concurrent_prefetch_requests would be more descriptive

ypang-neuraco (approved):
LGTM \w comments

neuracore/ml/config/config.yaml
stepjam: done

neuracore/ml/config/config.yaml
stepjam: done

PR #929, opened 2026-08-21
fix: logout event

StevenJacobs61:
### Features

 - None

### Bugfixes

 - Created a logout event so that managers can tear down after logout without causing auth error.
 - Auth errors now propagate up the workers to tear down the manager/resource instead of retying indefiitely.
 - Robot now gets initalised with recording state manager so when it tears down it closes that as opposed to a new one which might have juts been initalised.

### DIFF
--- a/neuracore/core/auth.py
+++ b/neuracore/core/auth.py
@@ -10,6 +10,7 @@ import os
 from typing import Any
 
 import requests
+from pyee import EventEmitter
 
 from neuracore.api.orgs_fetch import fetch_org_ids
 from neuracore.core.config.config_manager import get_config_manager
@@ -55,7 +56,7 @@ def _format_version_validation_error(response: requests.Response) -> str:
     )
 
 
-class Auth(metaclass=SingletonMetaclass):
+class Auth(EventEmitter, metaclass=SingletonMetaclass):
     """Singleton class for managing Neuracore authentication state.
 
     This class handles API key management, access token retrieval, configuration
@@ -64,8 +65,11 @@ class Auth(metaclass=SingletonMetaclass):
     loads saved configuration on initialization.
     """
 
+    LOGOUT_EVENT = "logout"
+
     def __init__(self) -> None:
         """Initialize the Auth instance and load saved configuration."""
+        super().__init__()
         self._access_token = None
 
     def login(self, api_key: str | None = None) -> None:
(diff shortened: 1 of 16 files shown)


neuracore/core/const.py
```diff
@@ -32,6 +32,8 @@
 STREAMING_MINIMUM_BACKOFF_TIME_S = 0.05
 STREAMING_MAXIMUM_BACKOFF_TIME_S = 5
 
+AUTH_LOGOUT_EVENT = "logout"
```
CougarTasker: Nit: i prefer when these events are on the emitter class themselves, keeps the related code together less imports etc

neuracore/core/auth.py
```diff
@@ -156,6 +158,7 @@ def logout(self) -> None:
         config_manager.config.api_key = None
         config_manager.config.current_org_id = None
         config_manager.save_config()
+        self.emit(AUTH_LOGOUT_EVENT)
```
CougarTasker: Issue: a listener raising an error will abort the logout before the singletons are reset in the caller

neuracore/core/streaming/recording_state_manager.py
```diff
@@ -145,6 +140,51 @@ def __init__(
...
         self._drain_callbacks: dict[RobotInstanceIdentifier, Callable[[str], None]] = {}
 
+        self._logout_listener = self._on_logout
+        self.auth.add_listener(AUTH_LOGOUT_EVENT, self._logout_listener)
+
+    def _on_logout(self) -> None:
+        """Schedule teardown on the streaming event loop."""
+        self._discard_cached_manager()
+        if not self.loop.is_closed():
+            self.loop.call_soon_threadsafe(self.close)
```
CougarTasker, line 150: Issue: Memory leak. the auth listener is only removed if the manager's loop was not closed `not self.loop.is_closed():` consider using pyee's once method or removing the listener here. 

basically if the recording state manger,org nodes manager or stream manager have allready had the event loop stopped then the instnance and listener are left to leak 

this issue repeated but i will just put it here

CougarTasker (approved):
LGTM \w comments

neuracore/core/const.py
StevenJacobs61: Moved into the Auth class

neuracore/core/auth.py
StevenJacobs61: Wrapped in a try finally block so singletons are rest on listener raised error

neuracore/core/streaming/recording_state_manager.py
StevenJacobs61: Used .once

PR #931, opened 2026-08-25
fix: late sse restart daemon

StevenJacobs61:
### Bugfixes

 - Late SSE start events could restart the data daemon causing flaky issues in tests with the error 'AssertionError: Producer/runner sub processes still alive after test — PIDs: [26886]'.
 - The fix is to keep track of the start times for stopped recordings, so if we receive a notification update from an SSE, we can can compare the start times of that SSE against the start time we have for a stopped recordings, and if the start time coming from the SSE is the same (or earlier) then we ignore it as it is a duplicate late arrival of an already received local stop.


muneebneura (approved):
LGTM 👍

PR #933, opened 2026-08-26
fix: make the daemon the sole authority on whether data is written

damonhayhurst-neuraco:
### Features

 - Producers publish unconditionally; the daemon routes each sample against its own window state, so a process that never called `start_recording` needs no local knowledge of the window. In prior, multi process recordings could only be informed by SSE response.

### DIFF
--- a/neuracore/api/logging.py
+++ b/neuracore/api/logging.py
@@ -226,7 +226,8 @@ def _record_json_to_daemon(
     ``log_json`` entry point is datatype-agnostic, so adding a new JSON type
     needs no daemon-side change.
 
-    A no-op unless a recording is in progress.
+    Always forwarded; the daemon is the one that decides whether a recording
+    is in progress and drops the sample if not.
 
     Args:
         robot: Robot instance owning the daemon recording context.
@@ -235,12 +236,9 @@ def _record_json_to_daemon(
         data: Data object to serialize and persist.
         timestamp: Capture timestamp in seconds.
     """
-    if robot.get_current_recording_id() is None:
-        return
+    context = robot._get_daemon_recording_context()
     payload = json.dumps(data.model_dump(mode="json")).encode("utf-8")
-    robot._get_daemon_recording_context().log_json(
-        data_type.value, storage_name, payload, timestamp
-    )
+    context.log_json(data_type.value, storage_name, payload, timestamp)
 
 
 def _publish_video_to_p2p(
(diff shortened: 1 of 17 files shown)


neuracore/api/logging.py
```diff
@@ -386,7 +384,10 @@ def _log_group_of_joint_data(
     if bindings_for_type is None:
         bindings_for_type = {}
         binding_cache[data_type] = bindings_for_type
+    # The bucket/live streams key off this process's own recording handle; the
+    # daemon decides for itself, from the same call, whether it's recording.
```
CougarTasker, line 388: nit: unnecessary comment

neuracore/api/logging.py
```diff
@@ -532,17 +531,19 @@ def _log_camera_data(
...
-            camera_type.value,
-            storage_name,
-            int(image.shape[1]),
-            int(image.shape[0]),
-            image.dtype.name,
-            memoryview(contiguous).cast("B"),
-            camera_data_without_frame.timestamp,
-        )
+    # The daemon decides whether this lands in a recording, not this process —
+    # always forward and let it route by window.
```
CougarTasker, line 535: nit: unnecessary comment

neuracore/api/logging.py
```diff
@@ -532,17 +531,19 @@ def _log_camera_data(
...
+    context = robot._get_daemon_recording_context()
```
CougarTasker, line 536: bloker: without the guard if the user does not start a recording the deamon will not be started, so nothing will drain the spool and it will error. we will now need to make sure it is running when we log the data to the deamon

rust/data_daemon_bridge/src/window_identity.rs
```diff
@@ -0,0 +1,97 @@
...
+//! gates logging, and nothing here talks to the daemon. The one thing a
+//! producer still needs to know locally is what to call the window a later
+//! `stop_recording()` from THIS process means, so a duplicate or delayed stop
+//! can't close the wrong recording. A process that opens its own window knows
+//! that boundary first-hand from the very call that opened it, so this is
+//! nothing more than a fork-safe cache of that value.
+//!
+//! A process that did **not** open the window itself has nothing cached here
+//! and asks the daemon instead, once, at `stop_recording()` time — see
+//! [`crate::query::resolve_window_started_at`].
```
CougarTasker, line 13: blocker: I disagree with this archetcture as we have discussed offline before the IPC witin the producer process it should be essentially stateless in terms of the recording. all the processes on this same machine share a clock since the start and stop messages are timestamped so with a small holdback queue the deamon process can de-duplicate and account for any latency jitter like what we have done with the data itself

CougarTasker (changes requested):

PR #936, opened 2026-08-26
fix: resolve video chunk membership per frame, not per chunk

damonhayhurst-neuraco:
### Features

 - Per-frame publish offsets on chunk announcements, so one video chunk can be claimed by several recordings

### Bugfixes

 - Frames logged either side of a recording boundary are no longer dropped with the chunk that held them

### DIFF
--- a/rust/data_daemon/src/encoding/video_encoder.rs
+++ b/rust/data_daemon/src/encoding/video_encoder.rs
@@ -203,9 +203,14 @@ pub struct ChunkEncodeRequest {
     /// Lossy codec selection for this trace. Controls whether a lossless
     /// archive is produced and how the lossy output is encoded.
     pub codec: LossyVideoCodec,
-    /// Frames to encode from the head of `raw_nut`: the whole file unless the
-    /// dispatcher cut the chunk, and always what the sidecar is built from.
+    /// Frames to encode once `skip_frames` are discarded: the rest of the file
+    /// unless the dispatcher cut the chunk, and always what the sidecar is
+    /// built from.
     pub frame_count: u32,
+    /// Leading frames of `raw_nut` to discard before encoding: the ones the
+    /// dispatcher resolved as published before this recording's window opened.
+    /// Zero for a chunk whose first frame the window already owns.
+    pub skip_frames: u32,
 }
 
 /// One NUT entry of a batched chunk encode.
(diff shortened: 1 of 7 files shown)


rust/data_daemon/src/encoding/video_encoder.rs
```diff
@@ -1731,6 +1775,102 @@ mod tests {
         );
     }
 
+    #[tokio::test]
```
CougarTasker, line 1778: Blocker: add a test to make sure we can cut the tail off a recording

CougarTasker (approved):
LGTM \w comment

rust/data_daemon/src/encoding/video_encoder.rs
damonhayhurst-neuraco: The tail cut is already covered — the collapsed diff hides the test names, so it's easy to miss:

- `frame_count_drops_the_tail_and_keeps_the_frames_the_recording_owns` (`video_encoder.rs:1721`, immediately above this comment) — single-chunk tail cut reaches ffmpeg on **both** outputs, so the mp4 can't outrun the sidecar. (Renamed from `frame_count_encodes_only_the_frames_the_recording_owns` — the old name never said "tail", which is presumably why it didn't turn up when you looked.)
- `batch_frame_count_encodes_only_the_frames_the_recording_owns` (`video_encoder.rs:2881`) — same for a batch, where the cap trims the concatenated stream.
- `straddling_video_chunk_keeps_its_in_window_prefix` (`dispatcher.rs:2335`) — end-to-end: a chunk whose frames run past the stop keeps its in-window prefix rather than being dropped.
- `frame_range_cuts_at_the_stop_boundary` (`dispatcher.rs:2028`) — the boundary maths itself.

The new head-cut test also cuts *both* ends in its second case (`skip_frames: 3, frame_count: 2`), which pins the ordering that actually matters: `-frames:v` counts frames after the filter graph, so the cap still bounds the tail of whatever the head cut left.

Chasing this did turn up a real gap next door, though: `ChunkClaim::timestamps` — the slice that keeps the sidecar in step with the cut at both ends — had no test at all, and a wrong slice there would silently misalign the sidecar against the mp4. Added `a_claim_slices_the_sidecar_to_the_frames_it_owns` (`dispatcher.rs:2131`), amended onto the branch.

PR #938, opened 2026-08-26
fix: recording start from web

muneebneura:
Start recording from web was broken because client only updated its own local state, data daemon was never notified of the start sse. which is why logged data was silently dropped. The frontend test for this has been enabled which should now catch error if this flow gets forgotten about in the future.
### Bugfixes

 - Notifies daemon when a recording is started remotely

### Items

 - [NCP-1240](https://neuracore.atlassian.net/browse/NCP-1240)

### Related PRs

 - [PULL-456](https://github.com/NeuracoreAI/neuracore_frontend/pull/456)

### DIFF
--- a/neuracore/core/streaming/recording_state_manager.py
+++ b/neuracore/core/streaming/recording_state_manager.py
@@ -55,10 +55,12 @@ class TrackedRecording(NamedTuple):
             backend reports in ``RecordingStartPayload.start_time``. Orders
             notifications against each other: one describing a recording that
             began earlier than this belongs to an already-finished recording.
+        opened_locally: Whether the recording was started by the local client
     """
 
     recording_id: str
     start_time: float
+    opened_locally: bool
 
 
 def _notify_data_bridge_of_expiry(robot_id: str, instance: int) -> None:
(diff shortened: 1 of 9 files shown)


neuracore/core/streaming/recording_state_manager.py
```diff
@@ -139,6 +141,9 @@ def __init__(
         self._recording_timers: dict[str, list[asyncio.TimerHandle]] = {}
         self.active_dataset_ids: dict[RobotInstanceIdentifier, str] = {}
         self._drain_callbacks: dict[RobotInstanceIdentifier, Callable[[str], None]] = {}
+        self._start_callbacks: dict[
+            RobotInstanceIdentifier, Callable[[str, str, float], None]
+        ] = {}
```
CougarTasker, line 146: Suggestion: use the event emitter pattern rather than juggling callbacks makes it easy to unsubscribe etc

neuracore/core/streaming/recording_state_manager.py
```diff
@@ -465,26 +479,29 @@ def _apply_recording_notification(
...
                 recording_id,
             )
-            # Daemon already ensured above; this is the state transition only.
+
+            # Only open the window if no one else has already.
+            opened_locally = previous is not None and previous.opened_locally
+            if not opened_locally:
+                start_callback = self._start_callbacks.get(instance_key)
+                if start_callback is not None:
+                    start_callback(recording_id, dataset_id, details.start_time)
```
CougarTasker, line 494: nit: use event emitter

CougarTasker (approved):
As a quick patch it is good, but this is starting to look pretty coupled a refactor will be needed as a follow-up

neuracore/core/streaming/recording_state_manager.py
muneebneura: Not too sure about this as our remote stop handler already follows this same callback dict pattern, so this keeps both consistent.
I also tried adding the event emitter here but it causes a deadlock. closing a robot can trigger garbage collector to clean it up mid callback. a callback dict doesn't have this problem as it doesn't involve a loclk

CougarTasker: Okay since this is an urgent fix and @damonhayhurst-neuraco is now looking at the recording state overhaul i mentioned so this is only temp https://neuracore.atlassian.net/browse/NCP-1289

PR #940, opened 2026-08-28
fix: seal video chunks on an age cap

damonhayhurst-neuraco:
### Features

 - Video chunks seal on an age cap as well as size and frame count

### Bugfixes

 - Video from a camera process that never starts or stops a recording is no longer dropped as an orphan


CougarTasker (approved):
LGTM

PR #942, opened 2026-08-28
fix: keep a local recording stoppable when its notification stream is rejected

damonhayhurst-neuraco:
### Bugfixes

 - A rejected notification stream no longer makes nc.stop_recording() a silent no-op


CougarTasker (approved):
LGTM

PR #943, opened 2026-08-28
feat: announce a producer's source bindings to the daemon

damonhayhurst-neuraco:
### Features

 - Producers announce their source bindings, so the daemon knows which robots have a producer on this machine

### DIFF
--- a/neuracore/api/logging.py
+++ b/neuracore/api/logging.py
@@ -226,7 +226,8 @@ def _record_json_to_daemon(
     ``log_json`` entry point is datatype-agnostic, so adding a new JSON type
     needs no daemon-side change.
 
-    A no-op unless a recording is in progress.
+    Always forwarded; the daemon is the one that decides whether a recording
+    is in progress and drops the sample if not.
 
     Args:
         robot: Robot instance owning the daemon recording context.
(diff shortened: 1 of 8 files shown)


neuracore/api/logging.py
```diff
@@ -532,17 +531,19 @@ def _log_camera_data(
...
-            camera_type.value,
-            storage_name,
-            int(image.shape[1]),
-            int(image.shape[0]),
-            image.dtype.name,
-            memoryview(contiguous).cast("B"),
-            camera_data_without_frame.timestamp,
-        )
+    # The daemon decides whether this lands in a recording, not this process —
+    # always forward and let it route by window.
```
CougarTasker, line 535: nit: historical comment

rust/data_daemon/src/pipeline/dispatcher.rs
```diff
@@ -332,6 +340,11 @@ struct Dispatcher {
     actor_handles: Vec<JoinHandle<()>>,
     /// Rate-limited orphan-drop counter (data outside any window).
     orphan_drops: u64,
+    /// Producers bound to each source, by OS pid, with when each was last heard
+    /// from. Says only that a producer for the source exists on this machine —
+    /// never anything about a recording. Read when deciding whether an org-wide
+    /// backend notification concerns a source here.
+    attached: HashMap<Source, HashMap<u32, Instant>>,
```
CougarTasker, line 347: Blocker: I don't understand why we need this. Following the ticket we can remove the recording state manager, recoridng expirey can bubble up with errors and we can use a ireroxy service to respond to nc.is_recording() requests

PR #945, opened 2026-08-28
fix: stop relaying web recording lifecycle through producers

damonhayhurst-neuraco:
### Features

- A process that merely learned of a web-initiated recording no longer publishes either end of its lifecycle: the daemon hears both start and stop from the backend itself, so one web event cannot become one envelope per connected process.
- Such a process still stops its own streams and seals its own tail chunks — statements about itself, safe from any number of processes.
- A local `nc.start_recording()` / `nc.stop_recording()` still publishes, since exactly one process calls it and there is nothing to disambiguate.
- Drops the SDK's remote-start callback registry along with it; the daemon ignores the `robot_name` and `dataset_name` only that path carried, and derives the same capture stamp from the notification.

### DIFF
--- a/neuracore/api/core.py
+++ b/neuracore/api/core.py
@@ -182,7 +182,6 @@ def connect_robot(
     StreamManagerOrchestrator().get_provider_manager(robot.id, robot.instance)
     get_recording_state_manager().register_connected_robot(robot.id)
     robot._register_remote_stop_handler()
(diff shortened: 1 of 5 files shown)


damonhayhurst-neuraco:
Superseded by #947 — same commit, branch renamed to match its content (it removes both ends of the relay, not just the stop).

PR #949, opened 2026-08-28
fix: never clear upload uris on reregistration

CougarTasker:
### Bugfixes
 - The registration coordinator accepted a `registered_traces` entry with an empty `upload_session_uris` map and persisted it as `"{}"`. The uploader reads an empty map as "nothing to send" and calls `finalise_upload(..., 0)`, so the trace settles as `Uploaded` with zero bytes and no PUT. The sensor then reaches the recording manifest with no object behind it.
 - An empty map is now retried through `handle_registration_setback`, the bounded budget already used for backend rejections. Every registered trace asks for at least a trace file, so an empty map is never legitimate.

The serialise-failure branch below already warned about this, and closed one route in:

> A serialise failure must NOT mark the trace registered with a `"{}"` placeholder — that records an empty URI map and the uploader later finalises it as 0 bytes uploaded (silent data loss).

This closes the route that actually fires: a successful response carrying no URIs.

CRC32C cannot catch it. It runs per chunk inside `upload_one_file`, which the empty-map branch returns before reaching.

### Testing
 - `empty_session_uris_are_not_persisted_as_registered` fails without the guard.
 - `cargo test -p data-daemon`: 275 passed. `cargo fmt --check` and `cargo clippy --all-targets` clean.

### Items

### Related PRs
 - NeuracoreAI/neuracore_backend#564

### DIFF
--- a/rust/data_daemon/src/cloud/coordinators/registration.rs
+++ b/rust/data_daemon/src/cloud/coordinators/registration.rs
@@ -281,6 +281,18 @@ async fn submit_batch(
 
                 for trace in &traces {
                     if let Some(uris) = registered_ids.get(&trace.trace_id) {
+                        // The uploader finalises an empty URI map as a 0 byte upload.
+                        if uris.is_empty() {
+                            handle_registration_setback(
+                                store,
+                                registration_attempts,
+                                &trace.trace_id,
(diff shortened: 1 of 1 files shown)


damonhayhurst-neuraco (approved):
LGTM

PR #950, opened 2026-08-28
feat: alpha release from a branch

CougarTasker:
### Features
 - `release.yaml` gains an `alpha_release` option and a `branch` input. The version bump is committed locally and only the tag is pushed, so the branch and `main` are untouched. PR labels and the pending changelog are left alone.
 
### Bugfixes
 - Cargo parses semver, so an alpha carries `15.0.1-alpha.7` in the rust workspace against `15.0.1a7` in `pyproject.toml`. The daemon version check compared the two as raw strings and would have raised `DaemonVersionMismatchError` on every alpha install, so it now reads the cargo spelling as its PEP 440 equivalent.

### Bugfixes
 - None

### Items
 - [NCP-1299](https://neuracore.atlassian.net/browse/NCP-1299)


muneebneura (approved):
LGTM

PR #951, opened 2026-08-28
fix: workflow test artifacts

StevenJacobs61:
### Features

 - Updated workflows to produce allure reports and daemon downloadable artefacts such as state.db and performance results.


CougarTasker (approved):
LGTM

PR #955, opened 2026-09-02
fix: normalize daemon version to PEP440 before comparing in version-c…

favour-neuraco:
### Bugfixes
 - normalize daemon version for PEP440 comparison in test


damonhayhurst-neuraco (approved):
LGTM

PR #957, opened 2026-09-02
fix: stop alpha pre releases reaching prod

StevenJacobs61:
### Bugfixes

 - Used regex to exclude the suffix we provide after a version number as a pre-release tag from testing when running job.


favour-neuraco (approved):
LGTM

PR #958, opened 2026-09-02
fix: support ffmpeg 9's removal of -vsync in frame-code decode

favour-neuraco:
Fixes flatframes test failures in CI by replacing the ffmpeg-9-incompatible -vsync flag with a version-aware probe for -fps_mode/-vsync.


damonhayhurst-neuraco (approved):
LGTM

PR #959, opened 2026-09-02
fix: cancel only the live recording, not the one still draining

damonhayhurst-neuraco:
### Bugfixes

 - A cancel discards only the live recording; one that already stopped and is still draining its tail is no longer discarded with it


CougarTasker (approved):
LGTM on the condition that the test is coming

PR #960, opened 2026-09-02
fix: keep a relayed stop from closing the next recording

damonhayhurst-neuraco:
### Features

 - The daemon reads the backend's recording notification stream and opens or closes windows by cloud recording id

### Bugfixes

 - A stop the SDK only relayed no longer closes whichever window is live when it lands, truncating the next recording to the network hop

### DIFF
--- a/rust/data_daemon/src/pipeline/dispatcher.rs
+++ b/rust/data_daemon/src/pipeline/dispatcher.rs
@@ -139,7 +139,10 @@ impl DispatcherHandle {
 }
 
 /// Optional runtime context passed to the dispatcher.
-#[derive(Clone, Default)]
+///
+/// Not `Clone`: `notifications_rx` is the single consumer of the notification
+/// stream, and the dispatcher is its only owner.
+#[derive(Default)]
 pub struct DispatcherContext {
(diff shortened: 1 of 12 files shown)


rust/data_daemon/src/pipeline/dispatcher.rs
```diff
@@ -762,6 +830,208 @@ impl Dispatcher {
...
+    }
+
+    /// Open a window for a pending backend recording now that `source` has
+    /// proved local by publishing at `publish_ts`.
+    ///
+    /// The lower bound is the recording's own start, so what the producer
+    /// published before this datum still lands inside. It is clamped to
+    /// `publish_ts` so the triggering datum is not excluded, and floored at any
+    /// predecessor's close so it cannot steal that recording's tail.
+    async fn adopt_remote_recording(&mut self, source: &Source, publish_ts: i64) {
```
CougarTasker: Issue: discussed offline that this added complxity for remote recordings is not needed and re-using the existing window machinery will be simpler with less edge cases. The idea is that remote start or stops are identical to local ones the dispatcher should not know the diffrence

CougarTasker (approved):
LGTM \w comment

PR #961, opened 2026-09-02
fix: stop gating logged data on the SDK's view of a recording

damonhayhurst-neuraco:
### Bugfixes

 - `log_*` forwards to the daemon unconditionally; the daemon decides what falls inside a recording window


CougarTasker (approved):
LGTM 👍

PR #962, opened 2026-09-02
fix: stop carrying recording identity in the data streams

damonhayhurst-neuraco:
### Bugfixes

 - Data streams carry no recording identity; the daemon owns it, so arming a stream only resets its timestamp timeline
 - The joint group cache is keyed on joint names alone and no longer rebuilds every binding at a recording boundary

### DIFF
--- a/neuracore/api/logging.py
+++ b/neuracore/api/logging.py
@@ -25,16 +25,13 @@ from neuracore_types import (
     PointCloudData,
     PoseData,
     RGBCameraData,
-    RobotInstanceIdentifier,
 )
 from neuracore_types.utils import validate_safe_name
 
 from neuracore.api.core import _get_robot
-from neuracore.api.globals import GlobalSingleton
 from neuracore.core.exceptions import RobotError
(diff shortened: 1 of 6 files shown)


neuracore/core/robot.py
```diff
@@ -409,6 +410,20 @@ def _register_remote_stop_handler(self) -> None:
         )
         self._recording_state_manager = manager
 
+    def _start_all_streams(self) -> None:
```
CougarTasker: Issue: if the streams can start lazly, e.g. a stream can be added after start recording then this functionality is redunant, please remove

CougarTasker (changes requested):

CougarTasker (commented):
discussed offline: the real call here may be to to remove the data streams and directly just call into the deamon

neuracore/core/robot.py
damonhayhurst-neuraco: removed

CougarTasker (approved):
LGTM 👍

PR #963, opened 2026-09-02
feat: answer which recording a source has open over local IPC

damonhayhurst-neuraco:
### Features

 - `query_recording_state` reports the recording a source has open, so a process can learn about one it did not start

### Bugfixes

 - A recording stopped from the web is owed a flush marker no producer will send; the window now settles on the source falling silent rather than sitting out the full 30s cap

### DIFF
--- a/neuracore/data_daemon/bridge.py
+++ b/neuracore/data_daemon/bridge.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import logging
 from importlib import import_module
 from types import ModuleType
+from typing import NamedTuple
 
 from neuracore.data_daemon.daemon_control import ensure_daemon_running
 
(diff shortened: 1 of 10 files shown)


rust/data_daemon/src/state/store.rs
```diff
@@ -1340,6 +1356,30 @@ impl StateStore for SqliteStateStore {
         Ok(recording_id.flatten())
     }
 
+    async fn live_recording_for_source(
+        &self,
+        robot_id: &str,
+        robot_instance: i64,
+    ) -> Result<Option<LiveRecording>, StateStoreError> {
+        let row = sqlx::query_as::<_, (i64, Option<String>, Option<i64>)>(
```
CougarTasker: Blocker: discussed offline, this is not correct. the source of truth at the moment for wheter a source is recording is in the dispatcher at the moement.

A test if you have gotten this right. on one node (machine A) call `nc.start_recording()` on another node (machine B) a little while later check `nc.is_recording()` it should be true and not require node B to have started logging

neuracore/data_daemon/bridge.py
```diff
@@ -43,6 +44,54 @@ def _load_native() -> ModuleType:
...
+            returned by :meth:`RecordingContext.start_recording`.
+    """
+
+    recording_index: int
+    recording_id: str | None
+    start_timestamp_ns: int | None
+
+
+def query_recording_state(
+    robot_id: str, robot_instance: int, timeout_s: float = 0.5
```
CougarTasker: Issue: this is used by some controllers in performance ciritcal parts the timeout should be much less than half a second and it its not met we should not guess false we should throw an error.

CougarTasker (changes requested):

neuracore/data_daemon/bridge.py
damonhayhurst-neuraco: `timeout_s` set to 0.1, and silence raises `RecordingStateUnavailableError` instead of returning "not recording".

rust/data_daemon/src/state/store.rs
damonhayhurst-neuraco: The `recording_state` service now answers from the dispatcher's `LiveRecording`.

neuracore/data_daemon/bridge.py
```diff
@@ -43,6 +44,67 @@ def _load_native() -> ModuleType:
...
+    recording_id: str | None
+    start_timestamp_ns: int | None
+
+
+class RecordingStateUnavailableError(RuntimeError):
+    """The daemon did not answer a recording-state query in time."""
+
+
+def query_recording_state(
+    robot_id: str, robot_instance: int, timeout_s: float = 0.1
```
CougarTasker, line 72: note: discussed offline lets tune this to observed times

rust/data_daemon/src/ipc/listener.rs
```diff
@@ -298,6 +306,57 @@ async fn serve_recording_id_queries(
...
+                    let response = response.write_from_slice(&bytes);
+                    if let Err(error) = response.send() {
+                        tracing::warn!(%error, "failed to send recording-state reply");
+                    }
+                }
+                Err(error) => tracing::warn!(%error, "failed to loan recording-state reply sample"),
+            },
+            Err(error) => tracing::warn!(%error, "failed to encode recording-state reply"),
+        }
+    }
```
CougarTasker: Issue: tower of terror, use an intermeidery vairbles to avoid this deep nesting.

CougarTasker (approved):
LGTM \w comments

rust/data_daemon/src/ipc/listener.rs
damonhayhurst-neuraco: Flattened into intermediary `bytes` and `response` bindings with early `continue`s, matching the receive/decode style already used in the same loop. Amended into the commit and the stack rebased.

PR #964, opened 2026-09-02
fix: give the daemon sole ownership of recording state

damonhayhurst-neuraco:
### Bugfixes

 - `RecordingStateManager` is deleted and `is_recording()` asks the daemon, so it answers for a recording started in another process or from the web


CougarTasker (approved):
LGTM 👍

PR #968, opened 2026-09-02
test: add joints across processes case

damonhayhurst-neuraco:
### Features

 - Case covering joints logged from more than one process into a single recording


muneebneura (approved):
LGTM

PR #969, opened 2026-09-02
test: tighten duration assertions, assert all camera

damonhayhurst-neuraco:
### Bugfixes

 - Duration assertions tightened, and every camera is asserted rather than only the first


CougarTasker (approved):
LGTM 👍

PR #970, opened 2026-09-02
test: add split process start stop

damonhayhurst-neuraco:
### Features

 - Case where one process starts a recording and a separate process stops it


muneebneura (approved):
LGTM

PR #971, opened 2026-09-02
test: cover cancel from a separate recording-control process

damonhayhurst-neuraco:
### Features

 - Case cancelling a recording from a process that did not open it


muneebneura (approved):
LGTM

PR #972, opened 2026-09-02
test: cancel a recording either side of a valid one

damonhayhurst-neuraco:
### Features

 - Regression coverage for #959 - a cancel must discard only the live recording, not one still draining


CougarTasker (approved):
LGTM

PR #973, opened 2026-09-02
test: read a source's recording state from the daemon

damonhayhurst-neuraco:
### Bugfixes

 - The harness reads its process-local handle per frame and asks the daemon only where a window can change under it


CougarTasker (changes requested):
Discussed offline: we can remove the `_local_recording_handle` now

damonhayhurst-neuraco:
> Discussed offline: we can remove the `_local_recording_handle` now

Done in #994

CougarTasker (approved):
LGTM 👍

PR #976, opened 2026-09-02
test: re-enable performance deadlines

damonhayhurst-neuraco:
### Bugfixes

 - The performance case asserts its timer deadlines again, with the trace wait bounded by the case timeout rather than its own 120s default


favour-neuraco (approved):
LGTM

PR #977, opened 2026-09-02
fix: resolve mcap importer timestamps in nanoseconds

aditya-neuraco:
### Features
 - None

### Bugfixes
 - Importing an MCAP file could throw away a whole episode because of a rounding problem in how timestamps were stored. MCAP records the time of each frame as a whole number of nanoseconds. The importer turned that into a decimal number of seconds, and a decimal number that large can only tell apart times that are about 238 nanoseconds apart. Two frames recorded closer together than that ended up with exactly the same time. Neuracore requires each frame in a recording to have a later time than the one before it, so the import stopped and the entire episode was skipped.
 - The MCAP importer now keeps the original whole numbers of nanoseconds and works out how far each frame is from the start of the episode using them, so no accuracy is lost along the way.
 - The importer also makes sure each frame's time is later than the previous one for the same sensor. If two times still come out identical, the second is moved forward by the smallest amount the number format allows, under a millionth of a second. Every frame is kept, and the shift is far too small to matter for a camera recording at 30 frames per second.
 - Times that are further out of order than a millionth of a second are left exactly as they are, so genuinely broken recordings still fail loudly rather than being quietly patched up.

### Items

### Related PRs
 - None

### DIFF
--- a/neuracore-dictionary.txt
+++ b/neuracore-dictionary.txt
@@ -331,6 +331,7 @@ neginf
 neuraco
 neuracore
 newbyteorder
+nextafter
 nhead
 nheads
 njoints
--- a/neuracore/importer/core/base.py
+++ b/neuracore/importer/core/base.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import logging
+import math
 import multiprocessing as mp
 import os
 import random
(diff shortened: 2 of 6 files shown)


stepjam (approved):
LGTM

PR #978, opened 2026-09-02
fix: build mcap decoder factories per episode file

aditya-neuraco:
### Features
 - None

### Bugfixes
 - Importing more than one MCAP file could decode messages as the wrong type. MCAP numbers its message types separately in each file, and the importer reused one decoder across every file, so it kept using the first file's numbering.
 - Episodes after the first either failed with a confusing parsing error or, when the two types happened to be compatible, imported wrong values with no error at all.
 - The importer now builds a fresh decoder for each file.

### Items

### Related PRs
 - None


stepjam (approved):
LGTM

PR #979, opened 2026-09-02
feat: reject mcap datasets whose episodes disagree on a topic type

aditya-neuraco:
### Features
 - Before importing, the importer now checks that every episode records the same message type on each topic it reads.
 - If two episodes disagree it stops with a message naming the topic, both types and both files, instead of failing partway through.
 - The check reads only each file's index, so nothing is uploaded until it passes.

### Bugfixes
 - None

### Items

### Related PRs
 - Stacked on #978


stepjam (approved):
LGTM

PR #980, opened 2026-09-02
feat: batch signalling track endpoint

CougarTasker:
### Features
 - Live stream tracks are now collected for a short moment and registered in one request, so a robot that starts many sensors no longer sends a burst of requests to the signalling service.
 - A resurrected stream resubmits its tracks in one request instead of one for each track.

### Items
 - [NCP-1126](https://neuracore.atlassian.net/browse/NCP-1126)

### Related PRs
 - NeuracoreAI/neuracore_backend#582

This must not be released before NeuracoreAI/neuracore_backend#582 is deployed, because the batch endpoint does not exist on the current backend.

### DIFF
--- a/tests/unit/core/p2p/test_client_provider_stream_manager.py
+++ b/tests/unit/core/p2p/test_client_provider_stream_manager.py
@@ -1,0 +1,226 @@
+"""Tests for the track batching in the client provider stream manager."""
+
+import asyncio
+import threading
+import time
+from typing import Any
+from unittest.mock import MagicMock
+
+import pytest
+from neuracore_types import DataType, RobotStreamTrack
+
+from neuracore.core.const import STREAM_API_URL
+from neuracore.core.streaming.p2p.enabled_manager import EnabledManager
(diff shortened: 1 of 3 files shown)


tests/unit/core/p2p/test_client_provider_stream_manager.py
```diff
@@ -0,0 +1,226 @@
...
+    async def post(
+        self, url: str, headers: dict | None = None, json: Any = None
+    ) -> None:
+        """Record one post request."""
+        self.posts.append((url, json))
+
+
+@pytest.fixture
+def nc_loop():
+    """Event loop that simulates the neuracore async loop (runs in background)."""
```
damonhayhurst-neuraco, line 43: Is this something we should be drawing from a shared helper?

damonhayhurst-neuraco (approved):
LGTM w/ comment

tests/unit/core/p2p/test_client_provider_stream_manager.py
CougarTasker: yeah we can extract it once it is used elsewhere

PR #981, opened 2026-09-03
fix: update robot name auth error

StevenJacobs61:
### Bugfixes

 - Raise Auth error instead of Value error when we are authenticated.


muneebneura (approved):
LGTM

PR #982, opened 2026-09-03
refactor: rename ActiveWindow to RecordingWindow

damonhayhurst-neuraco:
### Bugfixes

 - `ActiveWindow` is renamed `RecordingWindow` so the dispatcher names the recording it tracks


CougarTasker (approved):
LGTM 👍

PR #985, opened 2026-09-04
test: consolidate the daemon case matrix

damonhayhurst-neuraco:
### Features

 - Network-only tiers for both suites, carrying the remote and split recording-control paths the offline daemon has no backend to drive
 - Cross-process producers back in the shared matrix: process-per-camera, per-limb-per-camera, and a producer that never brackets its own window

### Bugfixes

 - Each case now owns an axis combination no other covers; `OldPerThread` goes with the duplicates it leaves behind, since joining its threads before `stop_recording` never crossed a live boundary

### DIFF
--- a/tests/integration/platform/data_daemon/daemon_test_cases.py
+++ b/tests/integration/platform/data_daemon/daemon_test_cases.py
@@ -1,17 +1,40 @@
+"""The shared case matrix for the data-daemon integration suites.
+
+Every case owns an axis combination no other case covers. A case that only
+re-runs a covered combination at a different size belongs in a performance
+workload, not here.
+"""
+
+from dataclasses import replace
+
 from tests.integration.platform.data_daemon.shared.test_case.build_test_case import (
-    OldPerThread,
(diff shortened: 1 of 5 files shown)


tests/integration/platform/data_daemon/daemon_test_cases.py
```diff
@@ -61,41 +85,28 @@
...
+    # The transport's high-rate case, four boundaries deep with phase jitter.
+    # 250Hz is its de-rated ceiling; above that a run measures scheduling noise.
+    PerThread(
         duration_sec=10,
         recording_count=4,
         video_count=1,
         image_height=120,
         image_width=120,
         video_fps=120,
-        joint_fps=250,  # previously 500 but flaky due to sync
```
CougarTasker, line 93: nit: can we keep this as a reminder that it would be good to push the platform to this frequency again

CougarTasker (approved):
LGTM on the condition with the extra cases it is not slower

PR #986, opened 2026-09-07
fix: Send SDK version to detect breaking API incompatibilities

sdas-neuraco:
### Features

- Send both `neuracore_version` and `neuracore_types_version` to `/auth/verify-version`.

### Bugfixes

- Previously, SDK versions with incompatible API URLs could pass validation when their types major version still matched. Providing the SDK version allows the backend to reject SDK major-version mismatches before authentication proceeds.
- Add a regression test for the exact query parameter names and package versions.

### Related PRs

- https://github.com/NeuracoreAI/neuracore_backend/pull/593

### DIFF
--- a/neuracore/core/auth.py
+++ b/neuracore/core/auth.py
@@ -177,11 +177,16 @@ class Auth(EventEmitter, metaclass=SingletonMetaclass):
         """
         from neuracore_types import __version__ as nc_types_version
 
+        from neuracore import __version__ as neuracore_version
+
         try:
             session = thread_local_session(retry_transient=True)
             response = session.get(
                 f"{API_URL}/auth/verify-version",
-                params={"version": nc_types_version},
+                params={
+                    "neuracore_types_version": nc_types_version,
+                    "neuracore_version": neuracore_version,
+                },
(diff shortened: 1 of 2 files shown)


damonhayhurst-neuraco:
I think this is the wrong approach here. We should move our API routes to neuracore_types, so versioning is a single source of truth.

sdas-neuraco:
> I think this is the wrong approach here. We should move our API routes to neuracore_types, so versioning is a single source of truth.

If we move route definitions into neuracore_types, what happens when an older SDK installs a newer, compatible types release that changes a route—could it start calling the new URL while still using the old request behavior?

damonhayhurst-neuraco:
> > I think this is the wrong approach here. We should move our API routes to neuracore_types, so versioning is a single source of truth.
> 
> If we move route definitions into neuracore_types, what happens when an older SDK installs a newer, compatible types release that changes a route—could it start calling the new URL while still using the old request behavior?

We could say your version of neuracore is incompatible with this version of types.

StevenJacobs61 (approved):
LGTM

PR #987, opened 2026-09-07
test: override lossy video for daemon tests

damonhayhurst-neuraco:
### Bugfixes

 - Daemon now lossy by default. Fails some functional tests. Add override so lossless is the default in the tests. Still acknowledge profile overrides further along.


favour-neuraco (approved):
LGTM

PR #988, opened 2026-09-08
fix: cleanup robot after importer runs

ypang-neuraco:
### Features
- Robots were being left behind after importer runs
- Changed worker instance base to be above max(max existing robot instance, 100k) in a 100k window to avoid collision with existing instances
- Clean up robot instances after importer runs
- Live run: archive robots instead of delete because the URDF viewer will error if robot is deleted
- If robot existed before the importer run, leave it in the same state and do not cleanup
- Skip cleanup on shared robots
- Dry run still deletes the robot
- Update importer integration tests

### Items
 - [NCRL-853](https://neuracore.atlassian.net/browse/NCRL-853)

### Related PR
- https://github.com/NeuracoreAI/neuracore_backend/pull/600


sdas-neuraco:
Discussed in person with @muneebneura and @ypang-neuraco:  

If the user has tried live uploading data and then run the importer this would still leave the extra instances around which is not good UX. We should try to add code to archive specific instances, but i do not know the engineering effort required for this change. @ypang-neuraco will take a look to see if it is small, if it is large maybe discuss with @CougarTasker ?

sdas-neuraco (changes requested):
see previous comment (doing this to stop daily notifications)

ypang-neuraco:
Updated:
- Changed worker instance base to be above max(max existing robot instance, 100k) in a 100k window to avoid collision with existing instances
- Clean up robot instances after importer runs

sdas-neuraco (approved):
LGTM

PR #990, opened 2026-09-09
ci: remove macOS from PR unit tests

CougarTasker:
### Features
 - Drops the macOS legs from the PR unit test matrix, leaving six Linux legs.
 - Removes about 45 percent of the organisation's Actions spend, since macOS runners bill at ten times the Linux rate.
 - Leaves the SDK without macOS test coverage, though Build Wheels still builds and publishes the macOS wheels.

### Items
 - [NCP-1320](https://neuracore.atlassian.net/browse/NCP-1320)


damonhayhurst-neuraco (approved):
LGTM

PR #991, opened 2026-09-09
fix: algorithm perf test gpu type

CougarTasker:
### Bugfixes
 - The algorithm performance integration test now takes the GPU type per algorithm from `algorithm_configs.yaml`, so Pi0, Pi05 and Groot start on an A100 instead of the hardcoded V100 they do not support.
 - The evaluate phase deploys its endpoint on the same per-algorithm GPU, so the large models are not served on a V100 that cannot hold them.


damonhayhurst-neuraco (approved):
LGTM

PR #992, opened 2026-09-09
fix: allow v100 for large vlas

CougarTasker:
### Bugfixes
 - Pi0, Pi05 and Groot declare the V100 as a supported GPU again, because A100 and H100 capacity is limited.
 - The algorithm performance integration test drops its per-algorithm GPU overrides, so these three algorithms go back to the default V100.

### Related PRs
 - Follows #991, which this partly reverts.


CougarTasker:
Superseded: keeping the supported GPU sets unchanged and exempting test-scale model configs from the check instead.

PR #819, opened 2026-07-28
feat: validate algorithm GPU compatibility

sdas-neuraco:
## Summary

- Declare supported GPU types for built-in algorithms.
- Validate the selected GPU before starting a cloud training run.
- Document the GPU declaration required for custom algorithms and add release notes.
- Add unit coverage for supported and unsupported GPU selections.

Ticket: [NCRL-759](https://neuracore.atlassian.net/browse/NCRL-759)

Related PR: [NeuracoreAI/neuracore_backend#480](https://github.com/NeuracoreAI/neuracore_backend/pull/480)

PR #993, opened 2026-09-09
fix: exempt small models from gpu check

CougarTasker:
### Features
 - None

### Bugfixes
 - The GPU compatibility check now skips configs that select a test-scale variant of a large model, so the integration tests can start Pi0, Pi05 and Groot on a V100 while the algorithms keep declaring only the GPUs their full-size variants need.
 - The algorithm performance test drops its per-algorithm GPU overrides, because the exemption lets these three algorithms use the default V100 again.

### Items
 - None

### Related PRs
 - Adjusts the check added in #819 and replaces the overrides added in #991.

### DIFF
--- a/neuracore/api/training.py
+++ b/neuracore/api/training.py
@@ -96,12 +96,33 @@ def get_algorithm(algorithm_id: str) -> dict:
     return response.json()
 
 
+# Test-scale variants of the large vision language models fit on any GPU. They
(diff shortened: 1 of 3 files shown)


ypang-neuraco (approved):
LGTM

aditya-neuraco (approved):
LGTM

PR #994, opened 2026-09-09
fix: scope the timestamp check to the daemon's recording

damonhayhurst-neuraco:
### Features

 - The producer caches the daemon's recording state per source, so the log path can tell one recording's timeline from the next without an IPC round trip. A recording this process brackets is seeded exactly, at once — `start_recording` returns the very capture timestamp the daemon stores as the recording's start — and one opened elsewhere is picked up by an off-thread refresh through the existing `recording_state` query

### Bugfixes

 - Monotonic timestamp checking is scoped to the recording the daemon has open rather than to `_local_recording_handle`. A recording started from the web or another process armed nothing and went unchecked; the timeline now resets at every boundary rather than only at a start this process published
 - `Robot.start_recording` returns nothing; the handle it returned was a client-minted UUID, never a recording id
 - Removes `_local_recording_handle`, as requested on #973


neuracore/core/streaming/data_stream.py
```diff
@@ -47,27 +46,6 @@ def data_type(self) -> DataType:
...
         return self._data_type
 
-    def start_recording(self) -> None:
-        """Arm the stream for a new recording.
-
-        The stream carries no recording identity — the daemon owns that. This
-        only marks a fresh timeline for :meth:`_enforce_monotonic_timestamp`.
-        """
-        self._recording = True
-        self._last_logged_timestamp = None
```
CougarTasker, line 57: Blocker: I overlooked this earlier, without this reset the check will fail valid recordings. e.g. if the user uploads recordings out of order from the importer. e.g. upload newest to oldest

Suggestion: I think instead of keeping the latest timestamp state on the python side per data stream we should keep a ceneral repostory of per robot instance timestamps closer to the deamon so that it can be reset when the recording is started, through SSE or otherwise

The current system is not much better it has the bug from the archetcutre shift introduced in this stack, you can reporoduce it it by logging a datapoint now starting a recording over sse and logging a frame with a timestamp from yesterday, this is valid but would raise an error.

CougarTasker (changes requested):

neuracore/core/streaming/data_stream.py
damonhayhurst-neuraco: I've added a recording cache. Its contents are seeded by two pathways: start/stop recording calls, and the recording_state query introduced in #963. The latter is bounded by IDLE_POLL_INTERVAL, but since the purpose of this cache is to raise an error for out-of-order timestamps, sub-ms latency was deemed not to be needed.

I added your SSE case as a test in `test_explicit_timestamps`.

rust/data_daemon_bridge/src/recording_boundary.rs
CougarTasker, line 1: nit: naming, too generic its not really a cache of a recording just the bounry state

CougarTasker (approved):
LGTM \w comment

PR #998, opened 2026-09-11
refactor: split transport from service semantics in the bridge queries

damonhayhurst-neuraco:
### Bugfixes

 - The bridge's request/reply transport is lifted into `request_reply_once` and `poll_until`, leaving the probes only their service semantics

### Items


CougarTasker (approved):
LGTM

PR #999, opened 2026-09-11
refactor: promote Source to the shared crate

damonhayhurst-neuraco:
### Bugfixes

 - `Source` moves to `data_daemon_shared` rather than being declared privately in the daemon and the bridge alike

### Items


CougarTasker (approved):
LGTM 👍

PR #1000, opened 2026-09-11
refactor: align the daemon's recording id and index names

damonhayhurst-neuraco:
### Bugfixes

 - `cloud_recording_id` becomes `recording_id` and `opened_index` becomes `recording_index`, so the daemon names each field one way

### Items


CougarTasker (approved):
LGTM 👍

PR #1002, opened 2026-09-11
feat: serve recording state from one cached display

damonhayhurst-neuraco:
### Features

 - Recording state is served from one daemon-backed cache rather than tracked per process, so `is_recording` answers the same for a window opened here, in another process, or from the web
 - `get_current_recording_id` is dropped from the public API surface, `get_cloud_recording_id` drops `timestamp_ns` and resolves the open recording only


neuracore/core/robot.py
```diff
@@ -361,44 +363,40 @@ def _recording_epoch(self) -> int | None:
...
-            timestamp_ns=timestamp_ns, timeout_s=timeout_s
-        )
+        deadline = time.monotonic() + timeout_s
+        while True:
+            recording = recording_context.query_recording_state(self.id, self.instance)
+            if recording is not None and recording.cloud_recording_id is not None:
+                return recording.cloud_recording_id
+            if time.monotonic() >= deadline:
+                return None
+            time.sleep(_CLOUD_RECORDING_ID_POLL_INTERVAL_S)
```
CougarTasker: nit: waiting with the GIL is inefficent
suggestion:  `recording_context.query_recording_state` should expose the timeout and wait on the rust side

CougarTasker (approved):
LGTM \w comment

PR #1004, opened 2026-09-14
test: list only a case's own timer rows in the session summary

damonhayhurst-neuraco:
### Bugfixes

 - The session summary lists only the timer labels a case produced, not every known label padded with `---`

### Items


CougarTasker (approved):
LGTM 👍

PR #1005, opened 2026-09-14
test: pace lifetime producers and refuse saturate with one

damonhayhurst-neuraco:
### Bugfixes

 - Producers that outlive their recordings are paced rather than unbounded, so a camera child cannot outrun the encode permits a CI runner grants
 - `PACING_SATURATE` is refused for anything but a recording-scoped producer, which keeps the combination from returning by accident
 - The three heaviest cross-process integrity cases drop to 160x120, where the shape under test is the process boundary rather than the frame size
 - Recording counts are reduced across the daemon case matrix to maintain the suite's total runtime from before this stack

### Items


tests/integration/platform/data_daemon/shared/test_case/build_test_case.py
```diff
@@ -87,6 +88,15 @@ def _unsupported_combination(case: DataDaemonTestCase) -> str | None:
             f"producer_pacing={PACING_BURST_VIDEO!r} needs video_count > 0 or "
             "depth_count > 0: it clumps the video streams and there are none"
         )
+    if (
+        case.producer_pacing == PACING_SATURATE
+        and case.producer_channels != PRODUCER_SYNCHRONOUS
+    ):
+        return (
+            f"producer_pacing={PACING_SATURATE!r} needs a recording-scoped "
```
CougarTasker: Question: why are these modes impcompatible, and this is not an issue with the test or code?

damonhayhurst-neuraco: You're right to question this. In my initial testing I thought there was little difference between saturate and deadline variants for lifetime producers. This made me decide to remove the variant from the case as I presumed they were simply logging past their intended number of log messages and causing serious contention in low resource environments. 
However I then realised this should not be the case regardless of how fast the producer offers frames. This then led me uncover that we set the publish timestamp after flushing the published data meaning that any data added to the queue during that time makes it in even if it is passed the publish time. This is reflected in #1009. 
This PR remains open to reduce the length of the test suite to pre-stack levels.

CougarTasker: Discussed offline: the best compormise here is to use saturate on the fixed log points we know about e.g. the ones before and during the recoding then pace the logging during the unbounded stop recording. this gives us not the continus logging

muneebneura (approved):
LGTM

PR #1006, opened 2026-09-14
test: budget and attribute the stop-recording wait

damonhayhurst-neuraco:
### Bugfixes

 - The `wait=True` stop-recording budget makes its fixed cost additive rather than a `duration_sec * 0.5` floor, which modelled a cost that does not scale with window length
 - `Timer` separates the CPU a thread burned from the time it merely waited, so a polling wait no longer reports `local-bound`

### Items


CougarTasker (approved):
LGTM 👍

PR #1007, opened 2026-09-14
fix: flush a producer's video on a stop it did not call

damonhayhurst-neuraco:
### Bugfixes

 - A process with video that brackets no recording of its own reports its flush barrier off the state it already watches, rather than sitting out the daemon's 30s cap

### Items


CougarTasker (changes requested):
Blocker: This is the third patch to a new and now large system where a simpler system would not have these weaknesses.

We have patched this same issue of bit of the code a few three times allready I think it doesn't realy fit naturally into the archetecture we have here. This is a new feature that has been repeatedly faulty, I don't think its worth comitting more resources to continue patch this system. We have now this informational back and fourth between the deamon and producer, this coupling will be error prone. 

The fundemental problem we are trying to solve here is that the deamon does not know about the producer's endcoding/logging backlog so does not know how long to wait to consider the recording comepleted.

I think a simpler approach would be to remove this SourceFlushed and related code and instead extend the VideoProducerActive with a lowest publish timestamp that is not sent "unsealed". This approach is commonly used in compleness of stream processing and called a watermark.

This way the logic gets much simpler on the deamon side, it just checks if the producer still has chunks that fit in the window. and we completley side step a lot of complexity and potential bugs


the prevous patches:
[#963](https://github.com/NeuracoreAI/neuracore/pull/963)
[#877](https://github.com/NeuracoreAI/neuracore/pull/877)

the feature added 
[#876](https://github.com/NeuracoreAI/neuracore/pull/876)


Watermark Paper: https://www.vldb.org/pvldb/vol14/p3135-begoli.pdf

CougarTasker (approved):
Discussed offline: as a temporary hack this will get CI to go green but ultimately we need to get this actually fixed before we waste more effort on it 

https://neuracore.atlassian.net/browse/NCP-1358

PR #1008, opened 2026-09-14
fix: re-download partially downloaded model archives

aditya-neuraco:
### Bugfixes
 - Model and checkpoint downloads no longer leave a broken file behind when the transfer is interrupted. The file is now written to a temporary name first and only moved into place once it has downloaded completely, so a failed or cancelled download leaves nothing for a later run to pick up.
 - A file already in the download cache is now only reused when its size matches the size the server reports for it. A file that is smaller than expected, for example one left over from an earlier interrupted download, is downloaded again instead of being loaded as if it were complete.
 - Both of these apply to loading a model from a training run and to loading a training checkpoint, which previously shared the same problem.

### Items
 - [NCRL-874](https://neuracore.atlassian.net/browse/NCRL-874)


neuracore/core/utils/download.py
```diff
@@ -92,8 +121,39 @@ def download_with_progress(
...
+        description: Description for the progress bar.
+
+    Returns:
+        Path to the cached file.
+    """
+    destination = Path(destination)
+    destination.parent.mkdir(parents=True, exist_ok=True)
+    if destination.exists():
+        expected = remote_content_length(url)
+        if expected is None or destination.stat().st_size == expected:
```
sdas-neuraco, line 157: if the remote size is None it should pass this seems incorrect maybe remove this?

sdas-neuraco (approved):
LGTM w/c

neuracore/core/utils/download.py
sdas-neuraco: if destination.stat().st_size == expected: is my suggeestion

neuracore/core/utils/download.py
```diff
@@ -63,11 +65,38 @@ def download_bytes(url: str) -> bytes:
...
+    """
+    session = thread_local_session(retry_transient=True, retry_read_timeout=True)
+    try:
+        with session.get(url, stream=True, timeout=DOWNLOAD_TIMEOUT_S) as response:
+            response.raise_for_status()
+            if response.headers.get("Content-Encoding"):
+                return None
+            length = response.headers.get("Content-Length")
+            return int(length) if length is not None else None
+    except (requests.RequestException, ValueError):
```
sdas-neuraco: this should throw an error not silently return None

neuracore/core/utils/download.py
aditya-neuraco: Good suggestion actually! 

Removed the `if expected is None`  None check as GCS returns a plain Content-Length with Content-Encoding as None, so None is unreachable in that case.

neuracore/core/utils/download.py
aditya-neuraco: Agreed.

PR #1009, opened 2026-09-14
fix: stamp the recording's upper bound when stop is called

damonhayhurst-neuraco:
### Bugfixes

 - `stop_recording` stamps the window's upper bound when called rather than after draining the publisher, so a producer's backlog no longer widens the window

### Items


CougarTasker (approved):
LGTM 👍

damonhayhurst-neuraco:
Closed because this seems to cause a regression in performance in the 8 context case.

PR #1010, opened 2026-09-14
fix: start the data daemon when a robot connects

damonhayhurst-neuraco:
### Bugfixes

 - `connect_robot` starts the data daemon, so a producer that never calls `start_recording` still hears a recording started from the web
 - A daemon that cannot start now fails the connect instead of being logged and swallowed


neuracore/api/core.py
```diff
@@ -179,6 +181,8 @@ def connect_robot(
     if robot.id is None:
         raise RobotError("Robot not initialized. Call init() first.")
     StreamManagerOrchestrator().get_provider_manager(robot.id, robot.instance)
+    # Only the daemon hears a recording started from the web.
+    ensure_daemon_running()
```
muneebneura, line 185: Question: Why are we launching data daemon on robot connect even for callers that never intend to record?

muneebneura (changes requested):
Issue: This does not fix the remote stop. the data daemon still doesnt start uploading, it sits on the closed window for the full flush marker cap before finalizing it.

Here are the logs on data daemon:
`WARN no producer flush marker before the cap; retiring the window anyway — any tail video chunk still in flight will be dropped as an orphan recording_index=1 robot_id="d28018aa-ac7c-4921-898d-71528f955c07" elapsed_s=30.020643038 outstanding_producers=1`

muneebneura (approved):
LGTM! Can you stack this one on top of other fixes as well.

PR #1014, opened 2026-09-15
fix: only return sync point data from the connected robot

aditya-neuraco:
### Bugfixes

 - `get_latest_sync_point` now uses the robot you pass to it. Before, it ignored that argument and always read whichever robot was connected most recently, so a program using more than one robot could ask for robot A and get robot B's sensor data back.
 - Asking for a robot that is not the currently connected one now fails with an error saying that robot is not connected, instead of quietly returning a different robot's data. This is the safer behaviour: stale data from the wrong robot is harder to spot than an error.
 - When no robot is connected at all, the error raised is now `RobotError` instead of `ValueError`. This matches the error the rest of the client raises in the same situation, and matches what the function's own documentation already promised.
 - Added two tests, one for each of the two errors above.

### Items

 - [NCRL-873](https://neuracore.atlassian.net/browse/NCRL-873)


neuracore/core/get_latest_sync_point.py
```diff
@@ -84,21 +83,33 @@ def check_remote_nodes_connected(robot: Robot, num_remote_nodes: int) -> bool:
...
+    if active_robot is None:
+        raise RobotError("No active robot. Call connect_robot() first.")
     if robot is None:
-        raise ValueError("No active robot found. Please initialize a robot instance.")
+        robot = active_robot
+    elif robot is not active_robot:
+        raise RobotError(
+            f"Robot {robot.name}:{robot.instance} is not connected. "
+            "Call connect_robot() to make it the active robot."
+        )
```
ypang-neuraco: I'm confused about this, if we would never use the robot passed in then why keep it as an input?
@sdas-neuraco can you please confirm the expected behaviour? Maybe we want to log a warning and still go ahead and return the latest sync point for the robot passed in.

mark-neuracore: I have the very same comment about this part. Robot is still pointless here. If i am sending a robot then it means i want the data from a robot i am not connected to (e.g. script is connected to robot a but i want data from robot b).  

Whether we agree this is a needed feature to have or not is another question, but if it is no, then we do not need to pass robot to this function at all.

Right now We HAVE to pass robot to be the active robot otherwise we fail, and if nothing is passed, we default to active robot.

mark-neuracore (changes requested):

neuracore/core/get_latest_sync_point.py
aditya-neuraco: Updated to follow the behavior discussed in the slack snippet.

ypang-neuraco (approved):
LGTM. @sdas-neuraco @mark-neuracore I think this is the right fix for now, but per the discussion on slack we will need to rethink this in the future

PR #1015, opened 2026-09-15
fix: patch transformers unconditionally rather than on a symbol check

damonhayhurst-neuraco:
### Bugfixes

 - `pi0` and `pi05` copy their `transformers_replace/` files on every import, so a half-patched install repairs itself rather than passing a gemma-only symbol check
 - The copy renames into place, so workers patching at once can no longer expose a truncated `modeling_paligemma.py` to a worker importing it


mark-neuracore (approved):
LGTM. We agreed this is a patch according to our discussions offline.

PR #1018, opened 2026-09-15
fix: adopt recordings already open when the daemon subscribes

damonhayhurst-neuraco:
### Bugfixes

 - The daemon now acts on the backend's `INIT` snapshot, so a recording minted while it was starting, down, or reconnecting is picked up rather than lost
 - `START` is a one-shot live event with no replay, so a daemon that was not subscribed when it fired reported the source idle for that recording's whole life
 - This is the flake behind `remote.start_gate_wait` timing out at 20s on the staging matrix, where the gate normally settles in ~0.3s
 - `test_connect_robot_starts_the_daemon` now proves the premise its docstring already claimed, a peer connecting into a window the web opened while no daemon was running
 - The startup test deletes the cloud dataset and robot it creates, which the connect-only version left behind


CougarTasker (approved):
LGTM 👍

PR #1020, opened 2026-09-16
fix: use metadata.name directly for recording name

sdas-neuraco:
## Summary
- Simplifies `Recording.__init__` to set `self.name = metadata.name` directly instead of falling back to `recording_id` via `getattr`.


damonhayhurst-neuraco (approved):
LGTM

PR #1030, opened 2026-09-17
Add local MCAP dataset exports with recording names

sdas-neuraco:
### Features

- Add `neuracore export mcap --dataset NAME --output PATH` 
- Save recordings as `nc_RECORDING_NAME.mcap`, sanitize unsafe filename characters, and add numeric suffixes for duplicates. Progress messages and validation errors show recording names; IDs remain in the manifest and embedded metadata for traceability.
- Keep conversion and media helpers inside `McapExporter`, with validation and writing implemented directly in its lifecycle methods.
- Designed for the core instrastrucutre to then expand to LeRobot in Seperate PR, and then later microservices on the backend.


tests/scripts/mcap_viewer.py
ypang-neuraco, line 1: please move this out of tests

ypang-neuraco: please move this out of tests

ypang-neuraco (approved):
LGTM \w comments. Maybe add an integration test in the future

PR #1031, opened 2026-09-17
test: use Dataset.delete_recording in integration tests

sdas-neuraco:
Integration tests now exercise `Dataset.delete_recording(recording_id=recording.id)` when removing recordings, replacing the direct HTTP helper.


aditya-neuraco (approved):
LGTM

PR #1032, opened 2026-09-17
fix: keep the daemon log across restarts and rotate it

damonhayhurst-neuraco:
### Features

 - The daemon owns its log through `logroller`, appending across restarts and rotating at 10 MB with five archives kept
 - `NEURACORE_DAEMON_LOG_PATH` names that file and is resolved identically either side of the process boundary, with `NCD_LOG_MAX_SIZE` and `NCD_LOG_MAX_FILES` tuning the rotation
 - Every data-daemon test gets its own `daemon.log` in its numbered artifact directory, attached to that test's report as a bounded tail

### Bugfixes

 - The launcher opened `daemon.log` with `"wb"`, so every start erased the log of the run before it
 - Nothing but the daemon writes that file now, the launch process's own stderr going to a sibling `daemon-launcher.log` that appends
 - The SDK spawns `launch` without `--background`, so tracing went to stderr and the Rust append branch never ran on the path that matters
 - `with_ansi(false)` was set on the file branch alone, so the log the SDK actually produced carried escape codes
 - `init_tracing` moved below the PID flock, so a launch that loses the race cannot open a second rotating writer over one path
 - A premature-exit error reads both logs from the offset they held before the launch, rather than quoting the previous daemon's shutdown
 - `daemon_log_action` is gone, dead machinery no case ever set, and wiping a run's logs stays the local runner's existing state purge


neuracore/data_daemon/daemon_control.py
```diff
@@ -310,40 +324,96 @@ def _start_daemon_subprocess(
...
+    try:
+        if path.stat().st_size <= _LAUNCHER_LOG_MAX_BYTES:
+            return
+        with open(path, "rb") as handle:
+            handle.seek(-_LAUNCHER_LOG_MAX_BYTES, os.SEEK_END)
+            tail = handle.read()
+        path.write_bytes(tail)
+    except OSError:
+        # Trimming is housekeeping; never fail a launch over it.
+        pass
```
CougarTasker: Question: I don't understasnd why we need to do this is the size not handled by log roller? what do you mean by a crash loop, did this happen and how?

CougarTasker (approved):
LGTM \w question

PR #1034, opened 2026-09-18
fix: detect zombie processes on macOS so daemon stop returns promptly

damonhayhurst-neuraco:
### Bugfixes

 - `pid_is_running` excludes zombies on macOS via `proc_pidinfo`, where the filter was a hardcoded `false` and an exited-but-unreaped daemon read as live
 - `data-daemon stop` on macOS returns as soon as the daemon exits, rather than waiting out its 10s graceful and 5s SIGKILL timeouts and exiting non-zero on a process that already died
 - Mac integration lanes lose roughly 12 minutes a job, since every case paid the 16s stop against the Linux 1s


rust/data_daemon/src/lifecycle/pidfile.rs
```diff
@@ -161,13 +157,36 @@ fn is_zombie(pid: i32) -> bool {
     after_comm.split_whitespace().next() == Some("Z")
 }
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(target_os = "macos")]
+fn is_zombie(pid: i32) -> bool {
```
CougarTasker, line 165: Question: Why does this implementation need to be mac specific, it looks generic. If so please use it for linux too.

CougarTasker (approved):
LGTM \w comment

PR #1037, opened 2026-09-18
fix: daemon default api url domain

CougarTasker:
### Bugfixes
 - The Rust data daemon now defaults to the neuracore.com API domain, the same as the Python SDK, in place of the old neuracore.app domain.


damonhayhurst-neuraco (approved):
LGTM

aditya-neuraco (approved):
LGTM

